PR #1, opened 2022-06-16
Feat/crypto

Solpatium:


lib/src/main/kotlin/starknet/types/Felt.kt
```diff
@@ -0,0 +1,33 @@
+package types
+
+import java.math.BigInteger
+
+data class Felt(val value: BigInteger){
```
cptartur, line 5:
```suggestion
data class Felt(val value: BigInteger) {
```

lib/src/main/kotlin/starknet/crypto/StarknetSignature.kt
```diff
@@ -0,0 +1,31 @@
...
+import types.toFelt
+import java.io.ByteArrayOutputStream
+
+data class StarknetSignature(val r: Felt, val s: Felt) {
+    companion object {
+        @JvmStatic
+        fun fromASN1(input: ByteArray): StarknetSignature {
+            val ret = ASN1Sequence.getInstance(input)
+            val r = (ret.getObjectAt(0) as ASN1Integer).positiveValue
+            val s = (ret.getObjectAt(1) as ASN1Integer).positiveValue
```
cptartur, line 14: I think we're risking an exception here in case ret[0] or [1] cannot be cast. Maybe it'd be better to use `as?` cast and check then validate r and s?

lib/src/main/kotlin/starknet/crypto/functions.kt
```diff
@@ -0,0 +1,70 @@
...
+    return ecdsaVerify.verify(signature.toASN1())
+}
+
+fun pedersen(first: Felt, second: Felt): Felt {
+    var point = SAMPLED_GENERATORS[0]
+    for (i in 0..1) {
+        var x = if (i == 0) first.value else second.value
+        for (j in 0 until 252) {
+            val iterationPoint = SAMPLED_GENERATORS[2 + i * 252 + j]
+            assert(point.xCoord != iterationPoint.xCoord)
```
cptartur, line 56: Wouldn't it be better to throw a (custom) exception here instead?

lib/src/main/kotlin/starknet/crypto/functions.kt
```diff
@@ -0,0 +1,70 @@
...
+
+            if (x.testBit(0)) {
+                point = point.add(iterationPoint);
+            }
+            x = x.shiftRight(1)
```
cptartur, line 61: I'd consider replacing a loop with helper function and two calls to it, that'd avoid `If (i == 0) ...`

cptartur (commented):
I'd consider replacing usages of `also` with `apple` where possible.

lib/src/main/kotlin/starknet/crypto/functions.kt
Solpatium: No, it is more like a sanity check.

lib/src/main/kotlin/starknet/crypto/StarknetSignature.kt
Solpatium: This means someone provided invalid input, tracing what went wrong wouldn't be a problem.

lib/src/main/kotlin/starknet/crypto/functions.kt
Solpatium: Change to an array to get index. I want it to look like other implementations

lib/src/main/kotlin/starknet/crypto/StarknetSignature.kt
cptartur: You're probably right about tracing, but I think without explicit validation here, program will fail with something like `NullPointerException` and I don't know if that's desired behavior.

Solpatium: I think we don't need it, we would throw in this case anyway. Throwing `IllegalArgumentException` provides less information than exception like `class X cannot be cast to ASN1Integer`- this way user at least knows what byte array really holds, this could be useful for debugging.

lib/src/main/kotlin/starknet/crypto/constants.kt
```diff
@@ -0,0 +1,2078 @@
+@file:JvmName("constants")
```
bartekryba, line 1: What about using PascalCase for these names?

lib/src/main/kotlin/starknet/crypto/functions.kt
```diff
@@ -0,0 +1,70 @@
+@file:JvmName("functions")
```
bartekryba, line 1: Same thing here

lib/src/main/kotlin/starknet/types/Felt.kt
```diff
@@ -0,0 +1,33 @@
...
+        val PRIME = BigInteger("800000000000011000000000000000000000000000000000000000000000001", 16)
+
+        @field:JvmField
+        val ZERO = Felt(BigInteger.ZERO)
+
+        @field:JvmField
+        val ONE = Felt(BigInteger.ONE)
+
+        @JvmStatic
+        fun fromHex(value: String): Felt = Felt(BigInteger(value.removePrefix("0x"), 16))
```
bartekryba, line 28: If we'd like to be more strict about hex vs dec strings, we could fail when the '0x' prefix is not provided. But I'm not sure about it.

PR #4, opened 2022-06-23
Add Account and Signer

Solpatium:


lib/src/main/kotlin/starknet/data/types/execution.kt
```diff
@@ -0,0 +1,38 @@
+package starknet.data.types
+
+import starknet.data.selectorFromName
+import types.Felt
+
+data class Call(
+    val contractAddress: Felt,
+    val entrypoint: String,
+    val calldata: Calldata,
```
bartekryba, line 9: Right now in Invocation data class, calldata is optional, do we want it to be optional here as well?

bartekryba (approved):

lib/src/main/kotlin/starknet/data/types/execution.kt
Solpatium: Calldata itself is an optional type, right?

PR #5, opened 2022-06-23
Feat/provider

bartekryba:
WIP of Provider.


lib/src/main/kotlin/starknet/provider/Service.kt
```diff
@@ -0,0 +1,27 @@
...
+    override fun <T: Response> send(request: Request<T>): T {
+        val inputStream = performIO(request.payload)
+
+        return Json.decodeFromString(request.deserializer, inputStream.toString())
+    }
+
+    override fun <T: Response> sendAsync(request: Request<T>): CompletableFuture<T> {
+        return CompletableFuture.supplyAsync {
+            send(request)
+        }
```
Solpatium, line 25: Shouldn't we use async handler from ok http?

Solpatium (changes requested):

lib/src/main/kotlin/starknet/provider/HttpService.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        val response = client.newCall(request).execute()
+        val responseBody = response.body
+
+        if (response.isSuccessful) {
+            return if (responseBody != null) {
+                ByteArrayInputStream(responseBody.bytes())
+            } else {
+                // TODO: Should it be empty byte array or null returned from performIO method?
+                ByteArrayInputStream(null)
```
Solpatium, line 34: Empty byte array

lib/src/main/kotlin/starknet/provider/HttpService.kt
```diff
@@ -0,0 +1,43 @@
...
+            }
+        } else {
+            val code = response.code
+            val text = response.body?.string() ?: "unknown"
+
+            throw java.lang.Exception("Invalid response, code: $code, response: $text")
```
Solpatium, line 40: Custom exception would be nice

lib/src/main/kotlin/starknet/provider/HttpService.kt
```diff
@@ -0,0 +1,79 @@
...
+        } else {
+            val code = response.code
+            val text = response.body?.string() ?: "unknown"
+
+            // TODO: Create custom exception
+            throw java.lang.Exception("Invalid response, code: $code, response: $text")
+        }
+    }
+
+    override fun <T : Response> send(request: Request<T>): T {
```
Solpatium, line 49: `Request` depends on `HttpService` and `HttpService` depends on `Request`. I think we can just accept payload as a string here and the Request should handle deserialization. This way we have one way dependency.

Solpatium (changes requested):

lib/src/main/kotlin/starknet/crypto/keccak.kt
```diff
@@ -5,7 +5,7 @@ import types.Felt
 import types.toFelt
 import java.math.BigInteger
 
-private val MASK_250 = BigInteger.TWO.pow(250) - BigInteger.ONE
+private val MASK_250 = BigInteger.valueOf(2).pow(250) - BigInteger.ONE
```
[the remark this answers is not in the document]
bartekryba: BigInteger.TWO is a private member of BigInteger

lib/src/main/kotlin/starknet/provider/HttpService.kt
bartekryba: Request depends only on Service interface, which abstracts details of HttpService.

bartekryba: I was also wondering whether there should be two kinds of requests for gateway and jsonrpc2.0 calls

lib/src/main/kotlin/starknet/crypto/keccak.kt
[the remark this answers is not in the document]
Solpatium: No it is not
```
public static final BigInteger TWO = valueOf(2);
```

lib/src/main/kotlin/starknet/provider/HttpService.kt
Solpatium: >Request depends only on Service interface, which abstracts details of HttpService.

Request takes in Service in constructor, Service takes Request as a parameter. This is a cycle that could be simplified.

>I was also wondering whether there should be two kinds of requests for gateway and jsonrpc2.0 calls

I think the request model should be the same for them, or at least they should have a common interface. I guess we could get rid of service and just create specific requests if you prefer that.

lib/src/main/kotlin/starknet/provider/service/HttpService.kt
```diff
@@ -34,15 +37,13 @@ class HttpService(private val url: String, private val method: String) : Service
...
-                throw java.lang.Exception("Empty response body")
+                throw HttpServiceException("Empty response body")
             }
         } else {
             val code = response.code
             val text = response.body?.string() ?: "unknown"
 
-            // TODO: Create custom exception
-            throw java.lang.Exception("Invalid response, code: $code, response: $text")
+            throw HttpServiceException("Invalid response, code: $code, response: $text")
```
Solpatium, line 46: We should save original message and code in exception. Writing them in string requires users to parse it to get it.

lib/src/main/kotlin/starknet/data/types/Felt.kt
```diff
@@ -22,6 +30,14 @@ data class Felt(val value: BigInteger) {
         return "Felt(${toHex(value)})"
     }
 
+    fun hexString(): String {
+        return "0x${toHex(value)}"
```
Solpatium, line 34: `toHex` already adds `0x`

lib/src/main/kotlin/starknet/data/types/Felt.kt
```diff
@@ -35,4 +51,21 @@ data class Felt(val value: BigInteger) {
...
+    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("Felt", PrimitiveKind.STRING)
+
+    override fun deserialize(decoder: Decoder): Felt {
+        val hex = decoder.decodeString()
+
+        return Felt.fromHex(hex)
+    }
+
+    override fun serialize(encoder: Encoder, value: Felt) {
+        val hex = "0x" + value.value.toString(16)
```
Solpatium, line 66: you can use hexString fn here

lib/src/main/kotlin/starknet/provider/JsonRpcMethod.kt
```diff
@@ -0,0 +1,5 @@
+package starknet.provider
+
+enum class JsonRpcMethod(val value: String) {
```
Solpatium, line 3: `val name: String`

bartekryba: Changed to 'method', as 'name' is already used in enum class and would have to be overriden

lib/src/main/kotlin/starknet/data/types/transactions.kt
```diff
@@ -24,12 +26,38 @@ enum class StarknetChainId(val value: Felt) {
...
+sealed class BlockHashOrTag() {
+    data class Hash(
+        val blockHash: Felt
+    ) : BlockHashOrTag() {
+        override fun string(): String {
+            return blockHash.hexString()
+        }
+    }
+
+    data class Tag(
```
Solpatium, line 44: Can't we place enum in here directly?

lib/src/main/kotlin/starknet/data/types/execution.kt
```diff
@@ -1,18 +1,32 @@
...
-data class ExecutionParams(
-    val nonce: Felt,
-    val maxFee: Felt,
-    val version: Felt = Felt.ZERO,
+data class CallExtraParams(
+    val blockHashOrTag: BlockHashOrTag
+)
+
+@Serializable
+data class CallContractPayload(
```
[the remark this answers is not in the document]
bartekryba: It's not. Right now we have common Provider interface for Rpc and Gateway, so the arguments in functions are the same for both as well. Gateway provider just uses it differently

lib/src/main/kotlin/starknet/data/types/transactions.kt
bartekryba: Actually not, enums cannot inherit from classes.
What I did here was an attempt to achieve algebraic data types in kotlin, as described here: https://medium.com/sharenowtech/kotlin-adt-74472319962a

lib/src/main/kotlin/starknet/provider/HttpService.kt
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+
+        fun <T : Response> send(request: Request<T>): T {
+            val client = OkHttpClient()
+            val httpRequest = buildHttpRequest(request)
+
+            val response = client.newCall(httpRequest).execute()
+
+            return processHttpResponse(response, request.deserializer)
```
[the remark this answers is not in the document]
bartekryba: Yeah, totally agree, it shouldn't. I'm working on it

lib/src/main/kotlin/starknet/data/types/transactions.kt
```diff
@@ -24,12 +26,38 @@ enum class StarknetChainId(val value: Felt) {
...
-)
+enum class BlockTag(val tag: String) {
+    LATEST("latest"),
+    PENDING("pending")
+}
 
-data class InvocationDetails(
-    val nonce: Felt?, val maxFee: Felt?, val version: Felt?
+@Serializable
+sealed class BlockHashOrTag() {
```
Solpatium, line 35: Using this with block tag requires java user to do 
```
new BlockHashOrTag.Tag(BlockTag.PENDING)
new BlockHashOrTag.Hash(new Felt(X))
```
What if we did something like this?
```
sealed class BlockHashOrTag() {
    data class Hash(
        val blockHash: Felt
    ) : BlockHashOrTag() {
        override fun string(): String {
            return blockHash.hexString()
        }
    }

    sealed class BlockTag: BlockHashOrTag() {
        abstract val tag: String
        override fun string(): String {
            return tag
        }

        object LatestBlock : BlockTag() {
            override val tag = "latest"
        }

        object PendingBlock : BlockTag() {
            override val tag = "pending"
        }

        companion object {
            @field:JvmField
            public val LATEST_BLOCK = LatestBlock

            @field:JvmField
            public val PENDING_BLOCK = LatestBlock
        }

    }


    abstract fun string(): String
}
```

It behaves like an enum, is type safe and doesn't require specifying tag twice.

lib/src/main/kotlin/starknet/service/http/HttpRequest.kt
```diff
@@ -0,0 +1,27 @@
...
+
+        return Json.decodeFromString(deserializer, response)
+    }
+
+    override fun sendAsync(): CompletableFuture<T> {
+        val future = HttpService.sendAsync(payload).thenApply { response ->
+            Json.decodeFromString(deserializer, response)
+        }
+
+        return future
```
Solpatium, line 25:
```
return HttpService.sendAsync(payload).thenApply { response ->
    Json.decodeFromString(deserializer, response)
}
````

File: lib/src/main/kotlin/starknet/data/types/execution.kt [hunk 15]
```diff
@@ -1,18 +1,38 @@
...
     val nonce: Felt,
     val maxFee: Felt,
-    val version: Felt = Felt.ZERO,
+    val version: Felt
+)
+
+@Serializable
+data class CallContractPayload(
+    val request: Call,
+    @SerialName("block_hash") val blockHashOrTag: BlockHashOrTag
```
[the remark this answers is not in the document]
bartekryba: no, rpc uses the same name for both of value types

lib/src/main/kotlin/starknet/data/types/transactions.kt
bartekryba: The reason I wanted to define BlockTag outside of BlockHashOrTag class was to be able to use it in BlockNumOrTag as well, which will be needed for some endpoints

Solpatium (approved):

PR #6, opened 2022-06-24
Add uint256

Solpatium:


lib/src/main/kotlin/starknet/data/types/Felt.kt
```diff
@@ -16,6 +18,10 @@ data class Felt(val value: BigInteger) {
         }
     }
 
+    override fun toString(): String {
```
bartekryba, line 21: We probably would like to choose between hex and dec string

bartekryba (approved):

lib/src/main/kotlin/starknet/data/types/Felt.kt
Solpatium: This is the default toString method, we can custom ones later

PR #8, opened 2022-06-30
Add pedersen implementation using native crypto-cpp package

Solpatium:
Please note that if you want to run tests using intellij you will have to add this parameter to JVM:
```
-Djava.library.path="YOUR_PATH/starknet.kt/lib/build/libs/shared/"
```
Shared library will be generated after running `gradle test`.


bartekryba:
We should add to README how to fetch crypto-cpp submodule, as well as intellij path config.

bartekryba (approved):

PR #19, opened 2022-07-12
Fix android build

Solpatium:
Also closes https://github.com/software-mansion/starknet.kt/issues/15


bartekryba:
minsdk in android demo should be also 24, to match one in android module

PR #22, opened 2022-07-13
Fix/provider

bartekryba:
Small changes and fixes of problems found while writing e2e tests. Mainly:
- Changed type of entrypoint from string to Felt
- Changed return type of getStorageAt method
- Fixed Json building in GatewayProvider


lib/src/main/kotlin/starknet/data/types/execution.kt
```diff
@@ -8,7 +8,7 @@ import starknet.data.types.Felt
 @Serializable
 data class Call(
     @SerialName("contract_address") val contractAddress: Felt,
-    val entrypoint: String,
+    val entrypoint: Felt,
```
Solpatium, line 11: Why?

lib/src/main/kotlin/starknet/data/types/transactions.kt
```diff
@@ -25,12 +29,13 @@ enum class StarknetChainId(val value: Felt) {
...
 }
 
+@Serializable
 enum class BlockTag(val tag: String) {
     LATEST("latest"),
     PENDING("pending")
 }
 
-@Serializable
+@Serializable()
```
Solpatium, line 38: Why do you call it here and not above?

lib/src/main/kotlin/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -13,7 +13,7 @@ class GatewayProvider(
     private val gatewayUrl: String,
     override val chainId: StarknetChainId
 ) : Provider {
-    private fun mapCalldataToDecimal(calldata: Calldata): JsonElement {
+    private fun mapFeltListToDecimal(calldata: List<Felt>): JsonElement {
```
Solpatium, line 16: Can we turn this into local `.toDecimals` extension to `List<Felt>`?

lib/src/main/kotlin/starknet/service/http/HttpRequest.kt
```diff
@@ -11,10 +11,16 @@ class HttpRequest<T>(
...
 ): Request<T> {
+    private val json = Json {
+        ignoreUnknownKeys = true
+    }
+
     override fun send(): T {
         val response = HttpService.send(payload)
 
-        return Json.decodeFromString(deserializer, response)
+        println(response)
```
Solpatium, line 21: 👁️

Solpatium (changes requested):

lib/src/main/kotlin/starknet/data/types/execution.kt
bartekryba: Imo it's more aligned with other imputs, as we require Felt almost everywhere. And it eliminates situation when it's unclear whether someone should type selector name or the hex string of it. But it is also less developer friendly, so let's discuss which one is better

Solpatium (approved):

Solpatium (approved):

PR #23, opened 2022-07-13
Initial docs

cptartur:
This PR adds initial work of docs by adding KDocs to most methods and classes and Dokka docs generation gradle tasks.

User guide will be added in separate PR.

### DIFF
--- a/lib/build.gradle.kts
+++ b/lib/build.gradle.kts
@@ -13,7 +13,7 @@ version = "0.0.1"
 plugins {
     // Apply the org.jetbrains.kotlin.jvm Plugin to add support for Kotlin.
     id("org.jetbrains.kotlin.jvm")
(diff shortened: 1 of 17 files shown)


lib/build.gradle.kts
```diff
@@ -6,18 +6,29 @@
...
  */
 
+import org.jetbrains.dokka.gradle.DokkaTask
+
 version = "0.0.1"
 
 plugins {
     // Apply the org.jetbrains.kotlin.jvm Plugin to add support for Kotlin.
     id("org.jetbrains.kotlin.jvm")
+    id("org.jetbrains.dokka") version "1.7.0"
```
Solpatium, line 16: Please specify plugin versions in the root.

Solpatium (approved):

PR #24, opened 2022-07-14
Add basic guide

### LINKED ISSUE #13: Create documentation
Both java classes and user guide. Having something like protostar documentation would be nice.

cptartur:
This PR adds a basic guide to the library. Closes #13.


lib/lib.md
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+```
+
+## Example usages in Kotlin
+
+```kotlin
+fun main() {
+    // Create a provider for interacting with StarkNet
+    val provider = GatewayProvider("feeder_gateway_url", "gateway_url", StarknetChainId.TESTNET)
```
Solpatium, line 37: Don't we provide those values as some constants? We probably should do that or expose provider configured for testnet already.

lib/lib.md
```diff
@@ -0,0 +1,106 @@
...
+        var provider = new GatewayProvider("feeder_gateway_url", "gateway_url", StarknetChainId.TESTNET);
+
+        // Create an account interface
+        Felt accountAddress = new Felt(45234235672347L);
+        Felt privateKey = new Felt(5326326273453L);
+        var account = new StandardAccount(provider, accountAddress, privateKey);
+
+        // Make an asynchronous request
+        Felt contractAddress = new Felt(42362362436L);
+        Felt storageKey = new Felt(13241253414L);
```
Solpatium, line 68: This is not realistic, please use `fromHex`.

Solpatium (changes requested):

lib/lib.md
cptartur: I don't think we have these declared as constants at the moment.

Solpatium: Then we should do something about it

cptartur: These has been added along with gateway client factory. Examples will need to be updated to reflect these changes.

Solpatium: Where is the factory?

cptartur: In the GatewayProvider companion object

Solpatium: Yes, please update the docs to use makeTestnetClient

lib/lib.md
```diff
@@ -0,0 +1,101 @@
+# Module lib
+
+StarkNet.kt is a library allowing for easy interaction with StarkNet gateway and nodes, including
+querying starknet state, executing transactions and deploying contracts.
+
+Although written in Kotlin, StarkNet.kt has been created with compatibility with Java in mind.
+
+## Example usages in Java
```
Solpatium, line 8: usage

Solpatium, line 8: We can structure it a little bit better:
- Asynchronous requests
  - java
  - kotlin
- Synchronous requests
  - java
  - kotlin

lib/lib.md
```diff
@@ -0,0 +1,131 @@
...
+
+        response.thenAccept(System.out::println);
+    }
+}
+```
+
+### In Kotlin
+
+It is also possible to make asynchronous requests. `Request.sendAsync()` returs a `CompletableFuture`
+that can be than handled in preferred way.
```
Solpatium, line 106: Move this at the beginning of asynchronous section.

Solpatium (approved):

PR #25, opened 2022-07-14
Add bundling shared libs with jars

Solpatium:


lib/src/main/kotlin/starknet/crypto/NativeLoader.kt
```diff
@@ -0,0 +1,33 @@
...
+            // Find the package bundled in this jar
+            val resource = sharedLibExtensions.map {
+                val fileName = "/lib$name.$it"
+                object {}::class.java.getResource(fileName)
+            }.firstOrNull { it != null } ?: throw e
+
+            val tmpDir = Files.createTempDirectory("$name-dir").toFile().apply {
+                deleteOnExit()
+            }
+            val tmpFile = File(tmpDir, name)
```
cptartur, line 27: I think it doesn't actually create a file in the filesystem, just the java object.

It should be replaced with `Files.createFile()`

lib/src/main/kotlin/starknet/crypto/NativeLoader.kt
```diff
@@ -0,0 +1,33 @@
...
+            resource.openStream().use { Files.copy(it, tmpFile.toPath()) }
```
cptartur, line 28: I think `Files.copy` would be simpler.

lib/src/main/kotlin/starknet/crypto/NativeLoader.kt
```diff
@@ -0,0 +1,33 @@
+package starknet.crypto
+
+import java.io.File
```
cptartur, line 3: I'd prefer we not use io.File at all. This is an old api. `nio.file` is newer and provides a better apis.

cptartur (changes requested):

bartekryba:
I got
`Error: Call requires API level 26 (current min is 24): java.nio.file.Files#createTempDirectory [NewApi]`
when building this

lib/src/main/kotlin/starknet/crypto/NativeLoader.kt
Solpatium: Why do I need to create the file?

lib/src/main/kotlin/starknet/crypto/NativeLoader.kt
```diff
@@ -0,0 +1,32 @@
...
+            val resource = sharedLibExtensions.map {
+                val fileName = "/lib$name.$it"
+                object {}::class.java.getResource(fileName)
+            }.firstOrNull { it != null } ?: throw e
+
+            // Android lint complains about it, but android should never reach this code
+            val tmpDir = Files.createTempDirectory("$name-dir").toFile().apply {
+                deleteOnExit()
+            }
+            val tmpFilePath = FileSystems.getDefault().getPath(tmpDir.absolutePath, name)
```
cptartur, line 26:
```suggestion
            val tmpFilePath = Paths.get(tmpDir.absolutePath, name)
```
It should work the same but is simpler. You should verify though

cptartur (commented):
I think we should create a issues for:
* Improve NativeLoader and get rid of lint ignores
* Add unit tests for NativeLoader

bartekryba (approved):

lib/src/main/kotlin/starknet/crypto/NativeLoader.kt
Solpatium: No, it doesn't work on android.

cptartur (approved):

PR #30, opened 2022-07-15
E2E tests

bartekryba:
Draft of e2e tests

DevnetClient class for running devnet and deploying contracts

Contract files in test/resources, /src for contract sources, /compiled for compiled contract files. Script for compiling all contracts in /src directory.


lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -0,0 +1,118 @@
...
+    @BeforeEach
+    fun before() {
+        devnetClient.start()
+        contractAddress = devnetClient.deployContract("providerTest")
+    }
+
+    @AfterEach
+    fun after() {
+        devnetClient.destroy()
+    }
```
Solpatium, line 46: Do we start a devnet before each test? This will make test run much much longer. I think having a single devnet is ok.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.MethodSource
+import starknet.data.selectorFromName
+import starknet.data.types.*
+import starknet.utils.DevnetClient
+import starknet.provider.gateway.GatewayProvider
+import starknet.provider.rpc.JsonRpcProvider
+
+class ProviderTest {
+    private var contractAddress: Felt? = null
```
Solpatium, line 15: You can use lazy init for that.

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -0,0 +1,62 @@
+package starknet.utils
+
+import starknet.data.types.Felt
+import java.util.concurrent.TimeUnit
+
+const val HEX_ADDRESS_LENGTH = 66
+
+class DevnetClient(val host: String = "localhost", val port: Int = 5050) {
+    private var devnetProcess: Process? = null
```
Solpatium, line 9: use lazy

Solpatium (changes requested):

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
cptartur, line 15: I think we could use `lateinit var` instead of nullable value.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -0,0 +1,118 @@
...
+
+        assertEquals(1, response.result.size)
+
+        return response.result.first()
+    }
+
+    @ParameterizedTest
+    @MethodSource("getProviders")
+    fun callContractTest(provider: Provider) {
+        // Currently not supported in devnet
```
cptartur, line 66:
```suggestion
        // FIXME Currently not supported in devnet
```
so we wont forget?

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -0,0 +1,62 @@
...
+        val result = String(deployProcess.inputStream.readAllBytes())
+
+        val searchPhrase = "Contract address: "
+
+        val addressStart = result.indexOf(searchPhrase) + searchPhrase.length
+        val address = result.substring(addressStart, addressStart + HEX_ADDRESS_LENGTH)
+
+        return Felt.fromHex(address)
+    }
+}
```
cptartur, line 62: Newline

lib/src/main/kotlin/starknet/data/types/transactions.kt
```diff
@@ -56,6 +58,26 @@ sealed class BlockHashOrTag() {
...
+class BlockHashOrTagSerializer(): KSerializer<BlockHashOrTag> {
+    override fun deserialize(decoder: Decoder): BlockHashOrTag {
+        val value = decoder.decodeString()
+
+        try {
+            val tag = BlockTag.valueOf(value)
+            return BlockHashOrTag.Tag(tag)
+        } catch (_: Exception) { }
+
+        return BlockHashOrTag.Hash(Felt.fromHex(value))
```
cptartur, line 70: I'm not a fan of using try catch like this, but I'm unsure how this could be improved.

Solpatium:
```
return try {
   val tag = BlockTag.valueOf(value)
   BlockHashOrTag.Tag(tag)
} catch (_: Exception) {
   BlockHashOrTag.Hash(Felt.fromHex(value))
}
```

PR #36, opened 2022-07-21
Add transaction related methods

### LINKED ISSUE #32: Add transaction related methods
Add getTransaction and getTransactionReceipt to provider interface and implementations.

cptartur:
Closes #32

### DIFF
--- a/lib/src/main/kotlin/starknet/data/types/execution.kt
+++ b/lib/src/main/kotlin/starknet/data/types/execution.kt
@@ -3,7 +3,6 @@ package starknet.data.types
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import starknet.data.selectorFromName
-import starknet.data.types.Felt
 
 @Serializable
 data class Call(
@@ -37,8 +36,20 @@ data class CallContractPayload(
 @Serializable
 data class GetStorageAtPayload(
     @SerialName("contract_address") val contractAddress: Felt,
-    val key: Felt,
+    @SerialName("key") val key: Felt,
     @SerialName("block_hash") val blockHashOrTag: BlockHashOrTag
+)
+
+@Serializable
+data class GetTransactionByHashPayload(
+    @SerialName("transaction_hash")
+    val transactionHash: Felt
+)
+
+@Serializable
+data class GetTransactionReceiptPayload(
+    @SerialName("transaction_hash")
+    val transactionHash: Felt
 )
 
 fun callsToExecuteCalldata(calls: List<Call>, nonce: Felt): List<Felt> {
@@ -53,7 +64,7 @@ fun callsToExecuteCalldata(calls: List<Call>, nonce: Felt): List<Felt> {
         wholeCalldata.addAll(call.calldata)
     }
 
-    return buildList() {
+    return buildList {
         add(Felt(calls.size))
         addAll(callArray)
         add(Felt(wholeCalldata.size))
--- a/lib/src/main/kotlin/starknet/provider/rpc/JsonRpcProvider.kt
+++ b/lib/src/main/kotlin/starknet/provider/rpc/JsonRpcProvider.kt
@@ -2,6 +2,9 @@ package starknet.provider.rpc
 
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.json.*
+import starknet.data.responses.Transaction
+import starknet.data.responses.TransactionReceipt
+import starknet.data.responses.serializers.JsonRpcTransactionPolymorphicSerializer
 import starknet.data.types.*
 import starknet.provider.Provider
 import starknet.provider.Request
@@ -78,6 +81,20 @@ class JsonRpcProvider(
         val payload = GetStorageAtPayload(contractAddress, key, BlockHashOrTag.Hash(blockHash))
 
         return getStorageAt(payload)
+    }
+
+    override fun getTransaction(transactionHash: Felt): Request<Transaction> {
+        val payload = GetTransactionByHashPayload(transactionHash)
+        val params = Json.encodeToJsonElement(payload)
+
+        return buildRequest(JsonRpcMethod.GET_TRANSACTION_BY_HASH, params, JsonRpcTransactionPolymorphicSerializer)
+    }
+
+    override fun getTransactionReceipt(transactionHash: Felt): Request<TransactionReceipt> {
+        val payload = GetTransactionReceiptPayload(transactionHash)
+        val params = Json.encodeToJsonElement(payload)
+
+        return buildRequest(JsonRpcMethod.GET_TRANSACTION_RECEIPT, params, TransactionReceipt.serializer())
     }
 
     override fun invokeFunction(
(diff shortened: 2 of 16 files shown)


lib/src/main/kotlin/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -2,6 +2,9 @@ package starknet.provider.rpc
 
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.json.*
+import starknet.data.responses.Transaction
+import starknet.data.responses.TransactionReceipt
+import starknet.data.responses.serializers.JsonRpcTransactionPolymorphicSerializer
```
bartekryba, line 7: I think you forgot to add this file

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -1,7 +1,9 @@
 package starknet.utils
 
 import starknet.data.types.Felt
+import java.nio.file.Path
 import java.util.concurrent.TimeUnit
+import kotlin.io.path.absolutePathString
 
 const val HEX_ADDRESS_LENGTH = 66
```
bartekryba, line 8: It can be removed now

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -51,12 +55,67 @@ class DevnetClient(val host: String = "localhost", val port: Int = 5050) {
         deployProcess.waitFor()
 
         val result = String(deployProcess.inputStream.readAllBytes())
+        val lines = result.lines()
+        return getTransactionResult(lines)
+    }
+
+    fun declareContract(contractPath: Path): TransactionResult {
```
bartekryba, line 62: Do you think it's better to take whole Path as a parameter or just a name, and add path inside this method?

lib/src/test/resources/compiled/providerTestAbi.json
```diff
@@ -0,0 +1,25 @@
...
+    "outputs": [
+      {
+        "name": "res",
+        "type": "felt"
+      }
+    ],
+    "stateMutability": "view",
+    "type": "function"
+  }
+]
```
bartekryba, line 25: Please add a flag to starknet-compile in compileContracts.sh to generate abi files

lib/src/main/kotlin/starknet/data/types/execution.kt
```diff
@@ -53,7 +64,7 @@ fun callsToExecuteCalldata(calls: List<Call>, nonce: Felt): List<Felt> {
         wholeCalldata.addAll(call.calldata)
     }
 
-    return buildList() {
+    return buildList {
         add(Felt(calls.size))
         addAll(callArray)
         add(Felt(wholeCalldata.size))
```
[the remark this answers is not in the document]
cptartur: `callArray` and `wholeCalldata` are list, not single elements and it doesn't play nicely with `listOf`.

lib/src/main/kotlin/starknet/data/types/execution.kt
```diff
@@ -37,10 +36,22 @@ data class CallContractPayload(
 @Serializable
 data class GetStorageAtPayload(
     @SerialName("contract_address") val contractAddress: Felt,
-    val key: Felt,
+    @SerialName("key") val key: Felt,
```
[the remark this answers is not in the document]
cptartur: I added it just for the sake of uniformity with the rest of parameters.

lib/src/main/kotlin/starknet/provider/rpc/JsonRpcProvider.kt
cptartur: Good catch, I'll add it.

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
cptartur: This is weird. because it needs to be imported even though it looks like a method on Path object. It seems like its because `Path` is `java.nio` object and absolutePathString is a kotlin inline function.

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
cptartur: I think it'd be better to take a whole path: It makes the whole method be more flexible. I guess `deployContract` could be changed to be uniform with `declareContract`. What do you think?

lib/src/main/kotlin/starknet/service/http/HttpService.kt
```diff
@@ -12,7 +12,13 @@ import java.util.concurrent.CompletableFuture
...
  */
 class HttpService {
-    data class Payload(val url: String, val method: String, val params: List<String>, val body: String?)
+    data class Payload(val url: String, val method: String, val params: List<String>, val body: String?) {
+        constructor(url: String, method: String, params: List<String>): this(url, method, params, null)
+
+        constructor(url: String, method: String, body: String) : this(url, method, emptyList(), body)
+
+        constructor(url: String, method: String): this(url, method, emptyList(), null)
+    }
```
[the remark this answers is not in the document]
cptartur: I think it would make more sense for me to use params than to build one large url from them.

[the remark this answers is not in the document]
cptartur: Refactored the client so it now uses params.

lib/src/test/resources/compiled/providerTestAbi.json
cptartur: Should we also add contents of `compiled/` to gitignore. Or do we want to commit compiled contracts and abis?

lib/src/main/kotlin/starknet/data/types/execution.kt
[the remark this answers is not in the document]
bartekryba: ahh sorry, I didn't see it

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
bartekryba: actually I meant `const val HEX_ADDRESS_LENGTH = 66` as it's not used now

lib/src/test/resources/compiled/providerTestAbi.json
bartekryba: I was also thinking about it, I'm not sure though. Adding it to git eliminates risk of using wrong version of the contract, but it unnecessarily makes project size bigger.

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
bartekryba: Yeah, we can do it this way

lib/src/test/resources/compiled/providerTestAbi.json
cptartur: In starknet.py we mostly keep only cairo source codes and not the compiled versions. We do have a logic for compiling contracts inside our package so it's not exactly 1:1 situation.

bartekryba (approved):

PR #39, opened 2022-07-22
Add CI

cptartur:


.github/workflows/checks.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Install devnet
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+
+      - name: Run tests
+        run: ./gradlew :lib:test --stacktrace --info
```
cptartur, line 39: I wonder if we need `--stacktrace` and `--info` in the actual setup. I've added this as part of the debugging, but maybe it makes sense to keep them?

crypto/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ set(JAVA_AWT_LIBRARY NotNeeded)
 set(JAVA_JVM_LIBRARY NotNeeded)
 set(JAVA_AWT_INCLUDE_PATH NotNeeded)
 set(JAVA_INCLUDE_PATH "$ENV{JAVA_HOME}/include")
-set(JAVA_INCLUDE_PATH2 "$ENV{JAVA_HOME}/include/darwin")
+set(JAVA_INCLUDE_PATH2 "$ENV{JAVA_HOME}/include/darwin;$ENV{JAVA_HOME}/include/linux")
```
cptartur, line 28: Ideally we would also support Windows but I guess it can be added along with the checks for platforms other than MacOS

PR #41, opened 2022-07-25
Feat/32bit support

4rgorok:


crypto/CMakeLists.txt
```diff
@@ -7,20 +7,38 @@ project(StarkwareCryptoLib VERSION 0.1.0 LANGUAGES CXX)
...
 # For Windows: Prevent overriding the parent project's compiler/linker settings
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
 FetchContent_MakeAvailable(googletest)
 
+FetchContent_GetProperties(boostorg)
+if (NOT boostorg_POPULATED)
+    FetchContent_Populate(boostorg)
+endif ()
+
+include_directories(${boostorg_SOURCE_DIR})
```
Solpatium, line 30: Is it possible to fetch it conditionally based on compilation target?

4rgorok: Yes, I've implemented it in newest commit.

crypto/CMakeLists.txt
```diff
@@ -16,11 +18,31 @@ FetchContent_Declare(
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
 FetchContent_MakeAvailable(googletest)
 
+if(NOT "${CMAKE_GENERATOR}" MATCHES "(Win64|IA64)")
```
Solpatium, line 21: Please add a comment here explaining the logic.

4rgorok: I've added comments and used a better method in the newest commit.

Solpatium (approved):

PR #48, opened 2022-07-27
Add support deploy and declaring contracts

### LINKED ISSUE #17: Support deploying contracts

cptartur:
Closes #17 

I've had to add some temporary fixes to the RPC provider due to a minor bug in devnet. Once [this PR ](https://github.com/Shard-Labs/starknet-devnet/pull/197) is merged, we remove them.


lib/src/main/kotlin/starknet/data/types/transactions.kt
```diff
@@ -79,12 +82,44 @@ class BlockHashOrTagSerializer() : KSerializer<BlockHashOrTag> {
...
 )
 
+data class ContractDefinition(
+    val contract: String
+) {
+    fun parseContract(): Triple<JsonElement, JsonElement, JsonElement> {
+        val compiledContract = Json.parseToJsonElement(this.contract).jsonObject
+        val program = compiledContract["program"]!!
+        val entryPointsByType = compiledContract["entry_points_by_type"]!!
+        val abi = compiledContract["abi"] ?: JsonArray(emptyList())
```
Solpatium, line 103: We have to ensure those keys are present in the json.

lib/src/main/kotlin/starknet/data/types/transactions.kt
```diff
@@ -79,12 +82,44 @@ class BlockHashOrTagSerializer() : KSerializer<BlockHashOrTag> {
...
+        return Triple(program, entryPointsByType, abi)
```
Solpatium, line 104: Use data class instead or even better - make this a constructor for `ContractDefinition` with those fields (don't have to be public).

lib/src/main/kotlin/starknet/extensions/json.kt
```diff
@@ -0,0 +1,10 @@
+package starknet.extensions
+
+import kotlinx.serialization.json.*
+import starknet.data.types.Felt
+
+fun JsonObjectBuilder.putFeltAsHex(key: String, value: Felt?): JsonElement? =
+    put(key, JsonPrimitive(value?.hexString() ?: Felt.ZERO.hexString()))
+
+fun JsonArrayBuilder.addFeltAsHex(value: Felt?): Boolean =
+    add(JsonPrimitive(value?.hexString() ?: Felt.ZERO.hexString()))
```
Solpatium, line 10: Is it more readable than value.hexString? We could turn hexString into a property.
```
put(key, value.hexString)
```
vs
```
putFeltAsHex(key, value)
```
I see you use a fallback for value, but what is the use case? Is it safe for us to fallback to 0x0? I am afraid in some cases it might be not what we want.

lib/src/main/kotlin/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -105,21 +108,69 @@ class GatewayProvider(
...
+            putJsonObject("contract_definition") {
+                put("program", compressedProgram)
+                put("entry_points_by_type", entryPointsByType)
+                put("abi", abi)
+            }
+            putFeltAsHex("version", payload.version)
+        }
+
+        val httpPayload = HttpService.Payload(url, "POST", body.toString())
+        return HttpRequest(httpPayload, DeployResponse.serializer())
```
Solpatium, line 148: We repeat those 2 lines everywhere. Can we replace it with
```
private fun makeRequest<T>(url, body, serializer, method = "POST")
```
?

Solpatium (changes requested):

lib/src/main/kotlin/starknet/data/types/responses.kt
```diff
@@ -18,6 +20,24 @@ data class InvokeFunctionResponse(
     @SerialName("transaction_hash") val transactionHash: Felt
 ) : Response()
 
+@Serializable
+data class DeployResponse @OptIn(ExperimentalSerializationApi::class) constructor(
```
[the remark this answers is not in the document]
cptartur: `@JsonNames` annotation comes from the experimental serialization api and we need to explicitly opt in to use it.

lib/src/main/kotlin/starknet/data/types/transactions.kt
cptartur: I can add some validation if you think that's better. However with the current RPC and Gateway responses, program and entryPointsByType are always present.

lib/src/main/kotlin/starknet/extensions/json.kt
cptartur: I guess we can just not support nullable values in these methods and we won't need to have a fallback value.

cptartur: And I think it is just a little bit more readable since it makes result of the function call immediately obvious unlike the cast

lib/src/main/kotlin/starknet/data/types/contracts.kt
```diff
@@ -2,39 +2,72 @@
...
 
+@Serializable
 data class StructAbi(
     val name: String,
 ) : AbiElement()
 
+@Serializable
 data class Abi(
     val values: List<AbiElement>
 )
```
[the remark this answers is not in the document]
cptartur: Hmm, I guess they carried over from before this PR and I just added `@Serializable` when I though Ill be using them. I can remove these if you think they're not needed at the momet.

lib/src/main/kotlin/starknet/data/types/transactions.kt
Solpatium: Please add it

lib/src/main/kotlin/starknet/data/types/responses.kt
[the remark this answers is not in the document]
Solpatium: Makes sense, just please include an explanation in the code.

lib/src/main/kotlin/starknet/service/http/HttpRequest.kt
```diff
@@ -25,4 +25,37 @@ class HttpRequest<T>(
             Json.decodeFromString(deserializer, response)
         }
     }
+
+    companion object {
```
Solpatium, line 29: Can we not expose it in public interface?

lib/src/main/kotlin/starknet/data/types/transactions.kt
Solpatium: Users will get an error that is more meaningful than NullPointerException - everybody hates that

lib/src/main/kotlin/starknet/extensions/json.kt
Solpatium: I don't see why value.hexString would be less readable in terms of the output. 
Do we need to use decimals for felts somehwere? If not then what if we added a `put` extension that accepts Felt and puts hex in? This way we extend offered native api.

lib/src/main/kotlin/starknet/service/http/HttpRequest.kt
```diff
@@ -25,4 +25,37 @@ class HttpRequest<T>(
...
+        fun <T> makeRequest(
+            url: String,
+            method: String,
+            params: List<Pair<String, String>>,
+            body: String,
+            deserializer: DeserializationStrategy<T>
+        ): HttpRequest<T> {
+            val httpPayload = HttpService.Payload(url, method, params, body)
+            return HttpRequest(httpPayload, deserializer)
+        }
```
Solpatium, line 59: I don't like all those methods, they add unnecesssary cluster. I think the thing that we need is to accept JSON object in Payload's secondary constructor. This way we can do:
```
return HttpRequest(Payload(url, "POST", body), InvokeFunctionResponse.serializer())
```

lib/src/main/kotlin/starknet/data/types/responses.kt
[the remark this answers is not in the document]
cptartur: There are multiple data classes with this opt in, where should the information be included?

lib/src/main/kotlin/starknet/data/types/transactions.kt
cptartur: Added basic validation and a custom exception for that.

lib/src/main/kotlin/starknet/extensions/json.kt
cptartur: I mean, we can just name it `put`. That's what I initially did but then changed it to make it more obvious. It doesn't change much on the user side, since they have to import this function anyway, no matter if it's named `put` or not.

Solpatium: I think we should either name it as native methods are named or not add it at all.

lib/src/main/kotlin/starknet/data/types/responses.kt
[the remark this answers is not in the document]
Solpatium: I think you can add it to the first usage within a file.

Solpatium (approved):

PR #49, opened 2022-07-29
Change name in readme

Solpatium:


README.md
```diff
@@ -1,5 +1,11 @@
-# starknet.kt
-StarkNet SDK for Kotlin
+<h2 align="center">☕ starknet jvm ☕</h2>
+
+[StarkNet](https://starkware.co/starknet/#:~:text=Live%20on%20Mainnet-,What%20is%20StarkNet%3F,-StarkNet%20is%20a) SDK for JVM languages:
+- Java
+- Kotlin
+- Scala
+- Clojure
+- Groovy
```
cptartur, line 8: I'm not sure if we should include languages we haven't done any testing agains. Or perhaps you've already verified our library works correctly in these?

cptartur (changes requested):

README.md
Solpatium: They have java inoperability. It's enough to make sure java version works fine. We have to include them here to make sure typing "Starknet LANG sdk" will make it possible to find our library.

cptartur: Alright, I'll trust you with this one.

cptartur (approved):

Issue #42, opened 2022-07-25
Add missing transaction properties.


bartekryba:
Make sure that all transaction properties that are present in starknet-spec are also present in our transaction data classes.

My proposal for working with both gateway and rpc:
* fields present in gateway, but not in rpc => Ignore them
* fields present in rpc but not in gateway => Make them non-optional (if they are so in the spec) and use some dummy value for them (like 0 for felts)

cptartur:
Dummy values is probably the most convenient solution for the RPC users but it can be a little bit confusing. I'd definitely rule out creating subclasses for gateway responses, so unfortunately it's either this or optional values.

What do you think @Solpatium

Solpatium:
> I'd definitely rule out creating subclasses for gateway responses

Why? This creates the best (type safe) UX for users.

cptartur:
I think maintaining different data classes for RPC and different for gateway will just add unnecessary complication plus it's not obvious how they would be related which would complicate the provider interface. If gateway is to be gone relatively soon I believe it's better if we only had data classes modeled after rpc where possible and somehow map gateway responses to them.

Solpatium:
> it's not obvious how they would be related

They would extend the same interface. 

I think we can start with things with common properties first.

cptartur:
Closed in #50

PR #50, opened 2022-07-29
Add missing parameters to data classses

cptartur:
Initial work on resolving #42.

Adding this as a draft since it's very much work in progres and also devnet will need to be updated to the newest rpc spec for us to test these changes.

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/responses/TransactionReceipt.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/responses/TransactionReceipt.kt
@@ -58,7 +58,7 @@ data class GatewayTransactionReceipt(
     @JsonNames("transaction_failure_reason")
     override val rejectionReason: String? = null,
 
-    override val type: TransactionReceiptType = TransactionReceiptType.GATEWAY
+    override val type: TransactionReceiptType = TransactionReceiptType.GATEWAY,
 ) : AcceptedTransactionReceipt()
 
 @OptIn(ExperimentalSerializationApi::class)
@@ -92,7 +92,7 @@ data class InvokeTransactionReceipt(
     @JsonNames("status_data")
     override val rejectionReason: String? = null,
 
-    override val type: TransactionReceiptType = TransactionReceiptType.INVOKE
+    override val type: TransactionReceiptType = TransactionReceiptType.INVOKE,
 ) : AcceptedTransactionReceipt()
 
 @OptIn(ExperimentalSerializationApi::class)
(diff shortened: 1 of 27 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/responses/TransactionReceipt.kt
```diff
@@ -6,22 +6,154 @@ import kotlinx.serialization.ExperimentalSerializationApi
...
+}
+
+@OptIn(ExperimentalSerializationApi::class)
+@Serializable
+data class GatewayTransactionReceipt(
+    @JsonNames("events")
+    val events: List<Event>,
+
+    @JsonNames("l2_to_l1_messages")
+    val messageToL1: List<MessageToL1>,
```
bartekryba, line 32: Let's change it to plural

lib/src/main/kotlin/com/swmansion/starknet/data/responses/TransactionReceipt.kt
```diff
@@ -6,22 +6,154 @@ import kotlinx.serialization.ExperimentalSerializationApi
...
+    abstract val hash: Felt
+    abstract val actualFee: Felt
+}
+
+@Serializable
+sealed class TransactionReceipt : CommonTransactionReceipt() {
+    abstract override val hash: Felt
+    abstract override val actualFee: Felt
+    abstract val status: TransactionStatus
+    abstract val rejectionReason: String?
```
bartekryba, line 20: In json rpc docs, it is called statusData, is there a chance it can be not null, when status is different than `rejected`?

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -42,6 +43,16 @@ class ProviderTest {
...
+        @JvmStatic
+        private fun gatewayProvider(): GatewayProvider {
+            return GatewayProvider(devnetClient.feederGatewayUrl, devnetClient.gatewayUrl, StarknetChainId.TESTNET)
+        }
+
+        @JvmStatic
+        private fun rpcProvider(): JsonRpcProvider {
+            return JsonRpcProvider(devnetClient.rpcUrl, StarknetChainId.TESTNET)
+        }
+
```
bartekryba, line 55: We can now use these two methods in getProviders() method

bartekryba (commented):
The data structures look good.

bartekryba (changes requested):
I think you might've forgot to add `JsonRpcTransactionReceiptPolymorphic` file

lib/src/main/kotlin/com/swmansion/starknet/data/responses/TransactionReceipt.kt
cptartur: Honestly I'm not sure. I don't think it'd make sense to have extra status data if transaction just got accepted but it's not very explicit from the spec.

Solpatium (approved):

bartekryba (approved):

PR #51, opened 2022-08-01
Refactor file structure

### LINKED ISSUE #34: Rename to starknet jvm
Current name might not be obvious for java users, it's better to add a jvm name. Also reorganise package names.

cptartur:
Closes #34


cptartur:
There's a problem with android lint related to the usage of `BigInteger.intValueExact`. This is however only used due to the bug in devnet. Once the fix is merged all usages of `intValueExact` can be removed.

lib/src/main/kotlin/com/swmansion/starknet/extensions/json.kt
```diff
@@ -0,0 +1,9 @@
+package com.swmansion.starknet.extensions
+
+import com.swmansion.starknet.data.types.Felt
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObjectBuilder
+import kotlinx.serialization.json.JsonPrimitive
+
+fun JsonObjectBuilder.put(key: String, value: Felt?): JsonElement? =
+    put(key, JsonPrimitive(value?.hexString() ?: Felt.ZERO.hexString()))
```
Solpatium, line 9: Do we want to use the default value here?

cptartur: Perhaps we could change this method to only accept non null Felts?

Solpatium (approved):

PR #61, opened 2022-08-04
Minor improvements to the package

cptartur:
This PR adds some unrelated minor improvements to the package that felt like they didn't necessitate their own PRs.

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/NetUrls.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/NetUrls.kt
@@ -1,5 +1,3 @@
-@file:JvmName("NetUrls")
-
 package com.swmansion.starknet.data
 
 object NetUrls {
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/NetUrls.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/NetUrls.kt
@@ -1,5 +1,3 @@
-@file:JvmName("Nets")
-
 package com.swmansion.starknet.data
 
 object NetUrls {
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/NetUrls.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/NetUrls.kt
@@ -1,3 +1,5 @@
+@file:JvmName("Nets")
+
 package com.swmansion.starknet.data
 
 object NetUrls {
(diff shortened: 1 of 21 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/SelectorFromName.kt
```diff
@@ -1,5 +1,3 @@
-@file:JvmName("Selector")
-
```
bartekryba, line 2: Removing it makes it not Java friendly. Now it is used like this:

`SelectorFromNameKt.selectorFromName(x)`

bartekryba:
The rest looks good

lib/src/main/kotlin/com/swmansion/starknet/data/SelectorFromName.kt
cptartur: Oh, if it includes the `Kt` at the end of each package, should we be adding JvmName annotation to all files?

lib/src/main/kotlin/com/swmansion/starknet/data/NetUrls.kt
```diff
@@ -1,5 +1,3 @@
-@file:JvmName("Nets")
```
Solpatium, line 1: Same here

lib/src/main/kotlin/com/swmansion/starknet/data/SelectorFromName.kt
```diff
@@ -1,4 +1,4 @@
-@file:JvmName("Selector")
```
Solpatium, line 1: What if we named this file `Selector`?

Solpatium (approved):
Just replace SelectorFromName to Selector (both file and the jvm name)

lib/src/main/kotlin/com/swmansion/starknet/data/SelectorFromName.kt
cptartur: Then ktLint would complain

lib/src/main/kotlin/com/swmansion/starknet/data/NetUrls.kt
bartekryba: Actually it works normally, it's better to remove this.

lib/src/main/kotlin/com/swmansion/starknet/data/SelectorFromName.kt
bartekryba: Not to all files. I'm not 100% sure how all of it works, but if the file contains only classes, interfaces and objects, this annotation is not needed. However, when there are some static functions which we would like to expose, we should add the annotation to avoid Kt suffix.

bartekryba (approved):
Approved, I would just also remove `@file:JvmName("NetUrls")`

PR #62, opened 2022-08-04
Minor tests improvements

cptartur:
This PR restores one test and makes minor improvements to the rest.


cptartur:
Interaction between tests causes a failure, perhaps we should consider some mechanism for test to request their own contract deployments so modifications to the state between tests doesn't cause failures like this one.

bartekryba (approved):

bartekryba:
Ah this problem is actually based on my assumption, that the order of test execution is the same as the order of the test cases in the class. Yeah, we can either provide a mechanism to perform some of the tests in isolation, or we can manually enforce order of the tests.

Solpatium (approved):

PR #70, opened 2022-08-10
Restore some of the disabled tests

### LINKED ISSUE #59: Restore disabled tests once devnet rpc is updated
There are severa tests marked as fixme that are disabled that will need to be restored once the rpc version in devnet is updated to the newest release spec.

cptartur:
This PR closes #59. There are some tests that remain commented out or disabled for the rpc client because of the devnet limitations. For these more descriptive comments has been added.

Tests that weren't restored subject methods that are not implemented at all and I think should be left for post the initial release.

### DIFF
--- a/lib/src/test/kotlin/starknet/provider/ProviderTest.kt
+++ b/lib/src/test/kotlin/starknet/provider/ProviderTest.kt
@@ -39,6 +39,14 @@ class ProviderTest {
         private fun rpcProvider(): JsonRpcProvider = JsonRpcProvider(devnetClient.rpcUrl, StarknetChainId.TESTNET)
 
         @JvmStatic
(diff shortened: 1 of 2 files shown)


lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -252,7 +255,7 @@ class ProviderTest {
     @ParameterizedTest
     @MethodSource("getProviders")
     fun `get class hash at block number`(provider: Provider) {
-        // FIXME: Currently not supported in devnet
+//        // Devnet only support's "latest" as block id in this method
```
Solpatium, line 258: Remove double // here

Solpatium (approved):

PR #75, opened 2022-08-11
Feat/demos

Solpatium:


lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -121,9 +121,9 @@ class GatewayProvider(
             putJsonArray("calldata") {
                 payload.invocation.calldata.toDecimal().forEach { add(it) }
             }
-            put("max_fee", payload.maxFee?.hexString())
+            put("max_fee", payload.maxFee.hexString())
             putJsonArray("signature") {
-                payload.signature?.toDecimal()?.forEach { add(it) }
+                payload.signature.toDecimal().forEach { add(it) }
```
Solpatium, line 126: These checks are not needed.

android/build.gradle
```diff
@@ -57,7 +57,6 @@ dependencies {
     testImplementation 'junit:junit:4.13.2'
     androidTestImplementation 'androidx.test.ext:junit:1.1.3'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.4.0'
-    implementation("com.swmansion.starknet:starknet:0.0.1")
```
cptartur, line 60: Why was this removed?

Solpatium: Because it was a mistake. Android is the lib for android ecosystem. This should be added in androiddemo package

cptartur (approved):

Issue #56, opened 2022-08-02
Add missing account methods.


bartekryba:
Add:
* `execute`
* `estimateFee`
* `getNonce`

methods to Account class and necessary methods for that to Provider.

PR #82, opened 2022-08-22
Feat/account methods

bartekryba:
Draft of the new account methods #56.

Main problem to consider here:
Provider returns Requests, that can be called both synchronously, or asynchronously as futures. Provider's methods however make only one call. Both estimateFee and execute in Account make multiple calls. I was wondering if we would like to align account methods interface with the provider methods interface, i.e. also use Request. But this would require rethinking our Request model a bit and modifying it.


bartekryba:
Relates to #56 and #84

lib/src/main/kotlin/com/swmansion/starknet/crypto/FeeUtils.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import kotlin.math.roundToInt
+
+object FeeUtils {
+    @JvmStatic
+    fun estimatedFeeToMaxFee(fee: Long, overhead: Double = 0.5): Long {
+        val overheadPercent = ((1 + overhead) * 100).roundToInt()
+
+        return fee * overheadPercent
+    }
```
Solpatium, line 11: Shouldn't we do `((1 + overhead) * fee).roundToInt()`? Right now for 0.5 we get 150*fee - 150 times the fee, not 1.5.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    @ParameterizedTest
+    @MethodSource("getAccounts")
+    fun `get nonce test`(account: Account) {
+        val nonce = account.getNonce()
+
+        assertEquals(nonce, Felt.ZERO)
+    }
```
cptartur, line 83: Isn't this depended on test execution order?

cptartur: Also it should be `assertEquals(expected, actual)` not the other way around.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -411,4 +411,31 @@ class ProviderTest {
...
         }
     }
+
+    @ParameterizedTest
+    @MethodSource("getProviders")
+    fun `estimate fee`(provider: Provider) {
+        // FIXME: Not yet supported in Devnet RPC
+        if (provider is JsonRpcProvider) {
+            return
+        }
```
cptartur, line 421: I believe should be supporting fee estimation. If it's not working for you this could be a bug either in devnet or in this pr.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -411,4 +411,31 @@ class ProviderTest {
...
+
+        val dummySig = listOf(Felt(0), Felt(0), Felt(0), Felt(0), Felt(0))
+
+        val invokeTx = TransactionFactory.makeInvokeTransaction(
```
cptartur, line 425: 👍 👍 👍

cptartur (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/crypto/FeeUtils.kt
bartekryba: Yeah, absolutely xd fixed it

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
bartekryba: Ah yeah, it is.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -101,7 +98,22 @@ data class InvokeTransaction(
     override val nonce: Felt = Felt.ZERO,
 
     override val type: TransactionType = TransactionType.INVOKE,
-) : Transaction()
+) : Transaction() {
+    fun toPayload(): InvokeFunctionPayload {
```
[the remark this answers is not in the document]
bartekryba: Yes, we use it in StandardAccount

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -45,3 +45,16 @@ data class TransactionFailureReason(
...
+@Serializable
+data class EstimateFeeResponse(
+    @JsonNames("gas_consumed", "gas_usage")
+    val gasConsumed: Long,
+
+    @JsonNames("gas_price")
+    val gasPrice: Long,
+
+    @JsonNames("overall_fee")
+    val overallFee: Long,
```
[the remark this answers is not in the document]
bartekryba: Changed it to felt

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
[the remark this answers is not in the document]
bartekryba: Made it internal

lib/src/main/kotlin/com/swmansion/starknet/crypto/FeeUtils.kt
```diff
@@ -0,0 +1,15 @@
+package com.swmansion.starknet.crypto
+
+import com.swmansion.starknet.data.types.Felt
+import java.math.BigInteger
+import kotlin.math.roundToInt
+
+internal object FeeUtils {
+    @JvmStatic
+    fun estimatedFeeToMaxFee(fee: Felt, overhead: Double = 0.5): Felt {
```
cptartur, line 9: I wonder if default overhead of `0.5` is a good choice. In starknet.py we're using 0.1 i believe.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -45,3 +47,51 @@ data class TransactionFailureReason(
...
+    val gasConsumed: Felt,
+
+    @JsonNames("gas_price")
+    val gasPrice: Felt,
+
+    @JsonNames("overall_fee")
+    val overallFee: Felt,
+)
+
+@OptIn(ExperimentalSerializationApi::class)
```
cptartur, line 64: Why exactly is a custom serializer needed for this class?

By the way, if you just want to modify the json object for serialization, you can use JsonTransformingSerializer, it's a little easier to use then full fledged custom serializers.

Solpatium:
We need to set target for this branch to development branch

Solpatium (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/crypto/FeeUtils.kt
```diff
@@ -0,0 +1,15 @@
...
+        val multiplier = ((1 + overhead) * 100).roundToInt().toBigInteger()
+        val result = fee.value.multiply(multiplier).divide(BigInteger.valueOf(100))
```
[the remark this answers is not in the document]
cptartur: Fee doesn't define `*` operator, and `BigInteger.multiply` and `.divide` expects another BigInteger as parameter.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
cptartur: Needed because of conversions from numeric values (not necessarily ints or longs) to felts.

PR #83, opened 2022-08-23
Provider errors and services refactor

### LINKED ISSUE #58: Create error data classes for Provider

cptartur:
This PR resolves #58 and introduces a major refactor to `HttpService`: stopped using singleton design pattern, separated several parts of `services` package into injected dependencies to facilitate simpler testing and mocking in the future.


lib/src/main/kotlin/com/swmansion/starknet/service/RequestProcessingService.kt
```diff
@@ -0,0 +1,52 @@
+package com.swmansion.starknet.service
+
+import kotlinx.serialization.json.JsonObject
+import java.util.concurrent.CompletableFuture
+
+internal interface RequestProcessingService {
+    data class Payload(
```
Solpatium, line 7: This is a Http Payload, it doesn't make sense for other protocols.

cptartur: What if we renamed this interface to indicated it's used for http communication? Since I don't think it's useful at this point to create intermediate implementation for Request payloads.

lib/src/main/kotlin/com/swmansion/starknet/service/http/OkhttpHttpService.kt
```diff
@@ -0,0 +1,96 @@
...
+import okhttp3.OkHttpClient
+import okhttp3.RequestBody.Companion.toRequestBody
+import okhttp3.Response
+import java.io.IOException
+import java.util.concurrent.CompletableFuture
+
+/**
+ * Service for making http requests.
+ */
+internal class OkhttpHttpService(private val errorHandler: HttpErrorHandler) : HttpService {
```
Solpatium, line 18: Why do we need to inject error handling here?

Solpatium: Can't we inject (optionally) OkHttpClient instead to check handling of invalid requests?

lib/src/main/kotlin/com/swmansion/starknet/service/http/OkhttpHttpService.kt
```diff
@@ -0,0 +1,96 @@
...
+        return urlBuilder.build().toString()
+    }
+
+    private fun processHttpResponse(response: Response): String {
+        val responseBody = response.body
+
+        if (response.isSuccessful) {
+            return responseBody?.string() ?: ""
+        } else {
+            errorHandler.handle(response.body?.string() ?: "Unknown error")
```
Solpatium, line 48: We should save http response code and the raw payload.

Solpatium (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -53,7 +60,47 @@ class GatewayProvider(
...
+
+        if (!response.isSuccessful || response.body == null) {
+            // FIXME add more descriptive error
+            throw RequestFailedException("Request failed")
+        }
+
+        return response.body
+    }
+
+    private fun <T> buildHttpResponseDeserializer(deserializationStrategy: DeserializationStrategy<T>): HttpRequestDeserializer<T> {
```
Solpatium, line 97: `= {response -> }`

Solpatium, line 97: we can just name this just `buildDeserializer`

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -53,7 +60,47 @@ class GatewayProvider(
...
+                throw GatewayRequestFailedException(deserializedError.code, deserializedError.message)
+            } catch (e: SerializationException) {
+                throw RequestFailedException(response.body)
+            }
+        }
+
+        if (!response.isSuccessful || response.body == null) {
+            // FIXME add more descriptive error
+            throw RequestFailedException("Request failed")
+        }
```
Solpatium, line 92: Use a better error and you can move it to the top. This way this becomes a guard that ensures the response is successful and there is a body.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcResponse.kt
```diff
@@ -35,23 +30,29 @@ internal data class JsonRpcError(
...
+            )
+        }
 
-    override fun deserialize(decoder: Decoder): T {
-        require(decoder is JsonDecoder)
+        val jsonRpcResponse =
+            Json.decodeFromString(
+                JsonRpcResponse.serializer(deserializationStrategy),
+                response.body!!,
+            ) // Can we assume that if the request was successful, it will have a body?
```
Solpatium, line 45: I think so https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response/body/
>Returns a non-null value if this response was passed to [Callback.onResponse](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-callback/on-response/) or returned from Call.execute

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
cptartur: I've added this FIXME myself but I now wonder if it's possible to be more descriptive in this case. The if at the beginning of this method handles cases where request failed and there's a body to work with. This one is more like a fallback.

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
```diff
@@ -1,28 +1,24 @@
 package com.swmansion.starknet.service.http
 
 import com.swmansion.starknet.provider.Request
-import kotlinx.serialization.DeserializationStrategy
-import kotlinx.serialization.json.Json
 import java.util.concurrent.CompletableFuture
 
+internal typealias HttpResponseDeserializer<T> = (HttpResponse) -> T
```
Solpatium, line 6: Let's add a comment describing why status code is important here. After gateway is removed we could simplify this.

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
Solpatium: I think we should handle this at higher level, in HttpService itself and response would always have a non-null body. In case body == null we can't handle it specific to service anyway and it leaks into providers.

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -82,7 +129,7 @@ class GatewayProvider(
         )
         val url = feederGatewayRequestUrl("get_storage_at")
 
-        return HttpRequest(Payload(url, "GET", params), Felt.serializer(), httpService)
+        return HttpRequest(Payload(url, "GET", params), buildHttpResponseDeserializer(Felt.serializer()), httpService)
```
Solpatium: I think we could add a method here for creating a request with just the payload and the serialiser. It would wrap serializer with buildHttpResponseDeserializer and inject httpService.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcResponse.kt
```diff
@@ -35,23 +30,29 @@ internal data class JsonRpcError(
...
 
-        if (response.result != null) {
-            return response.result
-        } else if (response.error != null) {
-            throw RpcRequestFailedException(response.error.code, response.error.message)
-        } else {
-            throw RequestFailedException()
+        if (jsonRpcResponse.result == null) {
+            // FIXME add more specific error
+            throw RequestFailedException(response.body)
```
Solpatium, line 53: Maybe we could also add a string explaining what's wrong? LIke "Empty result" in this case?

lib/src/test/kotlin/starknet/provider/handlers/BasicHttpErrorHandlerTest.kt
```diff
@@ -1,32 +1,88 @@
...
         val exception = assertThrows(GatewayRequestFailedException::class.java) {
-            handler.handle(message)
+            request.send()
         }
         assertEquals("error", exception.message)
     }
+
+    @Test
+    fun `rpc handler parses rpc error`() {
+        val message = "{\n" +
```
Solpatium: Please use https://realkotlin.com/tutorials/2018-06-26-multiline-string-literals-in-kotlin/

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcResponse.kt
```diff
@@ -22,27 +18,40 @@ internal data class JsonRpcResponse<T>(
...
+    val message: String,
+)
 
-    override fun deserialize(decoder: Decoder): T {
-        require(decoder is JsonDecoder)
+internal fun <T> buildJsonHttpDeserializer(deserializationStrategy: KSerializer<T>): HttpResponseDeserializer<T> {
+    return { response ->
+        if (!response.isSuccessful) {
+            throw RequestFailedException(
+                response.body,
```
Solpatium, line 37: This is a message in RequestFailedException. Does response.body contain an error message? I think we could just add payload field to the RequestFailedException to avoid confusion.

cptartur: We could include HTTP status message in the response so we get errors like `RequestFailedException("Internal server error"` etc. 

According to the okhttp3 Response docs
```
/**
   * Returns a non-null value if this response was passed to [Callback.onResponse] or returned
   * from [Call.execute]. Response bodies must be [closed][ResponseBody] and may
   * be consumed only once.
   *
   * This always returns null on responses returned from [cacheResponse], [networkResponse],
   * and [priorResponse].
   */
```
body should always be present (we are checking it in the service to be sure), but I guess it can be an empty string sometimes.

Solpatium: I just mean that the body might not be the message for the reason it all failed. I would rather add "payload" field there.

Solpatium (approved):

PR #87, opened 2022-08-25
Default value for blockTag argument in provider methods

bartekryba:
Added  default values to blockTag params in provider methods, so these methods can be called without specifying blockId.

Swapped order of arguments in getClassAt method, to align it with other Provider methods (potential breaking change)


cptartur (commented):
I like this addition. How default parameters behave in java code and documentation?

lib/src/main/kotlin/com/swmansion/starknet/provider/Provider.kt
```diff
@@ -43,14 +43,25 @@ interface Provider {
      */
     fun callContract(call: Call, blockNumber: Int): Request<List<Felt>>
 
+    /**
+     * Calls a contract deployed on StarkNet.
```
cptartur, line 47: Perhaps these overloaded method should have a info in their docs that they use default block tag `LATEST`.

cptartur (approved):

PR #88, opened 2022-08-25
Add account tests

bartekryba:


lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
```diff
@@ -23,7 +23,7 @@ interface Account : Provider {
      * @param params additional execution parameters for the transaction
      * @return signed invoke function
      */
-    fun sign(call: Call, params: ExecutionParams): InvokeTransaction {
+    fun sign(call: Call, params: ExecutionParams): InvokeFunctionPayload {
```
Solpatium, line 26: Result from sign has to be in a shape that allows us to use invoke

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -83,19 +93,19 @@ data class InvokeTransactionReceipt(
     override val actualFee: Felt,
 
     @JsonNames("block_hash")
-    override val blockHash: Felt,
+    override val blockHash: Felt? = null,
```
cptartur, line 96: I'd add FIXME comment to this indicating why this is not following RPC spec currently, and create an issue for updating these once rpc spec get resolved.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -18,18 +18,28 @@ sealed class TransactionReceipt {
     abstract val actualFee: Felt
     abstract val isPending: Boolean
     abstract val type: TransactionReceiptType
+
+    abstract val blockHash: Felt?
+    abstract val blockNumber: Int?
+    abstract val status: TransactionStatus?
```
cptartur, line 24: Why not add these to the `ProcessedTransactionReceipt`? Having these here defeats the purpose of even having separate `TransactionReceipt` and `ProcessedTransactionReceipt`.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -157,6 +167,10 @@ data class PendingTransactionReceipt(
     override val isPending: Boolean = true,
 
     override val type: TransactionReceiptType = TransactionReceiptType.PENDING,
+
+    override val blockHash: Felt? = null,
+    override val blockNumber: Int? = null,
+    override val status: TransactionStatus? = null,
```
cptartur, line 173: I don't like the inclusion of these parameters here when according to the RPC spec they will never exist.

Solpatium: Then they will be nulls. This way you don't need to check instance of to check the value of status.

cptartur: Having `status` here is probably fine but I'm unsure about values that are always null. This is kind of confusing.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
Solpatium: ProcessedTransactionReceipt has TransactionStatus that can't be null. Since all those fields can be null it is easier for users to use them without casting in the first place.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
cptartur: Well, in java handling nulls is not as simple as in kotlin so it'll increase boilerplate in PendingTransactions anyway. If cast to correct type it'd be explicit what fields are available. 

```java
if (blockHash != null) {
    // do smth with block hash
}
```

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
cptartur: Is it possible for the status to be null though? I think if you can query gateway/rpc for the TX, it'll at least have a status.

Solpatium: What if tx has just been sent?

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
Solpatium: Right, with current code they would need to do
```
if (blockHash != null) {
    // do smth with block hash
}
```
with casting in old java they first need to know which class is the right one and then
```
if (receipt instanceof SomeReceipt) {
 SomeReceipt someReceipt = (SomeReceipt) receipt;
 // do smth with receipt
}
```

Also you don't always need to use all fields, sometimes you just need to use some of them and in this case being able to always refer to those fields is very useful. I added it because I had this problem in my tests, previous solution was rather unergonomic.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
cptartur: On gateway it is possible for tx to have `NOT_RECEIVED` status (actually any invalid hash will also return that status), and for full node it'd be impossible to query for transactions that hasn't been somehow processed by StarkNet so an error will most likely be raised.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
cptartur: > with casting in old java they first need to know which class is the right one and then
That's exactly why there's a TransactionReceiptType present in all receipts so it's possible to know the type in java without using instanceof

I still cannot understand the benefit of having these fields declared on __base__ transaction type. We know for a fact from the rpc spec that the pending tx's will never have this fields. This change just creates confusion.

cptartur: Also, receipts will still need to be cast anyway since in rpc invoke, declare and deploy receipts are very different object. So the issue is still present in every non unique field, just not block hash and number.

Solpatium: >That's exactly why there's a TransactionReceiptType present in all receipts so it's possible to know the type in java without using instanceof
You'll need if for this anyway, so how is it different from instanceof?

>I still cannot understand the benefit of having these fields declared on base transaction type. We know for a fact from the rpc spec that the pending tx's will never have this fields. This change just creates confusion.
Let's say you have a process that goes over your txs and checks their statuses. With nullable field you can just use it right away, without the type cast. At the same time if you want you can still type cast and you'll know that this field is not nullable.
I understand your point, but without nice pattern matching it makes things harder.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
Solpatium: So I guess this should be non nullable type at the root of the tx receipts type?

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
cptartur: I can agree that the transaction status being always available may be useful sometimes 

What if we included two custom transaction statuses `PENDING` and `UNKNOWN` and used these as default where necessary instead of optional values. It is much safer to handle these in java than nulls. 

I still don't agree with inclusion of `blockHash` and `blockNumber`. The flow before your changes was better in my opinion

```java
TransactionReceipt receipt = provider.getReceipt(...);
if (receipt.isSuccessful) {
  AcceptedTransactionReceipt acceptedReceipt = (AcceptedTransactionReceipt) receipt;
  // here you have block hash and number available
}
```
Your approach may remove the cast on 3rd line of my snipped but instead you'd have to do `receipt.transactionHash != null` and I fail to understand how that's better than what I wrote.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
cptartur: I think so, we can add some default value just to be sure, I wrote about this in the answer below.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
Solpatium: I don't think adding more statuses changes anything. Nulls are only problematic when you receive them when you expect an object. Having them explicitly is ok in my opinion.

AcceptedTransactionReceipt (Processed) still has nullable block{Number, Hash}, so in your example you still need to handle nulls.

cptartur: I think it depends on the RPC specs. In both newest stable spec and current master branch, blockHash and blockNumber are required fields and cannot be null. If we can get confirmation from StarkWare that this is incorrect and will be changed, I can accept this as a temporary solution.

However if they are non-optional on purpose, Gateway receipt shouldn't extend `ProccesedTransactionReceipt` and only there should `blockHash` and `blockNumber` be optional.

cptartur (approved):

PR #103, opened 2022-09-09
Bump devnet to 0.3.0 and update tests

cptartur:
## Describe your changes

This PR bumps devnet to 0.3.0 (temporarily from github branch) and updates related tests.

This PR also simplifies return type of `Provider.callContract`.

## Linked issues

n/a

## Breaking changes

Return type of `Provider.callContract` changed.

### DIFF
--- a/lib/src/test/kotlin/starknet/utils/DevnetClient.kt
+++ b/lib/src/test/kotlin/starknet/utils/DevnetClient.kt
@@ -13,6 +13,8 @@ import java.util.concurrent.TimeUnit
 import kotlin.io.path.absolutePathString
 import kotlin.io.path.exists
 
+class DevnetSetupFailedException(message: String) : Exception(message)
+
 class DevnetClient(
     private val host: String = "0.0.0.0",
     private val port: Int = 5050,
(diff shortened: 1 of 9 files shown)


lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -1,19 +1,34 @@
...
+    private val accountDirectory = Paths.get("src/test/resources/account")
+    private val baseUrl: String = "http://$host:$port"
 
-    val gatewayUrl: String
-    val feederGatewayUrl: String
-    val rpcUrl: String
+    private lateinit var accountAddress: Felt
+    private lateinit var devnetProcess: Process
+
+    private var isDevnetRunning = false
```
Solpatium, line 27: Why do we need it?

cptartur: Mostly case you tried to do `.start()` multiple times, to make sure we don't spawn multiple devnet processes.

THenry14: Is this possible though? I think invoking `.start()` should fail, since the port is already in use (although let's verify that)

THenry14 (approved):

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
cptartur: It should fail but I think it's a little bit safer this way.

Issue #96, opened 2022-09-07
Separate Account from Provider


cptartur:
Account should not implement Provider methods, instead provider should be injected into Account and used accordingly. Current approach leads to very bloated interface.

PR #111, opened 2022-09-12
Separate Account from Provider

cptartur:
## Describe your changes

This PR separates Account from Provider so that the Account interface becomes less bloated.

## Linked issues

#96 

## Breaking changes


THenry14 (approved):

lib/src/main/kotlin/com/swmansion/starknet/account/StandardAccount.kt
```diff
@@ -20,8 +22,8 @@ class StandardAccount(
     private val provider: Provider,
     override val address: Felt,
     private val signer: Signer,
-) : Account,
-    Provider by provider {
+    private val chainId: StarknetChainId,
```
Solpatium, line 25: Please move it to signer.

Issue #57, opened 2022-08-02
Add JsonRpcProvider status endpoints


bartekryba:
Add status endpoints to `JsonRpcProvider` i.e.:
* `starknet_chainId `
* `starknet_blockNumber`
* `starknet_blockHashAndNumber`
* `starknet_syncing`
* `starknet_getBlockTransactionCount`

PR #112, opened 2022-09-13
Add status endpoints

THenry14:
## Describe your changes

Added gateway/rpc status endpoints

## Linked issues

https://github.com/software-mansion/starknet-jvm/issues/57

## Breaking changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,10 @@ local.properties
 
 # OS specific files
 .DS_Store
+venv
 
 # Exclude gradle wrapper jar
 !gradle/wrapper/gradle-wrapper.jar
+
+# Test artifacts
+lib/src/test/resources
--- a/.gitignore
+++ b/.gitignore
@@ -45,4 +45,4 @@ venv
 !gradle/wrapper/gradle-wrapper.jar
 
 # Test artifacts
-lib/src/test/resources
+lib/src/test/resources/account
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
@@ -59,22 +59,34 @@ data class GetBlockHashAndNumberResponse(
 @Serializable
 sealed class Syncing {
     abstract val status: Boolean
+
     abstract val startingBlockHash: Felt
+
     abstract val startingBlockNumber: Int
+
     abstract val currentBlockHash: Felt
+
     abstract val currentBlockNumber: Int
+
     abstract val highestBlockHash: Felt
+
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -39,6 +39,10 @@ local.properties
 
 # OS specific files
 .DS_Store
+venv
 
 # Exclude gradle wrapper jar
 !gradle/wrapper/gradle-wrapper.jar
+
+# Test artifacts
+lib/src/test/resources
```
cptartur, line 48: I don't think we should be excluding all resources from git. We keep contracts used for tests there etc.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -45,3 +45,53 @@ data class TransactionFailureReason(
...
+    override val startingBlockNumber: Int,
+    @JsonNames("current_block_hash")
+    override val currentBlockHash: Felt,
+    @JsonNames("current_block_num")
+    override val currentBlockNumber: Int,
+    @JsonNames("highest_block_hash")
+    override val highestBlockHash: Felt,
+    @JsonNames("highest_block_num")
+    override val highestBlockNumber: Int,
+) : Syncing()
```
cptartur, line 97: Small nitpick, but can you add newlines between each value as it is done in `Transaction.kt`? I think it improves readability when there are annotations on parameters.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -226,6 +227,54 @@ class JsonRpcProvider(
         return buildRequest(JsonRpcMethod.DECLARE, params, DeclareResponse.serializer())
     }
 
+    override fun getBlockNumber(): Request<Int> {
+        val params = Json.encodeToJsonElement(JsonArray(listOf()))
```
cptartur, line 231:
```suggestion
        val params = Json.encodeToJsonElement(JsonArray(emptyList()))
```

.gitignore
THenry14: Oh wow, I went over the top with this. There is a file that is being created as a result of running tests - `starknet_open_zeppelin_accounts.json` which I wanted to exclude from git. So I guess the question is, should I exclude it here explicitly OR we should maybe get rid of it after tests conclude?

cptartur: I think we should exclude only `lib/src/test/resources/account/` directory. Removing it after tests conclude can be done additionally, but I think that's for another PR.

cptartur (approved):

Issue #108, opened 2022-09-12
Allow deploying any account contracts in tests


cptartur:
Currently test are depended on account contract available through cairo-cli or predeployed in devnet. It should be changed, so any account contract can be easily tested.

PR #113, opened 2022-09-14
Improve deploying contracts for tests

cptartur:
## Describe your changes

This PR adds UDC contract to be used for tests and also improves contract compilation and devnet client.

## Linked issues

#108 

## Breaking changes


lib/src/test/resources/src/account.cairo
```diff
@@ -0,0 +1,379 @@
+// SPDX-License-Identifier: MIT
```
Solpatium, line 1: Can we do `oz_account/{contract.cairo, license}`?

lib/build.gradle.kts
```diff
@@ -57,10 +57,15 @@ val buildCryptoCpp = task<Exec>("BuildCryptoCpp") {
     commandLine("${project.projectDir}/build_crypto_cpp.sh")
 }
 
-// For tests we simply use version from crypto build
+val compileContracts = task<Exec>("compileContracts") {
+    commandLine("${project.projectDir}/src/test/resources/compileContracts.sh")
```
Solpatium, line 61: Update readme to inform that starknet cli is needed now.

Solpatium (changes requested):

lib/src/test/resources/src/account.cairo
cptartur: Maybe `oz_contracts` since it's both account and deployer? This however complicates our compilation scripts.

lib/build.gradle.kts
cptartur: It's only required to run tests, but I'll add that to readme.

lib/src/test/resources/src/account.cairo
Solpatium: however you want, just structure it a little bit. It can also be account_oz.cairo and account_oz_license

Solpatium (approved):

PR #122, opened 2022-09-16
Allow reuse of http client

### LINKED ISSUE #121: Reuse okhttp client
We should share ok http client between different requests: https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#okhttpclients-should-be-shared.

Additionally we should allow applications to inject their own client to share it even further.

Solpatium:
Closes #121

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
@@ -18,16 +18,19 @@ import kotlinx.serialization.builtins.*
 import kotlinx.serialization.json.*
 
 /**
- * A provider for interacting with StarkNet JSON-RPC
+ * A provider for interacting with StarkNet using JSON-RPC. You should reuse it in your application to share the
+ * httpService or provide it with your own httpService.
  *
  * @param url url of the service providing a rpc interface
  * @param chainId an id of the network
+ * @param httpService service used for making http requests
  */
 class JsonRpcProvider(
     private val url: String,
     override val chainId: StarknetChainId,
-    private val httpService: HttpService = OkhttpHttpService(),
+    private val httpService: HttpService,
 ) : Provider {
+    constructor(url: String, chainId: StarknetChainId) : this(url, chainId, OkhttpHttpService())
 
     private fun buildRequestJson(method: String, paramsJson: JsonElement): Map<String, JsonElement> {
         val map = mapOf(
--- a/lib/src/main/kotlin/com/swmansion/starknet/service/http/OkhttpHttpService.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/service/http/OkhttpHttpService.kt
@@ -12,9 +12,13 @@ import java.io.IOException
 import java.util.concurrent.CompletableFuture
 
 /**
- * Service for making http requests.
+ * Service for making http requests using OkHttp library. You can reuse provide it with your client to
+ * <a href="https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#okhttpclients-should-be-shared"> avoid
+ * wasting resources</a>.
+ * @param client OkHttpClient used for making requests
  */
-internal class OkhttpHttpService() : HttpService {
+class OkhttpHttpService(private val client: OkHttpClient) : HttpService {
+    constructor() : this(OkHttpClient())
 
     private val JSON_MEDIA_TYPE = "application/json; charset=utf-8".toMediaType()
 
@@ -51,7 +55,6 @@ internal class OkhttpHttpService() : HttpService {
      * @param payload a payload to be sent
      */
     override fun send(payload: Payload): HttpResponse {
-        val client = OkHttpClient()
         val httpRequest = buildRequest(payload)
 
         val response = client.newCall(httpRequest).execute()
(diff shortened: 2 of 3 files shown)


lib/src/main/kotlin/com/swmansion/starknet/service/http/OkhttpHttpService.kt
```diff
@@ -12,9 +12,13 @@ import java.io.IOException
...
 
 /**
- * Service for making http requests.
+ * Service for making http requests using OkHttp library. You can reuse provide it with your client to
+ * <a href="https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#okhttpclients-should-be-shared"> avoid
+ * wasting resources</a>.
+ * @param client OkHttpClient used for making requests
  */
-internal class OkhttpHttpService() : HttpService {
+class OkhttpHttpService(private val client: OkHttpClient) : HttpService {
```
Solpatium, line 20: Users can now provide their own `OkHttpClient` in here and pass it to providers.

lib/src/main/kotlin/com/swmansion/starknet/service/http/OkHttpService.kt
```diff
@@ -12,9 +12,13 @@ import java.io.IOException
 import java.util.concurrent.CompletableFuture
 
 /**
- * Service for making http requests.
+ * Service for making http requests using OkHttp library. You can reuse provide it with your client to
+ * <a href="https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#okhttpclients-should-be-shared"> avoid
```
cptartur, line 16: I think you can use markdown format for links in KDoc `[](url)`

lib/src/main/kotlin/com/swmansion/starknet/service/http/OkHttpService.kt
```diff
@@ -12,9 +12,13 @@ import java.io.IOException
...
+ * wasting resources</a>.
+ * @param client OkHttpClient used for making requests
  */
-internal class OkhttpHttpService() : HttpService {
+class OkHttpService(private val client: OkHttpClient) : HttpService {
+    constructor() : this(OkHttpClient())
```
cptartur, line 21: Perhaps we shouldn't be providing this secondary constructor at all if we want users to reuse the clients?

README.md
```diff
@@ -101,6 +101,44 @@ public class Main {
...
+import com.swmansion.starknet.service.http.OkHttpService;
+
+// ...
+
+var httpClient = new OkHttpClient();
+var httpService = new OkHttpService(httpClient);
+
+var provider = GatewayProvider.makeTestnetClient(httpService);
+var account1 = new StandardAccount(provider, accountAddress1, privateKey1);
+var account2 = new StandardAccount(provider, accountAddress2, privateKey2);
```
cptartur, line 137: We use explicit types in other places.

cptartur (commented):
I like the reusal of the OkHttpClient, however I'm not sure if exposing OkHttpService entirely is a good approach. 

Maybe we could abstract the creation of `OkHttpService`, for example by adding a factory method to the `HttpService` interface?

lib/src/main/kotlin/com/swmansion/starknet/service/http/OkHttpService.kt
Solpatium: This way they can reuse http service.

README.md
Solpatium: (the code above can be copied to an actual codebase, code here doesn't make sense apart from illustrative reasons)

lib/src/main/kotlin/com/swmansion/starknet/service/http/OkHttpService.kt
cptartur: I mean the constructor where `OkHttpClient` is automatically created, not the primary one.

Solpatium: Default construtor -> reuse HttpService
Constructor with client -> reuse OkHttpClient

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -12,22 +12,25 @@ import com.swmansion.starknet.provider.Request
...
-import com.swmansion.starknet.service.http.OkhttpHttpService
+import com.swmansion.starknet.service.http.OkHttpService
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.builtins.*
 import kotlinx.serialization.json.*
 
 /**
- * A provider for interacting with StarkNet JSON-RPC
+ * A provider for interacting with StarkNet using JSON-RPC. You should reuse it in your application to share the
+ * httpService or provide it with your own httpService.
```
cptartur, line 22:
```suggestion
 * HttpService or provide it with your own HttpService.
```

lib/src/main/kotlin/com/swmansion/starknet/service/http/OkHttpService.kt
```diff
@@ -12,9 +12,12 @@ import java.io.IOException
 import java.util.concurrent.CompletableFuture
 
 /**
- * Service for making http requests.
+ * Service for making http requests using OkHttp library. You can reuse provide it with your client to
```
cptartur, line 15:
```suggestion
 * Service for making http requests using OkHttp library. You can provide it with your client to
```

README.md
```diff
@@ -101,6 +101,27 @@ public class Main {
 }
 ```
 
+## Reusing http clients
+
+Make sure you don't create a new provider every time you want to use one. Instead, you should reuse existing instance.
+This way you reuse connections and thread pools. You can read more about it [here](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#okhttpclients-should-be-shared).
```
cptartur, line 107: I'd move the lines about reusing connection pools etc. to the OkHttpService docs.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
Solpatium: You share the property, not interface 🤔

cptartur (approved):

PR #136, opened 2022-09-23
Add Deployer contract

### LINKED ISSUE #130: Support deploy via Universal Deploy Contracts

### LINKED ISSUE #134: Add `isAccepted` method to transaction receipt
Currently, to check if transaction was accepted, one has to compare receipt.status with both `ACCEPTED_ON_L1` and `ACCEPTED_ON_L2`. We should add a helper method that performs this comparison an returns a boolean.

cptartur:
## Describe your changes

This PR adds initial support for universal deployer contract and some minor improvements

## Linked issues

Closes #130 
Closes #134

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
@@ -38,6 +38,10 @@ sealed class TransactionReceipt {
     abstract val isPending: Boolean
     abstract val type: TransactionReceiptType
     abstract val status: TransactionStatus
+
+    fun isAccepted(): Boolean {
+        return (status == TransactionStatus.ACCEPTED_ON_L1) || (status == TransactionStatus.ACCEPTED_ON_L2)
+    }
 }
 
 @Serializable
--- a/lib/src/main/kotlin/com/swmansion/starknet/deployer/Deployer.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/deployer/Deployer.kt
@@ -1,0 +1,40 @@
+package com.swmansion.starknet.deployer
+
+import com.swmansion.starknet.data.types.Calldata
+import com.swmansion.starknet.data.types.Felt
+import com.swmansion.starknet.provider.Request
+import com.swmansion.starknet.provider.exceptions.RequestFailedException
+
+data class ContractDeployment(
+    val transactionAddress: Felt,
+)
+
+class AddressRetrievalFailedException(message: String) : Exception(message)
+
+/**
+ * Universal Deployer Contract module
+ *
+ * A module for interacting with Universal Deployer Contracts (UDC).
+ */
+interface Deployer {
+    /**
+     * Deploy a contract through Universal Deployer Contract (UDC)
+     *
+     * @param classHash a class hash of the declared contract
+     * @param salt a salt to be used to calculate deployed contract address
+     * @param calldata constructor calldata
+     *
+     * @throws RequestFailedException
+     */
+    fun deployContract(classHash: Felt, salt: Felt, calldata: Calldata): Request<ContractDeployment>
+
+    /**
+     * Get a contract address from the deployment
+     *
+     * @param contractDeployment a contract deployment
+     *
+     * @throws RequestFailedException
+     * @throws AddressRetrievalFailedException
+     */
+    fun contractAddress(contractDeployment: ContractDeployment): Request<Felt>
+}
--- a/lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
@@ -1,0 +1,44 @@
+package com.swmansion.starknet.deployer
+
+import com.swmansion.starknet.account.Account
+import com.swmansion.starknet.data.selectorFromName
+import com.swmansion.starknet.data.types.Call
+import com.swmansion.starknet.data.types.Calldata
+import com.swmansion.starknet.data.types.Event
+import com.swmansion.starknet.data.types.Felt
+import com.swmansion.starknet.data.types.transactions.GatewayTransactionReceipt
+import com.swmansion.starknet.data.types.transactions.InvokeTransactionReceipt
+import com.swmansion.starknet.data.types.transactions.TransactionReceipt
+import com.swmansion.starknet.extensions.map
+import com.swmansion.starknet.provider.Provider
+import com.swmansion.starknet.provider.Request
+
+class StandardDeployer(
+    private val deployerAddress: Felt,
+    private val provider: Provider,
+    private val account: Account,
+) : Deployer {
+    override fun deployContract(classHash: Felt, salt: Felt, calldata: Calldata): Request<ContractDeployment> {
+        val invokeCalldata = listOf(classHash, salt, Felt.ONE, Felt(calldata.size)) + calldata
+        val call = Call(deployerAddress, "deployContract", invokeCalldata)
+
+        return account.execute(call).map { ContractDeployment(it.transactionHash) }
+    }
+
+    override fun contractAddress(contractDeployment: ContractDeployment): Request<Felt> {
+        return provider.getTransactionReceipt(contractDeployment.transactionAddress).map { receipt ->
+            val event = getDeploymentEvent(receipt)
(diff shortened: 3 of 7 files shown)


lib/src/main/kotlin/com/swmansion/starknet/deployer/Deployer.kt
```diff
@@ -0,0 +1,40 @@
+package com.swmansion.starknet.deployer
+
+import com.swmansion.starknet.data.types.Calldata
+import com.swmansion.starknet.data.types.Felt
+import com.swmansion.starknet.provider.Request
+import com.swmansion.starknet.provider.exceptions.RequestFailedException
+
+data class ContractDeployment(
+    val transactionAddress: Felt,
```
Solpatium, line 9: `transactionAddress`? You mean `transactionHash`?

lib/src/main/kotlin/com/swmansion/starknet/deployer/Deployer.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Get a contract address from the deployment
+     *
+     * @param contractDeployment a contract deployment
+     *
+     * @throws RequestFailedException
+     * @throws AddressRetrievalFailedException
+     */
+    fun contractAddress(contractDeployment: ContractDeployment): Request<Felt>
```
Solpatium, line 39: `findContractAddress`?

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
```diff
@@ -0,0 +1,44 @@
+package com.swmansion.starknet.deployer
```
Solpatium, line 1: I would call this package universalDeployerContact or something like this. Who knows what could be used as a deployer in the future.

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.swmansion.starknet.extensions.map
+import com.swmansion.starknet.provider.Provider
+import com.swmansion.starknet.provider.Request
+
+class StandardDeployer(
+    private val deployerAddress: Felt,
+    private val provider: Provider,
+    private val account: Account,
+) : Deployer {
+    override fun deployContract(classHash: Felt, salt: Felt, calldata: Calldata): Request<ContractDeployment> {
```
Solpatium, line 21: We don't expose unique here. Why?

Solpatium, line 21: let's name calldata `constructorCalldata`

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
```diff
@@ -0,0 +1,44 @@
...
+        }
+    }
+
+    private fun getDeploymentEvent(transactionReceipt: TransactionReceipt): Event? {
+        val events = when (transactionReceipt) {
+            is InvokeTransactionReceipt -> transactionReceipt.events
+            is GatewayTransactionReceipt -> transactionReceipt.events
+            else -> throw AddressRetrievalFailedException("Invalid transaction type")
+        }
+        return events.find { it.keys.contains(selectorFromName("ContractDeployed")) }
```
Solpatium, line 42: What if there were more than 2 events like this? It is possible.

Solpatium (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
cptartur: You mean this implementation or the whole package with `Deployer` interface?

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
cptartur: Seems kinda redundant but okay.

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
cptartur: [Find](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/find.html) would return the first match (how events are ordered is unknown to me).

This is only designed to handle deployments created by this specific `StandardDeployer`, and I'm believe it's only possible for them to emit one deployment event per transaction: Typing of the interface could perhaps be improved so its impossible to pass deployment from other interface implementations to this method.

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
cptartur: It didn't seem useful exposing this parameter.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -38,6 +38,8 @@ sealed class TransactionReceipt {
     abstract val isPending: Boolean
     abstract val type: TransactionReceiptType
     abstract val status: TransactionStatus
+    val isAccepted: Boolean
```
Solpatium, line 41: How is this translated to java?

lib/src/main/kotlin/com/swmansion/starknet/deployer/Deployer.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.swmansion.starknet.data.types.Calldata
+import com.swmansion.starknet.data.types.Felt
+import com.swmansion.starknet.provider.Request
+import com.swmansion.starknet.provider.exceptions.RequestFailedException
+
+data class ContractDeployment(
+    val transactionHash: Felt,
+)
+
+class AddressRetrievalFailedException(message: String) : Exception(message)
```
Solpatium, line 12: I think we can also include `ContractDeployment`. This way users will see in logs what exectly went wrong.

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
Solpatium: package itself

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
Solpatium: That's why I think we should fail if more of them are available - we don't know which one will it be.

Solpatium (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
cptartur: I guess we could do that.

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
Solpatium: Why is it not usefeul? It creates a virtual namespace based on your account by default.

lib/src/main/kotlin/com/swmansion/starknet/deployer/StandardDeployer.kt
cptartur: Renamed `universaldeployercontract` because it seems like it's the most popular java naming convention for multi word name.

Solpatium (approved):

lib/src/main/kotlin/com/swmansion/starknet/deployercontract/StandardDeployer.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.swmansion.starknet.extensions.map
+import com.swmansion.starknet.provider.Provider
+import com.swmansion.starknet.provider.Request
+
+class StandardDeployer(
+    private val deployerAddress: Felt,
+    private val provider: Provider,
+    private val account: Account,
+) : Deployer {
+    override fun deployContract(
```
Solpatium, line 21: I would also offer a deployContract with some default parameters: unique = True and salt generated with bouncy castle.

Solpatium (approved):

Issue #68, opened 2022-08-10
Manual promotion to maven cetral via gh


THenry14:
Right now to promote the staged package to maven central, we have to manually close repo via nexus gui as per https://central.sonatype.org/publish/publish-gradle/#distribution-management-and-authentication

There are some plugins to gradle, however, that could be used to move this flow to github actions (like https://github.com/gradle-nexus/publish-plugin/).

The objective of this task is to assess possibilties to automate the promotion of the package, and create a new gh actions flow that would handle this upon manual user interaction (https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow).

THenry14:
+ also docs should only be deployed (built?) after a successfull promotion of a new release

PR #140, opened 2022-09-27
New release mechanism

THenry14:
## Describe your changes

This PR changes the flow of package release to maven. 
In order to release bump a version in gradle build file and create a new tag matching this version. This will trigger build an publish flow.

## Linked issues

Closes https://github.com/software-mansion/starknet-jvm/issues/68


gradle.properties
```diff
@@ -1,2 +1,3 @@
 android.useAndroidX=true
-org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m
\ No newline at end of file
+org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m
+gnsp.disableApplyOnlyOnRootProjectEnforcement=true
```
cptartur, line 3: Why was this added?

lib/build.gradle.kts
```diff
@@ -141,70 +142,69 @@ publishing {
...
+                    val okHttpDependency = dependencyNode.appendNode("dependency")
+                    okHttpDependency.appendNode("groupId", "com.squareup.okhttp3")
+                    okHttpDependency.appendNode("artifactId", "okhttp")
+                    okHttpDependency.appendNode("version", "4.10.0")
+
+                    val kotlinxDependency = dependencyNode.appendNode("dependency")
+                    kotlinxDependency.appendNode("groupId", "org.jetbrains.kotlinx")
+                    kotlinxDependency.appendNode("artifactId", "kotlinx-serialization-json")
+                    kotlinxDependency.appendNode("version", "1.3.3")
                 }
```
cptartur, line 204: How complicated it'd be to include these automatically based on main build.gradle.kts? That way we won't have to remember to update these every time we bump a dependency version.

THenry14: The problem is, we have mutliple deps decalred in the main build file, and not all of them should be noted as dependencies for the package (eg. test stuff). So, not sure how complex, but I imagine quite, as we still would have to somehow specify what goes in here one way or the other

gradle.properties
THenry14: >* What went wrong:
An exception occurred applying plugin request [id: 'io.codearte.nexus-staging']
> Failed to apply plugin 'io.codearte.nexus-staging'.
   > Nexus staging plugin should ONLY be applied on the ROOT project in a build. See https://github.com/Codearte/gradle-nexus-staging-plugin/issues/47 for explanation. If you really know what you are doing it can be overridden with setting 'gnsp.disableApplyOnlyOnRootProjectEnforcement' property.
   
The problem is, we have multi-project setup, and I cannot take the whole lib' `publishing` block to the root

lib/build.gradle.kts
cptartur: I see we already used hardcoded dependencies in the previous version. Could you open an issue for making these automatically included? For the time being, we can merge it as is.

THenry14: https://github.com/software-mansion/starknet-jvm/issues/143

cptartur (approved):
Perhaps we should add a information to README.md or in other place that creating a tag is needed for a new version release?

PR #149, opened 2022-10-05
Support signed declare transaction

### LINKED ISSUE #145: Support signing and preparing declare transactions through Account
Currently, `Declare` transactions have to be manually prepared and signed. This flow should be improved to allow easy Declare preparation and signing through `Account` interface.

cptartur:
## Describe your changes

This PR adds support for signing declare transactions and sending them.

## Linked issues

Closes #145 

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
@@ -300,4 +300,35 @@ object TransactionFactory {
             signature = signature,
         )
     }
+
+    @JvmStatic
+    fun makeDeclareTransaction(
+        classHash: Felt,
+        senderAddress: Felt,
+        contractDefinition: ContractDefinition,
+        chainId: StarknetChainId,
+        maxFee: Felt,
+        version: Felt,
+        nonce: Felt,
+        signature: Signature = emptyList(),
+    ): DeclareTransaction {
+        val hash = TransactionHashCalculator.calculateDeclareTxHash(
+            classHash = classHash,
+            chainId = chainId,
+            senderAddress = senderAddress,
+            maxFee = maxFee,
+            version = version,
+            nonce = nonce,
+        )
+        return DeclareTransaction(
+            classHash = classHash,
+            senderAddress = senderAddress,
+            contractDefinition = contractDefinition,
+            hash = hash,
+            maxFee = maxFee,
+            version = version,
+            signature = signature,
+            nonce = nonce,
+        )
+    }
 }
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
@@ -127,7 +127,6 @@ data class InvokeTransaction(
 @SerialName("DECLARE")
 data class DeclareTransaction(
     @SerialName("class_hash")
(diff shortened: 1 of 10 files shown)


cptartur:
Thing to consider: with current supported version of the RPC spec, it's impossible to send signed declare transactions. Starknet-jvm will first have to be updated to support RPC 0.2.0 for these to be able to work.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -144,7 +143,24 @@ data class DeclareTransaction(
     override val nonce: Felt,
 
     override val type: TransactionType = TransactionType.DECLARE,
-) : Transaction()
+
+    private val contractDefinition: ContractDefinition? = null,
+) : Transaction() {
+    @Throws(ConvertingToPayloadFailedException::class)
```
Solpatium, line 149: I think we have toPayload in other places as well. I think we can resuse it there as well?

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionPayload.kt
```diff
@@ -31,8 +32,17 @@ data class DeployTransactionPayload(
...
 data class DeclareTransactionPayload(
     val contractDefinition: ContractDefinition,
+    val senderAddress: Felt,
     val maxFee: Felt,
     val nonce: Felt,
     val signature: Signature,
     val version: Felt,
-)
+) {
+    constructor(
```
Solpatium, line 41: Why is this needed?

Solpatium (approved):

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionPayload.kt
cptartur: I guess it's only used in tests using old declare transactions (unsigned). I'll remove this as we don't really want to support tx version 0.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
cptartur: I'll try to think of a way of making `toPayload` more defined and reusable.

cptartur: I don't think this comment still applies?

cptartur:
Note that the Rpc provider doesn't support sending signed Declare txs. I've added an error for that so it doesn't fail silently/is ignored. This will have to be addressed once rpc 0.2.1 is supported, shouldn't require any changes to the Account itself however.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -145,6 +143,34 @@ class StandardAccountTest {
...
+            classHash,
+            ExecutionParams(nonce, Felt(1000000000000000)),
+        )
+        val request = provider.declareContract(declareTransactionPayload)
+        val result = request.send()
+        val receipt = provider.getTransactionReceipt(result.transactionHash).send()
+
+        assertNotNull(result)
+        assertNotNull(receipt)
+        println(receipt)
```
THenry14, line 170: 👀

THenry14 (approved):
lgtm!

Solpatium (approved):

Issue #154, opened 2022-10-10
`GatewayProvider.callContract` fails when using `blockTag` or `blockNumber`.


cptartur:
Constructed request is probably incorrect - blockHash is always used.

florian-bellotti:
https://github.com/software-mansion/starknet-jvm/blob/main/lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt#L61

When the blockId is a number or a tag, it should set the `blockNumber` query parameter.
And, it should set `blockHash` when blockId is a hash.

PR #158, opened 2022-10-11
:bug: Fixing callContract when block id is a number or a tag

florian-bellotti:
## Describe your changes

GatewayProvider.callContract fails when using blockTag or blockNumber. It always sends `blockHash` even if the block id is a number or a tag.

## Linked issues

Closes --> https://github.com/software-mansion/starknet-jvm/issues/154


Solpatium:
Thank you for creating a PR for us, we always appreciate users who want to contribute ❤️
Unfortunately @cptartur was working on fix in the same time as you. We decided to merge his PR (https://github.com/software-mansion/starknet-jvm/pull/159/files) as it fixes this problem in other places as well in `GatewayProvider`. 
To avoid doing unnecessary work please make sure no one is assigned to the issue that you want to fix. If there is someone there is a chance someone is already working on it.

PR #159, opened 2022-10-11
Fix gateway block id

### LINKED ISSUE #154: `GatewayProvider.callContract` fails when using `blockTag` or `blockNumber`.

cptartur:
## Describe your changes

This PR fixes incorrect serialization of block id in gateway methods.

## Linked issues

Closes #154 

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/extensions/ToParam.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/extensions/ToParam.kt
@@ -1,0 +1,10 @@
+package com.swmansion.starknet.extensions
+
+import com.swmansion.starknet.data.types.BlockId
+
+@JvmSynthetic
+internal fun BlockId.toParam(): Pair<String, String> = when (this) {
+    is BlockId.Hash -> Pair("blockHash", this.blockHash.hexString())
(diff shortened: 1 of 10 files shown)


lib/src/main/kotlin/com/swmansion/starknet/extensions/ToParam.kt
```diff
@@ -0,0 +1,10 @@
+package com.swmansion.starknet.extensions
+
+import com.swmansion.starknet.data.types.BlockId
+
+@JvmSynthetic
+internal fun BlockId.toParam(): Pair<String, String> = when (this) {
```
Solpatium, line 6: This is only used in GatewayProvider, why not define it there as a helper function?

Solpatium (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/extensions/ToParam.kt
cptartur: It would complicate testing of this function. While generally we probably shouldn't need to tests internals like this, issue this PR is resolving is caused by us not covering these cases (because test e2e test are incomplete and because of devnet limitations).

Solpatium (approved):

Issue #164, opened 2022-10-12
Consider exposing more details as public


Solpatium:
> I did a quick check through the SDK and I think those functions / classes could be public instead of internal:
> - HttpRequest` & HttpResponseDeserializer
> - Request, Felt`, Json extensions
> - Transaction#toPayload
> - GatewayProvider` - all constructor params could be public, it's useful if u want to make the request on your own

Solpatium:
We've discussed this and decided to make public:
- HttpRequest and HttpResponseDeserializer. There is a problem with `HttpResponseDeserializer` - it is translated as kotlin's function to java. It has to be changed to either be `Function` from java or use https://kotlinlang.org/docs/fun-interfaces.html#sam-conversions.
- `BigInteger.toFelt`, but it has to be moved to extensions.BigInteger.
- `Transaction.toPayload`
- `GatewayProvider` - all constructor params can be public

We've discussed Requests and decided it would be the best if applications wrapped our Accounts within their custom class used for account interactions, selecting synchronous/asynchronous ways. Every operation that composes from other operations should be exposed with variant that accepts results of those other operations.
We might one day to decide to implement Requests differently (like web3j for instance) and we don't want users to depend on `compose` and `map`.

cptartur:
> HttpRequest and HttpResponseDeserializer. There is a problem with HttpResponseDeserializer - it is translated as kotlin's function to java. It has to be changed to either be Function from java or use https://kotlinlang.org/docs/fun-interfaces.html#sam-conversions.

I'd prefer we used `java.util.Function` there. SAM conversions can still be used with it, so it won't complicate our implementations (basically just use `Function { ... }` instead of `{ ... }` when defining lambdas.

PR #169, opened 2022-10-20
Change some things to public

THenry14:
## Describe your changes

Change:
- HttpRequest and HttpResponseDeserializer
- BigInteger.toFelt
- Transaction.toPayload
- GatewayProvider constructor parameters

from internal/private to public.

## Linked issues

Closes https://github.com/software-mansion/starknet-jvm/issues/164

## Breaking changes


lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -99,7 +99,7 @@ data class InvokeTransaction(
 
     override val type: TransactionType = TransactionType.INVOKE,
 ) : Transaction() {
-    internal fun toPayload(): InvokeFunctionPayload {
+    fun toPayload(): InvokeFunctionPayload {
         val invocation = Call(
             contractAddress = contractAddress,
             calldata = calldata,
```
cptartur, line 105: Perhaps we should add a comment indicating that this function is not exactly a part of the stable API? 

cc @Solpatium

lib/src/main/kotlin/com/swmansion/starknet/extensions/BigInteger.kt
```diff
@@ -0,0 +1,8 @@
+package com.swmansion.starknet.extensions
+
+import com.swmansion.starknet.data.types.Felt
+import java.math.BigInteger
+
+@get:JvmSynthetic
+val BigInteger.toFelt: Felt
+    get() = Felt(this)
```
cptartur, line 8: This being both public and jvmsynthetic will be a bit confusing for java users since they'll see this method in docs but will be unable to use it.

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -28,10 +29,10 @@ import kotlinx.serialization.json.*
  * @param httpService service used for making http requests
  */
 class GatewayProvider(
-    private val feederGatewayUrl: String,
-    private val gatewayUrl: String,
+    val feederGatewayUrl: String,
+    val gatewayUrl: String,
     override val chainId: StarknetChainId,
-    private val httpService: HttpService,
+    val httpService: HttpService,
```
cptartur, line 35: Are we sure we want to expose service here?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -28,9 +28,9 @@ import kotlinx.serialization.json.*
  * @param httpService service used for making http requests
  */
 class JsonRpcProvider(
-    private val url: String,
+    val url: String,
     override val chainId: StarknetChainId,
-    private val httpService: HttpService,
+    val httpService: HttpService,
```
cptartur, line 33: Same as above

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
Solpatium: I think service can remain private

lib/src/main/kotlin/com/swmansion/starknet/extensions/BigInteger.kt
Solpatium: What do you suggest? For java users this extension doesn't have a nice API.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
Solpatium: What does it change for the users? If we need to remove it we'll deprecate it first and offer an alternative anyway.

lib/src/main/kotlin/com/swmansion/starknet/extensions/BigInteger.kt
cptartur: I don't think there's a nice way to solve this: we either have to accept that we document and expose features that are unusable (prioritize kotlin users), or perhaps change make it a standard function instead of extension (prioritize java users).

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
cptartur: Maybe let's not expose it at all then and consider alternative approach now - because it doesn't make sense to me to expose it just to have it removed 2 versions from now (and deprecations aren't really practical at this stage of the development).

lib/src/main/kotlin/com/swmansion/starknet/extensions/BigInteger.kt
Solpatium: Is there really no way to ignore this in documentation for java users? For kotlin users having this as an extension would be very useful.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
THenry14: If I remember correctly, this is one of the functions mentioned specifically by someone from the community. I think I'd just expose it anyway now, and deprecate/remove later on. Is that ok with you guys?

Solpatium: It is ok for me, the functionality here is straightforward. In my opinion the problem is when deprecating something that touches a lot other code.

lib/src/main/kotlin/com/swmansion/starknet/extensions/BigInteger.kt
THenry14: Removed it from javadoc in gradle

Solpatium (approved):

cptartur (approved):

Issue #131, opened 2022-09-22
Support deploy account transaction


Solpatium:
It is not ready yet.

Issue #174, opened 2022-11-03
getEstimateFee should accept InvokeFunctionPayload


Solpatium:
Currently `getEstimateFee` currently accepts `InvokeTransaction`, but users don't get it from our API. We should use payload, just like we use it for sending transactions.

PR #176, opened 2022-11-03
Support deploy account tx

Solpatium:
## Describe your changes

- Added deploy_account support to GatewayProvider and Account (signing txs). 
- Unfortunately cairo 10.1 changed a lot (required for deploy_account) in CLI and had to change DevnetClient.
- Bumped devnet to version supporting cairo 10.1
- Moved hash calculations to a separate utility class

## Linked issues

Closes https://github.com/software-mansion/starknet-jvm/issues/131

## Breaking changes

Adds one more transaction type


lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -272,6 +286,8 @@ class GatewayProvider(
         return getClassHashAt(param, contractAddress)
     }
 
+    // TODO: Accept payload instead of tx
+    // TODO: Move serialization to a method
```
Solpatium, line 290: https://github.com/software-mansion/starknet-jvm/issues/174
https://github.com/software-mansion/starknet-jvm/issues/178

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -58,11 +61,12 @@ class StandardAccountTest {
         private fun deployAccount() {
             val contractDeployer = ContractDeployer.deployInstance(devnetClient)
             val (classHash, _) = devnetClient.declareContract(Path.of("src/test/resources/compiled/account.json"))
+            accountClassHash = classHash
             accountAddress = contractDeployer.deployContract(classHash, calldata = listOf(signer.publicKey))
             devnetClient.prefundAccount(accountAddress)
         }
 
-        data class AccountAndProvider(val account: Account, val provider: Provider)
+        data class AccountAndProvider(val account: Account, val provider: GatewayProvider)
```
Solpatium, line 69: Only GatewayProvider offers deploy_account at the moment.

lib/src/main/kotlin/com/swmansion/starknet/data/TransactionHashCalculator.kt
```diff
@@ -0,0 +1,97 @@
...
+import com.swmansion.starknet.crypto.StarknetCurve
+import com.swmansion.starknet.data.types.Calldata
+import com.swmansion.starknet.data.types.Felt
+import com.swmansion.starknet.data.types.StarknetChainId
+import com.swmansion.starknet.data.types.transactions.TransactionType
+
+/**
+ * Toolkit for calculating hashes of transactions.
+ */
+object TransactionHashCalculator {
```
Solpatium, line 12: I hate the name of this and ContractAddressCalculator and I would gladly accept suggestions.

cptartur (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -232,6 +235,17 @@ class GatewayProvider(
...
+    fun deployAccount(payload: DeployAccountTransactionPayload): Request<DeployAccountResponse> {
+        val url = gatewayRequestUrl("add_transaction")
+        val body = serializeDeployAccountTransactionPayload(payload)
+
+        return HttpRequest(
+            Payload(url, "POST", body),
+            buildDeserializer(DeployAccountResponse.serializer()),
+            httpService,
+        )
+    }
```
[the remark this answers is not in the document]
Solpatium: I disagree, this leads a situation where you expect everything to work with any Provider and then you get surprised. 
I want to add it to the interface once it is supported by RpcProvider.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -271,4 +275,62 @@ class StandardAccountTest {
...
+        )
+
+        val response = provider.deployAccount(payload).send()
+
+        // Make sure the address matches the calculated one
+        assertEquals(address, response.address)
+
+        // Make sure tx matches what we sent
+        val tx = provider.getTransaction(response.transactionHash).send() as DeployAccountTransaction
+        assertEquals(payload, tx.toPayload())
```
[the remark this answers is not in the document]
Solpatium: toPayload is a very simple method. Comparing it field by field is cumbersome.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -158,17 +207,46 @@ object TransactionFactory {
...
+    fun makeDeployAccountTransaction(
+        classHash: Felt,
+        salt: Felt,
+        calldata: Calldata,
+        chainId: StarknetChainId,
+        version: Felt,
+        maxFee: Felt = Felt.ZERO,
+        signature: Signature = emptyList(),
+    ): DeployAccountTransaction {
+        val nonce = Felt.ZERO
```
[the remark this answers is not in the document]
Solpatium: DeployAccount's nonce will always be 0.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -65,20 +66,30 @@ class ProviderTest {
         fun before() {
             try {
                 devnetClient.start()
-                val (deployAddress, deployHash) = devnetClient.deployContract(Path.of("src/test/resources/compiled/providerTest.json"))
+
+                val contract = Path.of("src/test/resources/compiled/providerTest.json").toFile().readText()
+                val deployTx = gatewayProvider().deployContract(
```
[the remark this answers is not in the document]
Solpatium: I like it this way, otherwise I need to create variables that are similar to contractAddress and deployTransactionHash from object's scope.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
Solpatium: I just used gateway right away, it is the simplest way

lib/src/main/kotlin/com/swmansion/starknet/data/TransactionHashCalculator.kt
Solpatium: @THenry14 any ideas?

THenry14: Nah, that's why I liked Artur's comment

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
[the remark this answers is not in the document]
cptartur: Having it in the interface and raising an exception indicates that it is intended to be implemented by all Providers (including custom ones, there's that usecase!). 

We did just fine with raising non implemented exceptions in the rpc in starknet.py so I don't think it's a big problem to handle for users.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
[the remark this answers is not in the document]
cptartur: This way the test is dependent on the `toPayload` implementation. This should be changed.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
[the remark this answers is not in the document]
cptartur: It's inconsistent and looks out of place. If you prefer this approach, lets change all other uses in this method as well.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -305,13 +307,38 @@ class StandardAccountTest {
...
+            privateKey,
+            provider,
+        )
         val payload = account.signDeployAccount(
             classHash = classHash,
             salt = salt,
             calldata = calldata,
-            maxFee = maxFee,
+            // 10*fee from estimate deploy account fee
+            maxFee = Felt.fromHex("0x11fcc58c7f7000"),
```
cptartur, line 341: I think it'd be more readable not to use hex here.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -305,13 +307,38 @@ class StandardAccountTest {
             calldata = calldata,
             maxFee = Felt.ZERO,
         )
-        val fee = provider.getEstimateFee(payloadForFeeEstimation).send()
-        val maxFee = fee.overallFee
+        val feePayload = provider.getEstimateFee(payloadForFeeEstimation).send()
+        assertTrue(feePayload.overallFee.value > Felt.ONE.value)
+        println(feePayload)
```
cptartur, line 312:
```suggestion
```

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -271,4 +280,84 @@ class StandardAccountTest {
         val receipt2 = provider.getTransactionReceipt(result2.transactionHash).send()
         assertEquals(TransactionStatus.ACCEPTED_ON_L2, receipt2.status)
     }
+
+    @Test
+    fun `estimate deploy account fee`() {
+        val provider = gatewayProvider
```
cptartur, line 286: Why not use `if (provider !is GatewayProvider) return` like I suggested? Once we want to extend this test test suite (which is likely with this test suite), we'll have to write more code here instead of just removing one line.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
Solpatium: I can't use integer, because it its bigger than the maximum value. What is not readable?

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
Solpatium: Because it overcomplicates a simple case. Why should I use parametrised test if I can run only one case? If we ever need to change that we will change that. Changes would be required in both cases, so I prefer to keep it simpler.

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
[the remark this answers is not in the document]
Solpatium: >We did just fine with raising non implemented exceptions in the rpc in starknet.py so I don't think it's a big problem to handle for users.
I think java users depend more heavily on interfaces than python users, but you are right, we shouldn't have handled that with exceptions.

Having an interface that is broken by one of two providers exposed by us breaks type safety and is surprising for users. SDK should not be surprising. 
If someone wants to create a custom Provider then they may as well create an interface extending ours, I won't trade type safety for handling a use case like this.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
cptartur: I believe you can create Felt from BigInteger and BigInteger allows construction from decimal string (more readable than hex imo).

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
cptartur: All other tests in this module are parametrized so it's inconsistent.

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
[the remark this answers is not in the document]
cptartur: Perhaps we should focus more on introducing required changes in both providers at the same time in the future. 
Situation where we only add features to GatewayProvider when RpcProvider is supposed to be a default one is very much not ideal.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
Solpatium: But it is not important what the value is, it is just a big number. Why do you need this?

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
Solpatium: Tests should be easy to read, doing it this way will make it more complex. Just because other tests use require some parameters doesn't mean that every test needs that.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
cptartur: It's just a little bit easier to read what value is being passed to transaction if it's in decimal.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
cptartur: You're trading ease to to maintain the test suite for the ease to read (arguably) and I'm not sure if that's a good thing.

Solpatium: I disagree that it makes maintenance of the tests harder, I think it makes is easier. Parametrised tests make sense when they actually parametrise something, otherwise they just obfuscate the real logic that is used. 

@THenry14 please provide your feedback here, I'll do what you decide.

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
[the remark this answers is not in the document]
Solpatium: I agree, but supporting RPC 2.1 was quite problematic. Hopefully moving forward we will be able to add changes to both of them at the same time.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
Solpatium: I prefer using hex values, because they are shorter and the actual value doesn't matter.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
THenry14: I guess it doesn't really matter which way we go, BUT imho tests should be as simple as possible and test only what they are supposed to test + returning nothing actually tests nothing. Moreover, it can be parametrized when actually needed. So, sorry Artur but 
obraz

cptartur: 😭

Solpatium: Also we agreed with Artur to add FIXME comment before the test.

cptartur (approved):
Let's create an issue for supporting deploy account in the rpc provider since this PR will close #131

Solpatium:
Deploy account is part of RPC 2.1, so it will be a part of https://github.com/software-mansion/starknet-jvm/issues/146

PR #177, opened 2022-11-03
Add `L1_HANDLER` transaction support for GatewayProvider

### LINKED ISSUE #167: Support L1_HANDLER tx
It is a valid tx type that we don't support currently. Similar issue was in starknet.py https://github.com/software-mansion/starknet.py/pull/452

cptartur:
## Describe your changes

This PR adds L1HandlerTransaction dataclass as well as suport for parsing these transactions to GatewayProvider.

It also inlines `transactionTypeToName` function.

Tests have been mocked until we resolve how to test L1<>L2 messaging with devnet (this is also being investigated by starknet.py team).

This PR however doesn't add support for L1HandlerTransactions to JsonRpcProvider as it is blocked by #146 

## Linked issues

Closes #167 

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
@@ -21,7 +21,7 @@ enum class TransactionType(val txPrefix: Felt) {
     INVOKE(Felt.fromHex("0x696e766f6b65")), // encodeShortString('invoke'),
 
     @SerialName("L1_HANDLER")
-    L1_HANDLER(Felt.fromHex("0x6c315f68616e646c6572")) // encodeShortString('1_handler')
+    L1_HANDLER(Felt.fromHex("0x6c315f68616e646c6572")) // encodeShortString('l1_handler')
 }
 
 @Serializable
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
@@ -149,6 +149,7 @@ data class DeclareTransaction(
     override val type: TransactionType = TransactionType.DECLARE,
 ) : Transaction()
 
+@Serializable
 @SerialName("L1_HANDLER")
 data class L1HandlerTransaction(
     @SerialName("contract_address")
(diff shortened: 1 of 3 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -19,6 +19,9 @@ enum class TransactionType(val txPrefix: Felt) {
     @SerialName("INVOKE")
     @JsonNames("INVOKE_FUNCTION")
     INVOKE(Felt.fromHex("0x696e766f6b65")), // encodeShortString('invoke'),
+
+    @SerialName("L1_HANDLER")
+    L1_HANDLER(Felt.fromHex("0x6c315f68616e646c6572")) // encodeShortString('1_handler')
```
Solpatium, line 24: shouldn't it be l1_handler? (comment)

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -373,13 +373,6 @@ class GatewayProvider(
         return getBlockTransactionCount(payload)
     }
 
-    // Copied values from TransactionType in cairo-lang
```
Solpatium, line 376: Why did you remove it? I don't want to repeat the same magic string multiple times.

Solpatium (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
cptartur: Other parts of the json are hardcoded and dependent on the type: it's impossible only the `"type"` would change leaving the other values intact so this function does nothing in terms in refactoring or maintainability. If anything it makes it unclear, what exactly is the json object that is being created.

Solpatium: I don't want to repeat magic constants.

cptartur: Let's replace it with some better logic then, after #176 is merged

Solpatium (approved):

PR #180, opened 2022-11-09
Add ConvertibleToTransactionPayload interface

cptartur:
## Describe your changes

This PR attempts to standardize the behavior of `toPayload` method in Transactions by adding `ConvertibleToTransactionPayload` (subject to change) interface that indicates if Transaction (or any object for that matter) is convertible to `TransactionPayload` object. 

This PR is a suggestion, if the used approach is not preferred by other developers, it can be closed.

## Linked issues

Closes

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
@@ -34,6 +34,21 @@ sealed class Transaction {
     abstract val type: TransactionType
 }
 
+/**
+ * Implementers of this interface allow conversion to the
+ * [com.swmansion.starknet.data.types.transactions.TransactionPayload] type to be used in Providers
+ *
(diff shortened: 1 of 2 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -34,6 +34,21 @@ sealed class Transaction {
     abstract val type: TransactionType
 }
 
+/**
+ * Implementers of this interface allow conversion to the
+ * [com.swmansion.starknet.data.types.transactions.TransactionPayload] type to be used in Providers
+ *
+ * @param T: Subclass of the TransactionPayload that will be returned by the conversion
+ */
+interface ConvertibleToTransactionPayload<T : TransactionPayload> {
```
THenry14, line 43: 👀

Solpatium:
I think this would add additional cognitive overhead for users and I don't see the benefit. 
If we want every tx to have toPayload fn then I suggest adding it as an abstract method to Transaction.

PR #198, opened 2022-11-25
`ConvertibleToFeltList` interface

### LINKED ISSUE #188: Consider creating Uint256 class/utils
Uint256 is often used in starknet. We should expose something that would be easy to use for users. I think about a class, similar to Felt, we should add some interface like `interface Calldata fun toCalldata(): List<Felt>`. This would make it possible to use Uint256 directly, next to felts.

cptartur:
## Describe your changes

This PR introduces `ConvertibleToFeltList` interface to indicate that an object can be converted to list of Felts, e.g. to be used as Calldata.

It makes Uint256 implement ConvertibleToFeltList.

It makes StarkCurveSignature implement ConvertibleToFeltList.

It also does some cleanup of Uint256 class and introduces Uint256 tests.

## Linked issues

Closes #188 

## Breaking changes

* StarkCurveSignature `toList()` method replaced with `toFeltList()`.

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt
@@ -3,12 +3,14 @@
 package com.swmansion.starknet.data.types
 
 import com.swmansion.starknet.data.selectorFromName
+import com.swmansion.starknet.data.types.conversions.ConvertibleToCalldata
 import com.swmansion.starknet.data.types.transactions.InvokeTransaction
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
 typealias Calldata = List<Felt>
 typealias Signature = List<Felt>
+typealias CallArguments = List<ConvertibleToCalldata>
 
 @Serializable
 data class Call(
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt
@@ -40,6 +40,19 @@ data class Call(
         entrypoint,
         emptyList(),
     )
+
+    companion object {
+        @JvmStatic
+        fun fromCallArguments(contractAddress: Felt, entrypoint: Felt, arguments: CallArguments): Call {
+            val calldata = arguments.flatMap { it.toCalldata() }
+            return Call(contractAddress, entrypoint, calldata)
+        }
+
+        @JvmStatic
+        fun fromCallArguments(contractAddress: Felt, entrypoint: String, arguments: CallArguments): Call {
+            return fromCallArguments(contractAddress, selectorFromName(entrypoint), arguments)
+        }
+    }
 }
 
 data class ExecutionParams(
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/Felt.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/Felt.kt
@@ -1,7 +1,7 @@
 package com.swmansion.starknet.data.types
 
 import com.swmansion.starknet.data.parseHex
-import com.swmansion.starknet.extensions.*
+import com.swmansion.starknet.data.types.conversions.ConvertibleToCalldata
 import com.swmansion.starknet.extensions.toHex
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.Serializable
@@ -13,7 +13,7 @@ import kotlinx.serialization.encoding.Encoder
 import java.math.BigInteger
 
 @Serializable(with = FeltSerializer::class)
-data class Felt(val value: BigInteger) : Comparable<Felt> {
+data class Felt(val value: BigInteger) : Comparable<Felt>, ConvertibleToCalldata {
     constructor(value: Long) : this(BigInteger.valueOf(value))
     constructor(value: Int) : this(BigInteger.valueOf(value.toLong()))
 
@@ -29,6 +29,8 @@ data class Felt(val value: BigInteger) : Comparable<Felt> {
     override fun compareTo(other: Felt): Int {
         return value.compareTo(other.value)
     }
+
+    override fun toCalldata(): List<Felt> = listOf(this)
 
     override fun toString(): String {
         return "Felt(${value.toHex()})"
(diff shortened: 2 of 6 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/types/conversions/ConvertibleToFeltList.kt
```diff
@@ -0,0 +1,14 @@
+package com.swmansion.starknet.data.types.conversions
+
+import com.swmansion.starknet.data.types.Felt
+
+/**
+ * Implementers of this interface support conversions to
+ * the list of Felts.
+ */
+interface ConvertibleToFeltList {
```
Solpatium, line 9: I don't like the name of it. Maybe `StarknetSerializable`/`CalldataSerializable`? In general starknet might support some other data types in the future, I don't want to tie it to felt. We can name `toFeltList` to `toCalldata`.

Solpatium, line 9: Felt itself should support this interface. This way you will be able to mix felts with other data structures.

cptartur: `CalldataSerializable` is definitely a no, since it can be used for other things. Take a `StarknetCurveSignature` for example where it felt list representation is used as a signature.

cptartur: And how exactly that would work? Converting a felt to a list with single felt doesn't make sense at all. A user can just do `listOf(Felt)`.

cptartur: At first I wanted to do something like `CalldataConvertible` and use it in dataclasses utilizing calldata, but it was more problematic than useful: You'd have to somehow convince the compiler that the (very ergonomic) `List<Felt>` is also `CalldataConvertible` and do it in such way, that it also works in Java.

And indicating that something is convertible to Calldata when in fact you mean it's just convertible to Felt is just misleading.

Solpatium: It would return a list with a single felt. Without Felt supporting it there is no point in doing that at all. The whole reason why I want this interface is to be able to do `calldata=[Felt.ONE, Uint256(somthing), otherStrucutre]` that would be serialised by sdk.

Solpatium: Signature is also in calldata format.

Solpatium: It is a calldata for verify function

cptartur: Technically yes, but we do have a separate type for it and I want to follow that. If we want to treat Signature the same as Calldata, we'd have to `Calldata` typing instead.

THenry14 (approved):

lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt
```diff
@@ -3,11 +3,12 @@
 package com.swmansion.starknet.data.types
 
 import com.swmansion.starknet.data.selectorFromName
+import com.swmansion.starknet.data.types.conversions.ConvertibleToCalldata
 import com.swmansion.starknet.data.types.transactions.InvokeTransaction
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
-typealias Calldata = List<Felt>
+typealias Calldata = List<ConvertibleToCalldata>
```
Solpatium, line 11: Hmmm this might little be a bit misleading. I think we are missing an additional type.
Calldata - calldata as accepted on starknet - list of felts
CallArguments (might need a better name?) - list of arguments for call - list of ConvertibleToCalldata

What do you think?

cptartur: Problem is `CallArgument` doesn't make much sense in many places where `Calldata` is currently used e.g. in `DeployAccountTransactionPayload` transaction, UDC deployer etc.

We could of course decide to continue using `Calldata` there, but we kind of loose all benefits of defining this `ConvertibleToCalldata` if it's only going to be used in InvokeTransactions.

Solpatium: `DeployAccountTransactionPayload` would use `Calldata`, not `CallArguments`. I want `CallArgument` to be high level interface for creating calladata.

cptartur: So you want `CallArgument` to only be usable when creating `Call` instances?

cptartur: Actually I like this approach, we could not introduce all these changes added in the PR then, and settle on some kind of conversion for `Call` keeping the base type there `List<Felt>` in other places. 

Perhaps having a factory method `Call.fromCallArguments` or something like that that accepts this higher level interface? (ideally we'd have a constructors with both `List<Felt>` and `List<ConvertibleToCalldata>` but we can have that in jvm.)

Solpatium: What if we used Array for calldata and List for arguments? Then we could support both.

lib/src/main/kotlin/com/swmansion/starknet/data/types/conversions/ConvertibleToFeltList.kt
Solpatium: Right, just don't support CalldataSerializable with our signature class in crypto module

lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt
cptartur: I don't like this approach, forcing users to use different data types. I believe more expected solution for Java users to this situation is to use static methods.

Solpatium (approved):
Beautiful 🤩 Please create an issue to document it nicely.

PR #199, opened 2022-11-25
Cleanup `data` module

cptartur:
## Describe your changes

This PR cleans up `data` module by:

* moving serializers to separate files in serializers module
* removing unused data classes 
* moving data class files (namely `Call`)

## Linked issues

Closes

## Breaking changes

* Import paths changed
* Removed some classes (unused in library)


THenry14 (approved):

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/BlockIdSerializer.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.JsonEncoder
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.encodeToJsonElement
+import kotlinx.serialization.json.put
+
+internal object BlockIdSerializer : KSerializer<BlockId> {
```
Solpatium, line 17: JvmSynthetic for internals?

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Call.kt
```diff
@@ -0,0 +1,58 @@
+package com.swmansion.starknet.data.types.transactions
+
+import com.swmansion.starknet.data.selectorFromName
+import com.swmansion.starknet.data.types.Calldata
+import com.swmansion.starknet.data.types.Felt
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class Call(
```
Solpatium, line 10: Why was this moved? This is related to execution, it is not a transaction.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -588,7 +587,7 @@ class ProviderTest {
         val payload =
             DeclareTransactionPayload(
                 ContractDefinition(contents),
-                DECLARE_SENDER_ADDRESS,
+                Felt.ONE, // Declare tx version 0 has a sender address of 0x1
```
Solpatium, line 590: I don't like replacing properly named constant with a comment. Code > comments.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Contracts.kt
```diff
@@ -2,78 +2,8 @@ package com.swmansion.starknet.data.types
...
-
-@Serializable
-data class AbiEntry(
-    val name: String,
-    val type: AbiEntryType,
-)
-
-@Serializable
-sealed class AbiElement
-
```
Solpatium, line 24: all that is no longer used due to deploy remove, right?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -46,16 +46,6 @@ data class DeployAccountResponse(
...
 
-@Serializable
-data class GetStorageAtResponse(
-    val result: Felt,
-)
-
-data class TransactionFailureReason(
-    val code: String,
-    val errorMessage: String,
-)
```
Solpatium, line 57: There are not used anywhere anymore? 🤔

Solpatium (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Call.kt
cptartur: It's a core part of `InvokeTransactionPayload` so I think it makes sense to keep them in the same module.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
cptartur: Generally I agree, but this constant only made sense in v0 declare transactions. Ideally we'll update our tests to only create declare v1 and this will be replaced entirely.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Contracts.kt
cptartur: It wasn't used by any part of library even before I removed it. I think these got declared a while ago and finally were never used in any module

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Call.kt
Solpatium: But you also use call without **any transaction**, so having it here is misleading.

cptartur: You are correct about that. I guess the problem is it doesn't make much sense to bundle it with all other data classes in `execution` since it's used in many different places.

Perhaps we should just put `Call.kt` in top level `types` module to indicate that it is it's own separate thing?

Solpatium: I started execution file as container for things related to execution and call is one of them. I see that over time we added things that are less relevant there, like GetBlockTransactionCountPayload. Maybe we should move the others to proper places and leave execution things as they are (Call, Calldata, ExecutionParams etc)?

cptartur: Perhaps having `Payloads.kt` file for all payloads defined in `Exectution.kt` would make sense. We could then keep call etc. in execution.

Though I don't know if it makes more sense than to extract call (and maybe calldata etc.) to separate file, keeping execution as it is?

Solpatium: For me the problem is that payloads don't match execution category, so I would rather assign them to specific categories.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Contracts.kt
THenry14: This will actually come in handy in rpc 2.1, so let's leave it. I will remove the unnecessary parts (if there will be some)

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/BlockIdSerializer.kt
cptartur: You cannot have this annotation on classes/objects unfortunately.

Solpatium (approved):
👌

THenry14 (approved):

Issue #204, opened 2022-12-07
Support for testnet2


THenry14:
### Feature Request

Support for testnet2

PR #205, opened 2022-12-07
Add support for testnet2

THenry14:
## Describe your changes

Add support for testnet2

## Linked issues

Closes https://github.com/software-mansion/starknet-jvm/issues/204

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
@@ -1,8 +1,8 @@
 package com.swmansion.starknet.provider.gateway
 
 import com.swmansion.starknet.data.NetUrls.MAINNET_URL
-import com.swmansion.starknet.data.NetUrls.TESTNET_URL
 import com.swmansion.starknet.data.NetUrls.TESTNET2_URL
+import com.swmansion.starknet.data.NetUrls.TESTNET_URL
 import com.swmansion.starknet.data.serializers.*
 import com.swmansion.starknet.data.types.*
 import com.swmansion.starknet.data.types.transactions.*
@@ -20,7 +20,6 @@ import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.SerializationException
 import kotlinx.serialization.json.*
-import java.lang.IllegalArgumentException
 import java.util.function.Function
 
 /**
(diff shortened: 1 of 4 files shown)


lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -439,23 +441,38 @@ class GatewayProvider(
...
     companion object Factory {
         @JvmStatic
-        fun makeTestnetClient(): GatewayProvider {
-            return GatewayProvider(
-                "$TESTNET_URL/feeder_gateway",
-                "$TESTNET_URL/gateway",
-                StarknetChainId.TESTNET,
-            )
-        }
+        fun makeTestnetClient(testnetId: StarknetChainId = StarknetChainId.TESTNET): GatewayProvider =
```
Solpatium, line 444: Default params don't translate well to java. Please overload methods instead.

Solpatium (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
THenry14: done

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -19,6 +20,7 @@ import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.SerializationException
 import kotlinx.serialization.json.*
+import java.lang.IllegalArgumentException
 import java.util.function.Function
```
cptartur, line 24: Let's not use exceptions from java.lang, but directly from kotlin.

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -447,6 +449,22 @@ class GatewayProvider(
             )
         }
 
+        @JvmStatic
+        fun makeTestnetClient(testnetId: StarknetChainId): GatewayProvider =
```
cptartur, line 453: Should we have some kind of unit tests for these constructors?

cptartur (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
THenry14: Up until now we didn't, not sure what we could do except maybe just create a client and check its chainId (+ I'm really not sure how to test custom httpService, if at all)

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
THenry14: Changed to `kotlin.IllegalArgumentException`

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
cptartur: Perhaps let's create an issue for it then and we can cover it sometime later.

THenry14: https://github.com/software-mansion/starknet-jvm/issues/206

Solpatium (approved):

Issue #189, opened 2022-11-22
Update actions


THenry14:
https://github.com/software-mansion/starknet-jvm/actions/runs/3524137325

THenry14:
https://github.com/software-mansion/starknet-jvm/pull/207

There is also a problem with gcc7. Let's consider if ami workaround can be dropped, or we need to work around it somehow different.

https://github.com/software-mansion/starknet-jvm/issues/171

Issue #171, opened 2022-11-01
starknet-jvm doesn't work on Amazon Linux 2 ARM


Solpatium:
Versions of GLIBCXX supported by Amazon Linux 2:
![image](https://user-images.githubusercontent.com/10278469/199220191-5ec75a7e-0783-4307-bfe2-d137233bef50.png)

Our native component uses symbol
```
0000000000000000      DF *UND*	0000000000000000 (GLIBCXX_3.4.26) _ZNSt7__cxx1119basic_ostringstreamIcSt11char_traitsIcESaIcEEC1Ev
```
which requires version 3.4.26, two minor versions above supported by AL2. We have to use older libstdc++ version during build to support Amazon Linux 2. Using GCC 7 should fix the issue.

Solpatium:
I've fixed it for amd64 architecture (https://github.com/software-mansion/starknet-jvm/pull/172), we should also fix it for arm probably.

cptartur:
Should we close this issue and create a more specific one for ARM? This is kind of misleading to have `starknet-jvm doesn't work on Amazon Linux 2` in issues which is not 100% the case now.

PR #213, opened 2022-12-15
Support for older glibc for arm + update actions

THenry14:
## Describe your changes

This PR adds support for older glibc (3.4.24) for arm architecture + updates gh actions to use newest versions without deprecation warnings. 
It also fixes ubuntu runner version to 20.04, as this is the newest version that easily supports g++-7; It still has ~3 years of support so we should be good. After amazon updates their Linux distros to use newer glibc versions we should think about also updating our runners and dropping older glibc.

## Linked issues

Closes https://github.com/software-mansion/starknet-jvm/issues/189
Closes https://github.com/software-mansion/starknet-jvm/issues/171

## Breaking changes

### DIFF
--- a/.github/workflows/Dockerfile.multiarch
+++ b/.github/workflows/Dockerfile.multiarch
@@ -4,9 +4,12 @@ RUN curl https://apt.corretto.aws/corretto.key > corretto.key \
     && apt-key add corretto.key \
     && add-apt-repository 'deb https://apt.corretto.aws stable main' \
     && apt update \
-    && apt install -y cmake clang java-11-amazon-corretto-jdk \
+    && apt install -y cmake clang java-11-amazon-corretto-jdk g++-7 \
+    && find /usr/lib/gcc/x86_64-linux-gnu/ -maxdepth 1 -mindepth 1 ! -name 7 -delete \
     && mkdir /build
 ENV JAVA_HOME=/usr/lib/jvm/java-11-amazon-corretto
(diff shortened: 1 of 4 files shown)


.github/workflows/Dockerfile.multiarch
```diff
@@ -4,9 +4,12 @@ RUN curl https://apt.corretto.aws/corretto.key > corretto.key \
     && apt-key add corretto.key \
     && add-apt-repository 'deb https://apt.corretto.aws stable main' \
     && apt update \
-    && apt install -y cmake clang java-11-amazon-corretto-jdk \
+    && apt install -y cmake clang java-11-amazon-corretto-jdk g++-7 \
+    && find /usr/lib/gcc/aarch64-linux-gnu/ -maxdepth 1 -mindepth 1 ! -name 7 -exec rm -rf {} + \
```
Solpatium, line 8: What does it to? Maybe a comment could be useful 😅

Solpatium (approved):
LGTM with comments

.github/workflows/Dockerfile.multiarch
THenry14: Hmm, I thought it was kinda self explenatory, but sure I can add a comment :+1:

Solpatium: Sorry I don't speak bash :D. Also make sure to explain why it is needed.

PR #218, opened 2022-12-15
Update docs and GatewayProvider factories

### LINKED ISSUE #98: Improve [package specific] docs
It is possible to document specific packages by adding `# Package full.path.to.package` sections to `starknet-jvm.md`. We can consider including example usages and more details about specific library modules there.

cptartur:
## Describe your changes

This PR extends the documentation.

It also renames GatewayProvider.Factory methods to `make_Provider` from `make_Client`.

## Linked issues

Closes #98 

## Breaking changes

- Renamed GatewayProvider.Factory methods to `make_Provider` from `make_Client`.


javademo/src/main/java/com/example/javademo/Main.java
```diff
@@ -24,7 +24,7 @@ public static void main(String[] args) throws Exception {
         // ⚠️ WARNING ⚠️ The key generated here is just for demonstration purposes.
         // DO NOT GENERATE YOUR KEYS THIS WAY. USE CRYPTOGRAPHICALLY SAFE TOOLS!
         Felt privateKey = new Felt(ThreadLocalRandom.current().nextLong(1, Long.MAX_VALUE));
-        Provider provider = GatewayProvider.makeTestnetClient();
+        Provider provider = GatewayProvider.makeTestnetProvider();
```
cptartur, line 27: Problem is, this way Javademo is up to date with changes, but cannot be run if using package version specified in its build.gradle.

lib/starknet-jvm.md
```diff
@@ -236,21 +238,72 @@ Data classes representing StarkNet transactions.
...
+```
+
+In the case of `Request.sendAsync()`, an exception would have to be handled in the returned `CompletableFuture`.
+
 # Package com.swmansion.starknet.provider.gateway
 
 Provider utilising StarkNet gateway and feeder gateway for communication with the network.
 
+```java
+// Create a provider using a factory
```
Solpatium, line 282: This is not really a factory. It might be called factory method. I like to call methods like this named constructors.

Solpatium (approved):

lib/starknet-jvm.md
cptartur: Let's just call it static method

THenry14 (approved):

Issue #216, opened 2022-12-15
StarknetCurve.getPublicKey(Felt.ZERO) throws crypto-cpp error


bartekryba:
StarknetCurve.getPublicKey(Felt.ZERO) throws crypto-cpp error. This is probably expected, but it could be verified earlier than in the native module.

PR #226, opened 2023-01-16
Throw an exception when illegal argument is passed in StarkCurve.getPublicKey

THenry14:
## Describe your changes

Added check in StarknetCurve.getPublicKey that fails if privateKey value is wrong, so exception is thrown from sdk rather than crypto-cpp.

## Linked issues

Closes https://github.com/software-mansion/starknet-jvm/issues/216

## Breaking changes

### DIFF
--- a/lib/src/test/kotlin/starknet/crypto/StarknetCurveTest.kt
+++ b/lib/src/test/kotlin/starknet/crypto/StarknetCurveTest.kt
@@ -95,6 +95,13 @@ internal class StarknetCurveTest {
     }
 
     @Test
+    fun getPublicKeyInvalid() {
+        assertThrows(IllegalArgumentException::class.java) {
+            StarknetCurve.getPublicKey(Felt.ZERO)
+        }
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


lib/src/test/kotlin/starknet/crypto/StarknetCurveTest.kt
```diff
@@ -94,6 +94,13 @@ internal class StarknetCurveTest {
         assertEquals(PUBLIC_KEY, publicKey)
     }
 
+    @Test
+    fun getPublicKeyInvalid() {
```
cptartur, line 98: Nitpick: you could wrap the test name in backticks and add spaces around the words.

cptartur (approved):

lib/src/test/kotlin/starknet/crypto/StarknetCurveTest.kt
THenry14: Yup, but I just wanted to conform to the names that are already in this file

Issue #224, opened 2023-01-12
Accept any payload in Providers' methods


THenry14:
### Feature Request

https://github.com/software-mansion/starknet-jvm/pull/212#discussion_r1067853170

PR #227, opened 2023-01-17
Accept any payload in getEstimateFee

THenry14:
## Describe your changes

Change getEstimateFee interface, to accept any valid payload as TransactionPayload

## Linked issues

Closes https://github.com/software-mansion/starknet-jvm/issues/224

## Breaking changes

- [X] This issue contains breaking changes
- change of interface of `getEstimateFee`

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionPayloadPolymorphicSerializer.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionPayloadPolymorphicSerializer.kt
@@ -1,0 +1,16 @@
+package com.swmansion.starknet.data.serializers
+
+import com.swmansion.starknet.data.types.transactions.*
+import kotlinx.serialization.DeserializationStrategy
+import kotlinx.serialization.json.*
+
+internal object JsonRpcTransactionPayloadPolymorphicSerializer :
+    JsonContentPolymorphicSerializer<TransactionPayload>(TransactionPayload::class) {
+    override fun selectDeserializer(element: JsonElement): DeserializationStrategy<out TransactionPayload> =
+        when (element.jsonObject["type"]?.jsonPrimitive?.content) {
+            "INVOKE" -> InvokeTransactionPayload.serializer()
+            "DECLARE" -> DeclareTransactionPayload.serializer()
+            "DEPLOY_ACCOUNT" -> DeployAccountTransactionPayload.serializer()
+            else -> throw IllegalArgumentException("Invalid transaction type '${element.jsonObject["type"]?.jsonPrimitive?.content}'")
+        }
+}
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionPayload.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionPayload.kt
@@ -1,8 +1,12 @@
 package com.swmansion.starknet.data.types.transactions
 
+import com.swmansion.starknet.data.serializers.JsonRpcTransactionPayloadPolymorphicSerializer
 import com.swmansion.starknet.data.types.*
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+
+@Serializable(with = JsonRpcTransactionPayloadPolymorphicSerializer::class)
+sealed class TransactionPayload
 
 @Serializable
 data class InvokeTransactionPayload constructor(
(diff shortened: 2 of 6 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionPayload.kt
```diff
@@ -1,9 +1,13 @@
 package com.swmansion.starknet.data.types.transactions
 
+import com.swmansion.starknet.data.serializers.JsonRpcTransactionPayloadPolymorphicSerializer
 import com.swmansion.starknet.data.types.*
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
+@Serializable(with = JsonRpcTransactionPayloadPolymorphicSerializer::class)
+sealed class TransactionPayload
```
cptartur, line 9: So you managed to get the `@Serializable(with = ...)` syntax working. 👍 I remember I had some issues with that previously.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionPayloadPolymorphicSerializer.kt
```diff
@@ -0,0 +1,16 @@
...
+import kotlinx.serialization.json.*
+
+internal object JsonRpcTransactionPayloadPolymorphicSerializer :
+    JsonContentPolymorphicSerializer<TransactionPayload>(TransactionPayload::class) {
+    override fun selectDeserializer(element: JsonElement): DeserializationStrategy<out TransactionPayload> =
+        when (element.jsonObject["type"]?.jsonPrimitive?.content) {
+            "INVOKE" -> InvokeTransactionPayload.serializer()
+            "DECLARE" -> DeclareTransactionPayload.serializer()
+            "DEPLOY_ACCOUNT" -> DeployAccountTransactionPayload.serializer()
+            else -> throw IllegalArgumentException("Invalid transaction type '${element.jsonObject["type"]?.jsonPrimitive?.content}'")
```
cptartur, line 14: If either of `?` calls is not successful, I believe the resulting string will be `"Invalid transaction type 'null'`. Do you think a default value other than that would be useful? It could be provided inside the ${} block using `:? "default"`

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -270,82 +270,33 @@ class GatewayProvider(
...
-        payload: InvokeTransactionPayload,
+        payload: TransactionPayload,
         blockId: BlockId,
     ): Request<EstimateFeeResponse> {
         val url = feederGatewayRequestUrl("estimate_fee")
-        val body = serializeInvokeTransactionPayload(payload)
+        val body = when (payload) {
+            is InvokeTransactionPayload -> serializeInvokeTransactionPayload(payload)
+            is DeployAccountTransactionPayload -> serializeDeployAccountTransactionPayload(payload)
+            is DeclareTransactionPayload -> Json.encodeToJsonElement(DeclareTransactionPayloadSerializer, payload).jsonObject
```
cptartur, line 280: Why do we need to call a function for invoke and deploy and declare is just encoded using serializer?

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionPayloadPolymorphicSerializer.kt
THenry14: I figured, that if that's the case we should be fine with just 'null', since it actually will be null; pointing us to the fact that the response is invalid; perhaps the message should be changed to reflect that somehow?

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
THenry14: First two are defined in GatewayProvider, and all they do is basically build json object specific to gateway. 
For declare, there is a serializer defined in serializers that works for both rpc and gateway, so in order to not have a boiler plate and reuse that serializer we need to call it a bit differently.

cptartur: Okay, makes sense.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionPayloadPolymorphicSerializer.kt
cptartur: Maybe creating handling for cases that are not very common like this is not entirely needed. I though that maybe we could have a separate error messages for a) type not being valid b) type missing entirely. But if you think it's not necessary, we can leave it as it is.

cptartur (approved):

PR #229, opened 2023-01-18
Enable parallel test execution

### LINKED ISSUE #228: Explore a possibility to run tests in parallel
### Feature Request

It should be possible to parallelize at least some of the tests that are not state dependent.

cptartur:
## Describe your changes

This PR configures most tests modules to run in parallel. 

StandardAccount and StandardDeployer still run in single thread, because parallel execution caused the tests to fail because of the invalid nonces.

## Linked issues

Closes #228 

## Breaking changes

### DIFF
--- a/lib/src/test/kotlin/starknet/utils/DevnetClient.kt
+++ b/lib/src/test/kotlin/starknet/utils/DevnetClient.kt
@@ -29,8 +29,8 @@ class DevnetClient(
     private val host: String = "0.0.0.0",
     private val port: Int = 5050,
     private val httpService: HttpService = OkHttpService(),
+    private val accountDirectory: Path = Paths.get("src/test/resources/account"),
 ) : AutoCloseable {
-    private val accountDirectory = Paths.get("src/test/resources/account")
     private val baseUrl: String = "http://$host:$port"
     private val seed: Int = 1053545547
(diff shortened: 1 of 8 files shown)


cptartur:
This works on my device, but it seems to be failing in checks. I suspect that's because of some interactions between devnet instances but it's yet to be verified.

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -29,8 +29,8 @@ class DevnetClient(
     private val host: String = "0.0.0.0",
     private val port: Int = 5050,
     private val httpService: HttpService = OkHttpService(),
+    private val accountDirectory: Path = Paths.get("src/test/resources/account"),
 ) : AutoCloseable {
```
cptartur, line 33: I think it's a fine initial solution. One idea for possible refinement, would be to use random UUID for file name on every instance creation, so we wouldn't have to explicitly set up a path per every test module (we could create an issue for that if you think that's useful).

THenry14: I the way we have it now makes it easier to debug some account related things later on (eg right away you see what were the addresses for providerTest)

THenry14 (approved):

PR #230, opened 2023-01-20
Use different versioning for fee estimation

THenry14:
## Describe your changes

Use different versioning for fee estimation

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
@@ -34,7 +34,7 @@ interface Account {
      * @param params additional execution parameters for the transaction
      * @return signed invoke function payload
      */
-    fun sign(calls: List<Call>, params: ExecutionParams): InvokeTransactionPayload
+    fun sign(calls: List<Call>, params: ExecutionParams, forFeeEstimate: Boolean = false): InvokeTransactionPayload
 
     /**
      * Sign deploy account transaction.
@@ -51,6 +51,7 @@ interface Account {
         calldata: Calldata,
         salt: Felt,
         maxFee: Felt,
+        forFeeEstimate: Boolean = false,
     ): DeployAccountTransactionPayload
 
     /**
@@ -67,6 +68,7 @@ interface Account {
         contractDefinition: ContractDefinition,
         classHash: Felt,
         params: ExecutionParams,
+        forFeeEstimate: Boolean = false,
     ): DeclareTransactionPayload
 
     /**
--- a/lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
@@ -19,7 +19,7 @@ interface Account {
      *
      * @param call a call to be signed
      * @param params additional execution parameters for the transaction
-     * @param forFeeEstimate when set to 'true', it changes the version to `2^128+version` so the signed transaction can only be used for fee estimation
+     * @param forFeeEstimate when set to `true`, it changes the version to `2^128+version` so the signed transaction can only be used for fee estimation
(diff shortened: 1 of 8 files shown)


lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
```diff
@@ -19,10 +19,11 @@ interface Account {
      *
      * @param call a call to be signed
      * @param params additional execution parameters for the transaction
+     * @param forFeeEstimate parameter indicating whether the sign will be only used for cost estimation
```
cptartur, line 22:
```suggestion
     * @param forFeeEstimate setting it to `true` changes the signed transaction version so it can only be used for fee estimation
```

cptartur: I wonder if we should add a note somewhere to explain what version should be used in case someone is implementing the Account interface themselves.

lib/src/main/kotlin/com/swmansion/starknet/account/StandardAccount.kt
```diff
@@ -23,6 +24,7 @@ class StandardAccount(
     private val provider: Provider,
 ) : Account {
     private val version = Felt.ONE
+    private val estimateVersion: BigInteger = BigInteger.valueOf(2).pow(128).add(BigInteger(version.decString()))
```
cptartur, line 27:
```suggestion
    private val estimateVersion: BigInteger = BigInteger.valueOf(2).pow(128).add(version.value)
```

I think you can use the underlying BigInteger directly here.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -306,16 +307,17 @@ object TransactionFactory {
         nonce: Felt,
         maxFee: Felt = Felt.ZERO,
         signature: Signature = emptyList(),
+        version: Felt = INVOKE_VERSION,
```
cptartur, line 310: Parameters with default values don't really work well in Java. Let's verify how this is translated to java. 
If it just requires explicitly passing version parameter every time, it's fine by me.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -19,6 +19,7 @@ import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.MethodSource
```
cptartur, line 19: Maybe it'd be also useful to add some kind of a test verifying that trying to execute transaction signed for fee estimation fails?

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
THenry14:
```
* @param forFeeEstimate when set to 'true', it changes the version to `2^128+version` so the signed transaction can only be used for fee estimation
```
🤔 ?

cptartur: Sounds ok!

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
```diff
@@ -19,10 +19,11 @@ interface Account {
      *
      * @param call a call to be signed
      * @param params additional execution parameters for the transaction
+     * @param forFeeEstimate when set to 'true', it changes the version to `2^128+version` so the signed transaction can only be used for fee estimation
```
cptartur, line 22:
```suggestion
     * @param forFeeEstimate when set to `true`, it changes the version to `2^128+version` so the signed transaction can only be used for fee estimation
```

cptartur (approved):

PR #231, opened 2023-01-26
:sparkles: Split PENDING and RECEIVED transaction statuses

florian-bellotti:
The TransactionStatus enumeration should include a distinction between PENDING and RECEIVED as they represent different statuses.


THenry14:
Hi @florian-bellotti ,

We have mapped both statuses as `pending`, because there is no `received` status in RPC spec (you can find it [here](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L1881)), and it made more sense since we aim to support RPC first and foremost (especially since the gateway will be eventually dropped).

That's why I have to reject your proposed changes, but thank you for the contribution :+1:

PR #233, opened 2023-02-08
Refactor transaction serialization in gateway

### LINKED ISSUE #220: Refactor transaction serialization in gateway
### Feature Request

https://github.com/software-mansion/starknet-jvm/pull/212#discussion_r1054198311

The input is being encoded, decoded and then encoded egain to achieve two serialization techniques (transform -> polymorphic). We should try and do it somehow better if possible.

cptartur:
## Describe your changes

as in title

## Linked issues

Closes #220 

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/GatewayTransactionTransformingSerializer.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/GatewayTransactionTransformingSerializer.kt
@@ -1,0 +1,15 @@
+package com.swmansion.starknet.data.serializers
+
+import com.swmansion.starknet.data.types.transactions.Transaction
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonTransformingSerializer
+import kotlinx.serialization.json.jsonObject
+
+object GatewayTransactionTransformingSerializer :
+    JsonTransformingSerializer<Transaction>(TransactionPolymorphicSerializer) {
+    override fun transformDeserialize(element: JsonElement): JsonElement {
+        return element.jsonObject["transaction"]
+            ?: throw SerializationException("Provided input does not contain a transaction - \"transaction\" key missing")
+    }
+}
--- a/lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
@@ -120,15 +120,6 @@ class GatewayProvider(
         return body
     }
 
-    private fun transformResponseForTransactionAndSerialize(body: String): Transaction {
-        try {
-            val transformedResponse = json.parseToJsonElement(body).jsonObject.getValue("transaction")
-            return json.decodeFromJsonElement(TransactionPolymorphicSerializer, transformedResponse)
-        } catch (e: Exception) {
-            throw RequestFailedException(payload = "Response is not a valid transaction", message = body)
-        }
-    }
-
     override fun callContract(call: Call, blockTag: BlockTag): Request<List<Felt>> {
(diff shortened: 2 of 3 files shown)


lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -122,10 +122,9 @@ class GatewayProvider(
 
     private fun transformResponseForTransactionAndSerialize(body: String): Transaction {
         try {
-            val transformedResponse = json.parseToJsonElement(body).jsonObject["transaction"]
-            val encodedResponse = json.encodeToString(transformedResponse)
-            return json.decodeFromString(TransactionPolymorphicSerializer, encodedResponse)
-        } catch (e: IllegalArgumentException) {
+            val transformedResponse = json.parseToJsonElement(body).jsonObject.getValue("transaction")
```
THenry14, line 125: have you tried combining transfirming and polymorphic serializers? I guess this would be an ideal way to refactor. I have attempted it while doind rpc2.1 but failed, and did want to spend too much time on it;

With your proposed changes we still convert string to json, and then again to string.

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
```diff
@@ -122,10 +122,9 @@ class GatewayProvider(
...
+            return json.decodeFromJsonElement(TransactionPolymorphicSerializer, transformedResponse)
+        } catch (e: Exception) {
```
THenry14, line 127: Isn't it too broad?

cptartur: I though the same too. However, kotlin doesn't (why???) allow catching multiple exceptions in one catch. And `getValue` raises `NoSuchElementException`. 

We could have two catch blocks (duplications), or it can be changed to `json.parseToJsonElement(body).jsonObject["transaction"] :? throw IllegalArgumentException()` so we have a single catch.

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
cptartur: I actually the same issues a while before rpc2.1. But I can try it again.

lib/src/main/kotlin/com/swmansion/starknet/provider/gateway/GatewayProvider.kt
THenry14: Hmm I see. Maybe we could catch Exception and then decide what to do in `when` block? It is not perfect, but we could still throw `RequestFailedException` with specific, custom info, so maybe it's worth it

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/GatewayTransactionTransformingSerializer.kt
```diff
@@ -0,0 +1,15 @@
+package com.swmansion.starknet.data.serializers
+
+import com.swmansion.starknet.data.types.transactions.Transaction
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonTransformingSerializer
+import kotlinx.serialization.json.jsonObject
+
+object GatewayTransactionTransformingSerializer :
+    JsonTransformingSerializer<Transaction>(TransactionPolymorphicSerializer) {
```
cptartur, line 10: Don't know if there's a better way of doing this, but turns out you can just pass any serializer you won't to be used after transformation as argument. Don't know why I didn't think of that before.

THenry14: 🪄

THenry14 (approved):

PR #237, opened 2023-02-09
Support signing TypedData

cptartur:
## Describe your changes

This PR adds support for signing TypedData both in `Account` and `Signer` interfaces

## Linked issues

Closes

## Breaking changes

* Account interface added `signTypedData` and `verifyTypedDataSignature`
* Signer interface added `signTypedData`

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
@@ -1,5 +1,6 @@
 package com.swmansion.starknet.account
 
+import com.swmansion.starknet.data.TypedData
 import com.swmansion.starknet.data.types.*
 import com.swmansion.starknet.data.types.transactions.*
 import com.swmansion.starknet.provider.Request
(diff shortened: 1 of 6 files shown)


lib/src/main/kotlin/com/swmansion/starknet/signer/Signer.kt
```diff
@@ -15,10 +16,16 @@ interface Signer {
...
-    // TODO: add more params
     fun signTransaction(transaction: Transaction): Signature
 
+    /**
+     * Sign TypedData
+     *
+     * @param typedData TypedData instance to sign
+     * @return a signature of provided typedData
+     */
+    fun signTypedData(typedData: TypedData, accountAddress: Felt): Signature
```
bartekryba, line 27: Missing param comment for accountAddress

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
```diff
@@ -76,6 +77,23 @@ interface Account {
...
+
+    /**
+     * Verify a signature of TypedData on StarkNet
+     *
+     * @param typedData a TypedData instance which signature will be verified
+     * @param signature a signature of typedData
+     * @return `true` if signature is valid, `false` otherwise
+     */
+    fun verifyTypedDataSignature(typedData: TypedData, signature: Signature): Request<Boolean>
+
```
THenry14, line 96: why not just `sign` and `verify`?

lib/src/main/kotlin/com/swmansion/starknet/account/StandardAccount.kt
```diff
@@ -106,6 +109,45 @@ class StandardAccount(
...
+                        return@handle handleValidationError(exception)
+                    }
+                    return@handle result[0] > Felt.ZERO
+                }
+            }
+        }
+    }
+
+    private fun handleValidationError(e: RequestFailedException): Boolean {
+        val regex = """Signature\s.+,\sis\sinvalid""".toRegex()
```
THenry14, line 144: Maybe add some comment about what this regex does?

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
cptartur: I wanted to make `signTypedData` a distinct method, similarly to `signDeclare` etc. `sign` is already used for InvokeTransactions.

`verifyTypedDataSignature` followed that convention.

THenry14 (approved):

PR #263, opened 2023-07-03
Support for cairo1 account calldata

### LINKED ISSUE #261: `callsToExecuteCalldata` for Cairo 1 contracts
The way how calls are transformed into calldata has changed in Cairo 1.

Implement a function to build Cairo 1 calldata.

For reference, see:
https://github.com/0xs34n/starknet.js/blob/4b2fa60d932408eb896baa3daabd87971e5258d0/src/utils/transaction.ts#L85

bartekryba:
## Describe your changes

Support for cairo1 account calldata

## Linked issues

Closes #261 

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt
@@ -97,7 +97,23 @@ data class ExecutionParams(
 )
 
 @JvmSynthetic
-internal fun callsToExecuteCalldata(calls: List<Call>): List<Felt> {
+private fun callsToExecuteCalldataCairo1(calls: List<Call>): List<Felt> {
+    val result = mutableListOf<Felt>()
+
+    result.add(Felt(calls.size))
+
+    for (call in calls) {
+        result.add(call.contractAddress)
+        result.add(call.entrypoint)
+        result.add(Felt(call.calldata.size))
+        result.addAll(call.calldata)
+    }
+
+    return result
+}
+
+@JvmSynthetic
+private fun callsToExecuteCalldataCairo0(calls: List<Call>): List<Felt> {
     val wholeCalldata = mutableListOf<Felt>()
     val callArray = mutableListOf<Felt>()
     for (call in calls) {
@@ -116,3 +132,13 @@ internal fun callsToExecuteCalldata(calls: List<Call>): List<Felt> {
         addAll(wholeCalldata)
     }
 }
+
+@JvmSynthetic
+internal fun callsToExecuteCalldata(calls: List<Call>, cairoVersion: Felt = Felt.ZERO): List<Felt> {
+    if (cairoVersion == Felt.ONE) {
+        return callsToExecuteCalldataCairo1(calls)
+    }
+
+    return callsToExecuteCalldataCairo0(calls)
+}
+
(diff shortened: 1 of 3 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt
```diff
@@ -97,7 +97,23 @@ data class ExecutionParams(
...
-internal fun callsToExecuteCalldata(calls: List<Call>): List<Felt> {
+private fun callsToExecuteCalldataCairo1(calls: List<Call>): List<Felt> {
+    val result = mutableListOf<Felt>()
+
+    result.add(Felt(calls.size))
+
+    for (call in calls) {
+        result.add(call.contractAddress)
+        result.add(call.entrypoint)
+        result.add(Felt(call.calldata.size))
```
THenry14, line 108: are you sure size should be included here?

bartekryba: From what I understand about execute calldata in these new contracts, that's how it should be constructed now. Without calldata size, there wouldn't be an option to determine how many calldata arguments are in the call

THenry14 (approved):

lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt
kamilargent: I can confirm that's the correct implementation 👍

lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt
```diff
@@ -97,7 +97,23 @@ data class ExecutionParams(
 )
 
 @JvmSynthetic
-internal fun callsToExecuteCalldata(calls: List<Call>): List<Felt> {
+private fun callsToExecuteCalldataCairo1(calls: List<Call>): List<Felt> {
```
kamilargent, line 100: Could u please expose both of `callsToExecuteCalldataX` as a global function. We are using it on our end and now that we have 2 version of it, I would want to avoid copy pasting it into our codebase.

bartekryba: Sure

bartekryba: Do you actually need both of the subfunctions, or is exposing just `callsToExecuteCalldata` enough?

kamilargent: Temporarily we need both - one of the contracts that we have still uses the old encoding

bartekryba: But that function supports both versions:

```kotlin
@JvmSynthetic
 internal fun callsToExecuteCalldata(calls: List<Call>, cairoVersion: Felt = Felt.ZERO): List<Felt> {
     if (cairoVersion == Felt.ONE) {
         return callsToExecuteCalldataCairo1(calls)
     }

     return callsToExecuteCalldataCairo0(calls)
 }
```

kamilargent: Oh, that's what u meant. In that case - yeah just exposing this method to the clients will be enough

PR #266, opened 2023-07-17
Support estimate_message_fee when implemented in rpc (RPC 0.3.1)

### LINKED ISSUE #253: Support estimate_message_fee when implemented in RPC (RPC 0.3.1)
### Feature Request

A rough estimate is it could be done "in a few weeks". Blocked until then.

DelevoXDG:
## Describe your changes

- Bump starknet-devnet to `0.5.5`
- Update Messages to match openrpc `0.4.0` spec
- Add getEstimateMessageFee (not fully tested: not currently implemented on devnet)
- Test w/ mocked response

## Linked issues

Closes #253 

## Breaking changes

Changed Messages

### DIFF
--- a/lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
+++ b/lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
@@ -178,7 +178,7 @@ class StandardAccountTest {
         // Note to future developers experiencing failures in this test. Compiled contract format sometimes
         // changes, this causes changes in the class hash.
         // If this test starts randomly falling, try recalculating class hash.
-        val classHash = Felt.fromHex("0x48e8cd3b3d95cb6d9cbc9b3c0a39b1a7341f04b7daa5a13461dbc3116c5423")
+        val classHash = Felt.fromHex("0x37475b8cd1e7360416bae6bc332ba4bc50936cd2d0f9f2207507acbac172e8d")
         val declareTransactionPayload = account.signDeclare(
             contractDefinition,
             classHash,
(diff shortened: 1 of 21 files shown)


lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -201,6 +201,12 @@ class StandardAccountTest {
         assertNotNull(feeEstimate)
     }
 
+    @Test
+    fun `estimate message fee`() {
+        // TODO: test provider.getEstimateMessageFee
+        // BLOCKED: pending devnet update
+    }
+
```
THenry14, line 209: any idea when devnet will be updated?
how were your changes tested?

DelevoXDG: Unfortunately, no idea. They weren't tested so probably should be blocked until the testing is possible.

THenry14: I see. We could wait, but I can't even see a PR for this in devnet and it would be a shame to hold on for who knows how long.

My suggestion is to mock this test case for now. If you want to, you can even run this against testnet to see what is an example response. For some example on ocked tests, you can check ProviderTests. To get rid of this mocks, let's create a new issue and deal with it when devnet support is in place.

What do you think about this?

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -201,6 +201,62 @@ class StandardAccountTest {
...
+        assertEquals(gasPrice, response.gasPrice)
+        assertEquals(gasConsumed, response.gasConsumed)
+        assertEquals(overallFee, response.overallFee)
+    }
+
+    @Test
+    fun `estimate message fee`() {
+        // TODO: test provider.getEstimateMessageFee
+        // BLOCKED: pending devnet update
+    }
```
THenry14, line 258: let's remove this

THenry14 (approved):
🔥

PR #267, opened 2023-07-17
Bump deps

### LINKED ISSUE #260: Bump deps
### Feature Request

at least bouncycastle

DelevoXDG:
## Describe your changes

- Bump starknet-devnet to 0.5.5 
- Bump gradle, deps, plugins, android sdk

## Linked issues

Closes #260 

## Breaking changes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,16 +2,16 @@ plugins {
     id 'com.android.library'
     id 'org.jetbrains.kotlin.android'
     id 'org.jetbrains.kotlin.plugin.serialization'
-    id("org.jetbrains.kotlinx.kover")
+    id('org.jetbrains.kotlinx.kover')
 }
 
 android {
-    compileSdk 34
+    compileSdk 32
     namespace "com.swmansion.starknet"
 
     defaultConfig {
         minSdk 24
-        targetSdk 34
+        targetSdk 32
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles "consumer-rules.pro"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,6 +2,7 @@ plugins {
     id 'com.android.library'
     id 'org.jetbrains.kotlin.android'
     id 'org.jetbrains.kotlin.plugin.serialization'
+    id("org.jetbrains.kotlinx.kover")
 }
 
 android {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -5,12 +5,12 @@ plugins {
 }
 
 android {
-    compileSdk 32
+    compileSdk 34
     namespace "com.swmansion.starknet"
 
     defaultConfig {
         minSdk 24
-        targetSdk 32
+        targetSdk 34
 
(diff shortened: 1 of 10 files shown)


android/build.gradle
```diff
@@ -55,9 +55,9 @@ android {
...
 dependencies {
     testImplementation 'junit:junit:4.13.2'
-    androidTestImplementation 'androidx.test.ext:junit:1.1.3'
-    androidTestImplementation 'androidx.test.espresso:espresso-core:3.4.0'
-    implementation 'org.bouncycastle:bcprov-jdk15on:1.70'
-    implementation("com.squareup.okhttp3:okhttp:4.10.0")
-    implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.3.3")
+    androidTestImplementation 'androidx.test.ext:junit:1.1.5'
+    androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
+    implementation 'org.bouncycastle:bcprov-jdk15to18:1.75'
```
DelevoXDG, line 60: It can be either [bcprov-jdk18on](https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk18on) or [bcprov-jdk15to18](https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk15to18) (currently selected)

android/build.gradle
```diff
@@ -2,15 +2,16 @@ plugins {
...
 
 android {
-    compileSdk 32
+    compileSdk 34
     namespace "com.swmansion.starknet"
 
     defaultConfig {
         minSdk 24
-        targetSdk 32
+        targetSdk 34
```
THenry14, line 14: let's not bump sdk version, it should stay at 32

android/build.gradle
```diff
@@ -2,15 +2,16 @@ plugins {
     id 'com.android.library'
     id 'org.jetbrains.kotlin.android'
     id 'org.jetbrains.kotlin.plugin.serialization'
+    id("org.jetbrains.kotlinx.kover")
```
THenry14, line 5:
```suggestion
    id('org.jetbrains.kotlinx.kover')
```

THenry14: btw, why add `kover` ?

THenry14 (changes requested):

android/build.gradle
DelevoXDG: It's not required since we do not currently measure coverage for android as a part of checks, but added for consistency since in newer version of kover [it should be applied](https://kotlin.github.io/kotlinx-kover/gradle-plugin/) to each submodule for which we might want a coverage report.

THenry14: But we didn't have this for lib as well, right? Maybe I misunderstand something, but I don't think we should add this. We are running codecov on gh actions, I think this should be enough

DelevoXDG: I applied it for lib as well, the gh actions fail otherwise. Newer versions of kover won't generate a report to upload to codecov if kover is not explicitly applied to submodule.

THenry14 (approved):

PR #273, opened 2023-07-19
Add caching on CI for Python and Rust deps

### LINKED ISSUE #256: Cache cairo deps on CI
### Feature Request

It shouldn;t be compiled each run

DelevoXDG:
## Describe your changes
- [x] Locally cache compiled contracts.
- [x] Cache python deps on gh actions.
- [x] Cache cairo rust deps on gh actions. 
- [ ] Cache contracts on gh actions. 
- [ ] Cache poseidon and petersen on gh actions.

## Linked issues

Closes #256 

## Breaking changes

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -27,12 +27,14 @@ jobs:
         uses: jwlawson/actions-setup-cmake@4f73d30c2fc44a9466a3ed890fb8db7a7b554302
         with:
           cmake-version: '3.16.x'
-      - name: Set up Python 3.9.12 and devnet
+      - name: Set up Python 3.9.12
         uses: actions/setup-python@v4.5.0
         with:
           python-version: 3.9.12
           cache: 'pip'
-      - run: |
+
+      - name: Install devnet
+        run: |
           python -m pip install --upgrade pip
           pip install -r requirements.txt
 
--- a/lib/src/test/resources/compileContracts.sh
+++ b/lib/src/test/resources/compileContracts.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 pushd "$(dirname "$0")" || exit
-
 mkdir -p "compiled_v0"
 
 echo "Compiling v0 contracts.."


.github/workflows/checks.yml
```diff
@@ -24,22 +27,31 @@ jobs:
         uses: jwlawson/actions-setup-cmake@4f73d30c2fc44a9466a3ed890fb8db7a7b554302
         with:
           cmake-version: '3.16.x'
-      - name: Set up Python 3.9.12
-        uses: actions/setup-python@v4
+      - name: Set up Python 3.9.12 and devnet
```
THenry14, line 30: This action only sets up python, it does not install devnet

.github/workflows/checks.yml
```diff
@@ -24,22 +27,31 @@ jobs:
...
+        uses: actions/setup-python@v4.5.0
         with:
           python-version: 3.9.12
-
-      - name: Install devnet
```
THenry14, line 32: Why remove step name? :(

.github/workflows/checks.yml
```diff
@@ -24,22 +27,31 @@ jobs:
...
       - name: Lint
         run: ./gradlew :lib:lintKotlin
 
+      - name: Cache cairo (rust) compiled deps
+        uses: actions/cache@v3
+        with:
+          path: cairo/target/debug
+          key: ${{ runner.os }}-cairo-target-debug-${{ hashFiles('cairo/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cairo-target-debug-
```
THenry14, line 53: probably missing Cargo.lock hash, right?

lib/build.gradle.kts
```diff
@@ -70,7 +70,22 @@ val buildCrypto = task<Exec>("BuildCrypto") {
...
+    val compileDirectoryV1 = layout.buildDirectory.dir("lib/src/test/resources/compiled_v1")
+
+    inputs.file(compileScriptPath)
+    inputs.dir(sourceDirectoryV0)
+    inputs.dir(sourceDirectoryV1)
+
+    outputs.dir(compileDirectoryV0)
+    outputs.dir(compileDirectoryV1)
+
+    commandLine(compileScriptPath)
```
THenry14, line 88: what does this do?

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 
 pushd "$(dirname "$0")" || exit
+
+OUT_DIR_V0="$(dirname "$0")/compiled_v0"
+OUT_DIR_V1="$(dirname "$0")/compiled_v1"
```
THenry14, line 6: what is the purpose of this?

.github/workflows/checks.yml
DelevoXDG: Because to run a setup-python job and install deps with caching those need to be in the same sub-task.

.github/workflows/checks.yml
DelevoXDG: It's meant actually as a fallback to at least partially restore dependencies. I guess we could remove it entirely since we do not check whether the rust deps have changed, only that they are present in `compileContracts`.

lib/build.gradle.kts
DelevoXDG: Sets input and output for compileContracts task. This allows to locally cache compiled contracts. If there are no changes to any files in input dirs, script and output dirs, the task will not recompile contracts.

.github/workflows/checks.yml
THenry14: Perhaps I don't understand something, but the way your change works is that you just removed step name, and did not combine two steps into one. So `Setup python 3.9.12 and devnet` is a single step that just sets up python, and `-run` is a separate, nameless step that installs devnet.

See -> https://github.com/software-mansion/starknet-jvm/actions/runs/5670681846/job/15389409119

I'd rather we leave the second step name as it was. Also, not sure if installing devnet takes this much time that it has to be cached, but if you want to proceed with this that's fine.

lib/build.gradle.kts
THenry14: But we already have something like this in compileContracts.sh right? If contracta are built, the scripts skips this step.

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 pushd "$(dirname "$0")" || exit
+
```
THenry14, line 4: 👀

lib/build.gradle.kts
DelevoXDG: No, we don't. We check whether compilers are built, but we recompile contracts each time.

DelevoXDG: I'm not sure it uses checksums per se, but it is based on both metadata and file contents and works as expected: https://docs.gradle.org/current/userguide/incremental_build.html#sec:task_inputs_outputs

PR #281, opened 2023-07-25
Support RPC 0.4.0

### LINKED ISSUE #262: Support starknet 0.12
### Feature Request

Rpc version 0.4.0-rc0 has just been released, we should work towards supporting it (it hits integration in ~2 weeks)
https://github.com/starkware-libs/starknet-specs/tree/v0.4.0-rc0

Let's also check if there are other (non rpc-specific things) that's changed in this release and support them.

### LINKED ISSUE #270: Update estimate message fee to match RPC 0.4.0 schema changes
### Feature Request

Update estimate message fee to match RPC 0.4.0 schema changes.
Currently cannot be fully tested since it's not yet implemented on devnet.
Blocked until then.

### LINKED ISSUE #280: Support RPC 0.4.0
### Feature Request

https://github.com/starkware-libs/starknet-specs/tree/v0.4.0-rc2

DelevoXDG:
## Describe your changes

- Updated `Message` dataclasses, `get_estimate_message_fee`
- Added `execution_status`, `finality_status`, `revert_reason` to `TransactionReceipt`
- Updated `Transaction` dataclasses, added `DeclareTransactionV0` dataclass
- ~~Updated `BlockWithTx`~~
- Bumped `devnet` to `0.6.0`
- Added mock tests: adjust to pythonic devnet not having RPC 0.4.0 support
- Added integration tests

## Linked issues

Closes #280
Closes #270
Closes #262 

## Breaking changes

- Changes to `TransactionReceipt` dataclasses
- Changes to `Message` dataclasses, `get_estimate_message_fee`


DelevoXDG:
We should review the current changes, but block this until devnet update arrives or at least there's some clarity regarding what changes will be included in the release so it is reasonable to write mock tests.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionReceiptPolymorphicSerializer.kt
```diff
@@ -9,7 +9,7 @@ internal object JsonRpcTransactionReceiptPolymorphicSerializer :
     JsonContentPolymorphicSerializer<TransactionReceipt>(TransactionReceipt::class) {
     override fun selectDeserializer(element: JsonElement): DeserializationStrategy<out TransactionReceipt> {
         val jsonElement = element.jsonObject
-        val isPending = "status" !in jsonElement
+        val isPending = "status" !in jsonElement && "execution_status" !in jsonElement
```
THenry14, line 12: is this a new property in rpc? what does it indicate?

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionPolymorphicSerializer.kt
```diff
@@ -25,10 +25,15 @@ internal object TransactionPolymorphicSerializer : JsonContentPolymorphicSeriali
...
+    private fun selectDeclareDeserializer(element: JsonElement): DeserializationStrategy<out DeclareTransaction> {
+        if ((element.jsonObject["contract_class"] == null) && (element.jsonObject["class_hash"] == null)) {
+            throw IllegalArgumentException("Either classHash or contractDefinition must be present.")
+        }
+
         when (element.jsonObject["version"]?.jsonPrimitive?.content) {
-            Felt.ONE.hexString() -> DeclareTransactionV1.serializer()
-            Felt(2).hexString() -> DeclareTransactionV2.serializer()
+            Felt.ONE.hexString() -> return DeclareTransactionV1.serializer()
+            Felt(2).hexString() -> return DeclareTransactionV2.serializer()
```
THenry14, line 35: why this change?

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionPolymorphicSerializer.kt
```diff
@@ -25,10 +25,15 @@ internal object TransactionPolymorphicSerializer : JsonContentPolymorphicSeriali
             else -> throw IllegalArgumentException("Invalid invoke transaction version")
         }
 
-    private fun selectDeclareDeserializer(element: JsonElement): DeserializationStrategy<out DeclareTransaction> =
+    private fun selectDeclareDeserializer(element: JsonElement): DeserializationStrategy<out DeclareTransaction> {
+        if ((element.jsonObject["contract_class"] == null) && (element.jsonObject["class_hash"] == null)) {
+            throw IllegalArgumentException("Either classHash or contractDefinition must be present.")
+        }
```
THenry14, line 31: what if both are set?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Messages.kt
```diff
@@ -20,6 +20,9 @@ data class GatewayMessageL1ToL2(
     @JsonNames("from_address")
     val fromAddress: Felt,
 
+    @JsonNames("to_address")
+    val toAddress: Felt,
+
```
THenry14, line 25: are you sure this was added in gateway?

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -41,11 +42,35 @@ sealed class Transaction {
...
+data class TransactionHashPair(
+    @Serializable(with = TransactionPolymorphicSerializer::class)
+    @SerialName("transaction")
+    @JsonNames("txn")
+    val transaction: Transaction,
+
+    @SerialName("transaction_hash")
+    @JsonNames("txn_hash")
+    val hash: Felt,
+)
```
THenry14, line 56: Not sure this is a way to go. I'd rather we didn't merge those two into one data class - perhaps there could be some other data class that is super class to them?

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -41,11 +42,35 @@ sealed class Transaction {
...
+
 @OptIn(ExperimentalSerializationApi::class)
 @Serializable
 @SerialName("DEPLOY")
 // OptIn needed because @JsonNames is part of the experimental serialization api
 data class DeployTransaction(
+    // not in RPC spec
+    @SerialName("transaction_hash")
+    @JsonNames("txn_hash")
+    override val hash: Felt = Felt.ZERO,
```
THenry14, line 66: I don't think this is right, and might lead to all kinds of problems

THenry14: \+ please do not change the order of properties, it is really hard ro review other changes :(

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -221,35 +249,41 @@ data class DeclareTransactionV1(
...
-    @SerialName("transaction_hash")
-    @JsonNames("txn_hash")
-    override val hash: Felt,
-
     @SerialName("max_fee")
     override val maxFee: Felt,
 
     @SerialName("version")
-    override val version: Felt = Felt(2),
+    override val version: Felt = Felt.ONE,
```
THenry14, line 267: this is `DeclareTransactionV2`, please do not change the version :P

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -201,55 +201,55 @@ class StandardAccountTest {
...
-    }
+    // @Test
+    // fun `mock estimate message fee`() {
+    //     // Note for future developers experiencing failures in this test:
+    //     // This test is designed with RPC 0.3.1 in mind (which was never released).
+    //     // The schema will be changed as of RPC 0.4.0 and provider.getEstimateMessageFee will have a different payload.
+
+    //     val gasConsumed = Felt(45100)
+    //     val gasPrice = Felt(2)
+    //     val overallFee = Felt(45100 * 2)
```
war-in, line 212: 👀 I suppose this is temporary, right?

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionReceiptPolymorphicSerializer.kt
war-in: Should we have isPending variable if there are no pending transactions?

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
war-in: And I would definitely change the name if we leave it as it is

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
war-in: Yup, agree. Please return to the old order 🙏

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -442,6 +501,9 @@ object TransactionFactory {
     fun makeDeclareV2Transaction(
         senderAddress: Felt,
         contractDefinition: Cairo1ContractDefinition,
+//            contractDefinition: Cairo1ContractDefinition? = null,
+        // TODO: Not sure we want @contractDefition to be nullable.
+        //       It wasn't before as makeDeclareV2Transaction was used only for actually declaring contracts, not creating the object for already executed declare transaction.
```
war-in, line 506: Does makeDeclareV2Transaction is used when fetching state?

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
DelevoXDG: This was removed in 0.4-rc4 and replaced with allO, will go back to using Transaction class.

lib/src/test/kotlin/starknet/utils/MockUtils.kt
```diff
@@ -0,0 +1,54 @@
+package starknet.utils
```
THenry14, line 1: Wouldn't it be easier, and more explixit, to include mocked responses in places they are needed? I suppose this appraoch might work, but what if something breaks? I don't really like that some code is modifying real responses, because I don't know if the things it makes are still necessary. With fixed mocked responses if something changes I am 100% sure that nothing is being hidden/mnodified on the fly.

Not sure, would like to hear what others think.

.github/workflows/checks.yml
```diff
@@ -15,6 +15,12 @@ jobs:
         ]
 
     runs-on: ${{ matrix.os }}
+    env:
+      STARKNET_JVM_INTEGRATION_TEST_SOURCE: "env"
+      STARKNET_JVM_SKIP_INTEGRATION_GETTERS_TEST: "false"
```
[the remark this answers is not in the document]
DelevoXDG: It is.
- https://github.com/software-mansion/starknet-jvm/blob/09d20093dc7be422db0abe891c6e64c83314e7b1/lib/src/test/kotlin/integration/utils/ConfigUtils.kt#L69C1-L70C1
Used to determine whether to get secrets from file or env.
- https://github.com/software-mansion/starknet-jvm/blob/09d20093dc7be422db0abe891c6e64c83314e7b1/lib/src/test/kotlin/integration/getters/GettersTest.kt#L47C10-L47C10
Checking whether we want to run a test on CI. This is useful because node could be down, not updated, or we do not want run some gas-heavy tests on CI.

.github/workflows/checks.yml
```diff
@@ -15,6 +15,12 @@ jobs:
...
+      STARKNET_JVM_RPC_URL: ${{ secrets.STARKNET_JVM_RPC_URL }}
+      STARKNET_JVM_INTEGRATION_ACCOUNT_ADDRESS: ${{ secrets.STARKNET_JVM_INTEGRATION_ACCOUNT_ADDRESS }}
+      STARKNET_JVM_INTEGRATION_PRIVATE_KEY: ${{ secrets.STARKNET_JVM_INTEGRATION_PRIVATE_KEY }}
```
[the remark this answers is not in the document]
DelevoXDG: > I think we can drop the STARKNET_JVM prefix.

We could, but if those are to be used locally, it would be confusing. What if at some point similar approach is used in starknet.py or starknet.swift?

> btw I was under the impression, that we'll be only running getter tests on CI. If so, why we need an account and private key?

Those tests can be used both locally and on CI. We will need an account if we want to make any tests that require gas, this is the account used in starknet.py that already has gas.

.gitignore
```diff
@@ -48,6 +48,7 @@ venv
 # Exclude account temp files
 lib/src/test/resources/*account
 lib/src/test/resources/compiled*
+lib/src/test/kotlin/integration/config.json
```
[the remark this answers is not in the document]
DelevoXDG: Since it is only used for tests in `lib` and root dir is quite cluttered, maybe move it there?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -135,6 +135,8 @@ data class BlockWithTransactionsResponse(
     @SerialName("status")
     val status: BlockStatus,
 
+    // Block header
+
     @SerialName("parent_hash")
     override val parentHash: Felt,
```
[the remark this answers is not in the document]
DelevoXDG: I just added this comment because in the RPC schema it is annotated that way. I think having comments about the schema structure instead of guessing where the flattened structure came from would simplify the process if there's ever a need to update any of it.
```json      
      "BLOCK_WITH_TX_HASHES": {
                "title": "Block with transaction hashes",
                "description": "The block object",
                "allOf": [
                    {
                        "title": "Block status",
                        "type": "object",
                        "properties": {
                            "status": {
                                "title": "Status",
                                "$ref": "#/components/schemas/BLOCK_STATUS"
                            }
                        },
                        "required": [
                            "status"
                        ]
                    },
                    {
                        "title": "Block header",
                        "$ref": "#/components/schemas/BLOCK_HEADER"
                    },
                    {
                        "title": "Block body with transaction hashes",
                        "$ref": "#/components/schemas/BLOCK_BODY_WITH_TX_HASHES"
                    }
                ]
            },
```

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -200,11 +206,17 @@ data class DeclareTransactionV1(
     @SerialName("type")
     override val type: TransactionType = TransactionType.DECLARE,
 
+//    @SerialName("contract_class")
     private val contractDefinition: Cairo0ContractDefinition? = null,
+
+//        @SerialName("contract_class")
+//        val contractClass: DeprecatedContractClass? = null,
+
```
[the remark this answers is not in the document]
DelevoXDG: This will be removed but has no impact due to the fact how `DeclareTransaction`s are serialized (manually)

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -221,15 +233,18 @@ data class DeclareTransactionV1(
 @Serializable
 @SerialName("DECLARE")
 data class DeclareTransactionV2(
+    // TODO: Make this Felt.ZERO instead?
+    //       There shouldn't be any situation where we create this dataclasses with this value set to null
```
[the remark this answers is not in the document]
DelevoXDG: Probably class hashes, but the approach to handle those fields has been decided since the comment.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -30,16 +30,60 @@ enum class TransactionStatus {
...
     UNKNOWN,
+
+    @JsonNames("REVERTED")
+    REVERTED,
+}
+
+@OptIn(ExperimentalSerializationApi::class)
+@Serializable
+enum class TransactionExecutionStatus {
+    // Remove normal case jsonNames when pathfinder is updated
```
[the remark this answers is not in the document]
DelevoXDG: I assume we remove the comment then.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -30,16 +30,60 @@ enum class TransactionStatus {
...
+
+    @JsonNames("SUCCEEDED", "Succeeded")
+    SUCCEEDED,
+
+    @JsonNames("REVERTED", "Reverted")
+    REVERTED,
+
+    @JsonNames("REJECTED", "Rejected")
+    REJECTED,
```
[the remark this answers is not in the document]
DelevoXDG: https://community.starknet.io/t/efficient-utilization-of-sequencer-capacity-in-starknet-v0-12-1/95607
Apparently this is only supported by gateway, since it's not in spec.

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
```diff
@@ -16,8 +16,9 @@ class HttpRequest<T>(
 ) : Request<T> {
 
     override fun send(): T {
+        println("IN: ${payload.body}")
         val response = service.send(payload)
-
+        println("OUT: ${response.body}")
```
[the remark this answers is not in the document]
DelevoXDG: Been trying to find a bug that only occurred on CI. 
It's curious, but there are often different error messages on CI than locally. Will remove now.

lib/src/test/kotlin/integration/config-template.json
```diff
@@ -0,0 +1,7 @@
+{
+  "rpc_url": "http://example-node-url.com/rpc",
+  "gateway_url": "http://example.com/gateway",
+  "feeder_gateway_url": "http://example.com/feeder_gateway",
+  "private_key": "0x123456789",
+  "integration_account_address": "0x123456789123456789123456789123456789123456789123456789123456789"
+}
```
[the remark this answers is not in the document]
DelevoXDG: Let's see what other think, for me it is more intuitive to have it as a file and replace name and values, but maybe I'm the only one.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -30,16 +30,60 @@ enum class TransactionStatus {
...
+    ACCEPTED_ON_L1,
+
+    @JsonNames("ACCEPTED_ON_L2", "AcceptedOnL2")
+    ACCEPTED_ON_L2,
+
+    @JsonNames("RECEIVED", "Received")
+    RECEIVED,
+
+    @JsonNames("NOT_RECEIVED", "NotReceived")
+    NOT_RECEIVED,
```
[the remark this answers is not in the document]
DelevoXDG: https://community.starknet.io/t/efficient-utilization-of-sequencer-capacity-in-starknet-v0-12-1/95607
Those are only supported by gateway, `NOT_RECEIVED` was taken from starknet.py. 

Although having it certainly won't hurt, maybe @tomek0123456789 could comment whether it's actually used anywhere.

lib/src/test/kotlin/integration/config-template.json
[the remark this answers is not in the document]
THenry14: I think this is better approach: https://github.com/software-mansion/starknet-jvm/pull/281#discussion_r1290173031

that way, you still have a file, I'm happy because it is not just some random json, and we get to make things a lot simpler :+1:

.github/workflows/checks.yml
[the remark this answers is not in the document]
THenry14: 1st link links to `STARKNET_JVM_INTEGRATION_TEST_SOURCE`
2nd link links to `assumeFalse(ConfigUtils.isTestSetSkipped(testName), "$testName is skipped in the current environment.")`

I assume you wanted to link this https://github.com/software-mansion/starknet-jvm/blob/09d20093dc7be422db0abe891c6e64c83314e7b1/lib/src/test/kotlin/integration/utils/ConfigUtils.kt#L38

I think you can make it a lot simpler and just do "CI_SKIP_TEST", and that's it. no need to get `$screamingSnakeName` there

.github/workflows/checks.yml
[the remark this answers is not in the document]
THenry14: > We could, but if those are to be used locally, it would be confusing. What if at some point similar approach is used in starknet.py or starknet.swift?

Very slim chances for that. And even if, I don't think running those with the same account is a bad thing.

> Those tests can be used both locally and on CI. We will need an account if we want to make any tests that require gas, this is the account used in starknet.py that already has gas.

Please, do not run anything on CI that needs funds. This is bad. What if the funds dry out? We have failing tests and a headache. CI should only do things, that does not require an account. Locally, it's fine.

.gitignore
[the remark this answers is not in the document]
THenry14: This will only be locally, so one more file in root won't clutter things that much imo. I'd still put this in root.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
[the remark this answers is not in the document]
THenry14: I meant, since spec has a field `block_header` and not `parent_hash`, shouldn't we rename `parent_hash` to `block_header` instead having a comment?

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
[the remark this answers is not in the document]
THenry14: Sure, but a ready PR should not have this. Ready PR should be free of commented code blocks, printlns etc :+1:

lib/src/test/kotlin/integration/utils/ConfigUtils.kt
```diff
@@ -0,0 +1,79 @@
...
+            )
+        }
+        private fun makeConfigFromFile(path: String): Config {
+            val configPath = Paths.get(path)
+            if (!Files.exists(configPath)) {
+                throw IllegalStateException("Config file not found")
+            }
+
+            return json.decodeFromString(Config.serializer(), configPath.readText())
+        }
```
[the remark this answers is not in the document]
DelevoXDG: It certainly would make it cleaner. But it depends on what do you mean by sourcing the file:
We could use something like https://github.com/cdimascio/dotenv-kotlin (there's no built-in implementation I'm aware of) to load the variables from .env file, but since you have different syntax from System.getenv(...) this would pretty much mirror the json approach.

If you mean sourcing in a regular way, through a terminal, it should be noted that some IDEs will require a restart to include newly sourced vars or supplying each one by one to every run configuration, which would be suboptimal speed and convenience wise.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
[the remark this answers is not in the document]
THenry14: You can leave a comment, and annotate it with TODO: #issue number

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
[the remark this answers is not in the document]
THenry14: if it's used in gateway, I guess we have to leave it for now

.github/workflows/checks.yml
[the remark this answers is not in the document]
DelevoXDG: 1st links links to where it's used from what I see?
2nd link:
> I think you can make it a lot simpler and just do "CI_SKIP_TEST", and that's it. no need to get $screamingSnakeName there
This what I did at first, but I assume we are planning to add the integration tests for functions that do change the state and skip those on CI. So I thought we could make it so we have a variable for every test file.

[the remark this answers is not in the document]
THenry14: > 1st links links to where it's used from what I see?

My question was about `STARKNET_JVM_SKIP_INTEGRATION_GETTERS_TEST`, and it's not used where 1st link points to

> This what I did at first, but I assume we are planning to add the integration tests for functions that do change the state and skip those on CI. So I thought we could make it so we have a variable for every test file.

Then you will have as many variables as test files, right? All doing basically the same thing - stopping tests from execution (presumably on CI). It seems to me, that one variable should be enough to achieve exactly this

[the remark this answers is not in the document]
DelevoXDG: Well, I'm intending to run GettersTest on CI and run tests that require gas only locally. Can't achieve that with one variable.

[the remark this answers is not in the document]
THenry14: why?

.github/workflows/checks.yml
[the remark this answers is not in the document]
THenry14: Discussed offline, clarification is:

on CI there will be no tests that require any funds. an account will only be used to sign fee estimation transaction, which does nto require any fee.

.github/workflows/checks.yml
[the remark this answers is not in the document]
DelevoXDG: Ok, do I understand correctly that your idea is that for every test file that is not intended to be ran on CI we will have a check in `@BeforeAll`, and for those which are intended to run wherever we won't .

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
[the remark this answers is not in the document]
DelevoXDG: `block_header` is not a field in a schema, `BLOCK_WITH_TX_HASHES ` just expands fields from `block_header`.

.github/workflows/checks.yml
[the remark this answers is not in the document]
THenry14: Basically - yes. 

Although, I don't think you **_need_** to have a `BeforeAll`, we could, for example, flip the scenario and have a `RUN_INTEGRATION_TEST` variable that, when it's set, triggers those tests (by default - no variable, so tests won't run), and have a check in gradle script, something like this: https://www.baeldung.com/gradle-skip-tests#using-the-gradle-build-script. That would probably make some sense, as those tests won't be triggered very often, due to the cost etc

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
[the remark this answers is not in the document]
THenry14: ah yes, you're right :+1:

.github/workflows/checks.yml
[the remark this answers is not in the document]
DelevoXDG: I just think that modifying this: 
```yml
STARKNET_JVM_SKIP_INTEGRATION_GETTERS_TEST: "false"
STARKNET_JVM_SKIP_INTEGRATION_ACCOUNT_TEST: "true"
STARKNET_JVM_SKIP_INTEGRATION_PROVIDER_TEST: "true"
```
is a bit simplier than modifying lines of code in every test. 

If I want to disable getters test both on integration and locally (because node is down etc.), I don't need to go to that specific file and look where do I have a check and modify any code - I just change a variable.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionReceiptPolymorphicSerializer.kt
DelevoXDG: You are correct, do you think we should check for both or just one?

DelevoXDG: You're right, and there's currently no test that that checks if pending receipt is correctly deserialized, so I'll add one as well.

.github/workflows/checks.yml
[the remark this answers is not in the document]
THenry14: I think you misunderstood what I wrote.

My proposition was to run integration tests more or less like so:

```
./gradlew lib:test -Pintegration
```

in this scenario, 

```
./gradlew lib:test
```

does not run integration tests.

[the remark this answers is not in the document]
DelevoXDG: > My proposition was to run integration tests more or less like so:
> `./gradlew lib:test -Pintegration`
> in this scenario,
> `./gradlew lib:test`
> does not run integration tests.

Ok, but I'm not sure how would that be better. It just seems to me that it is more convenient to have everything checked after running `./gradlew lib:test`, unless something is disabled explicitly. Specifically GettersTest are fast to run, so don't know why not include them in lib:test by default.

> edit: OR using env variable (this was not covered in the link I provided, but still I think may be a good idea) - just checking the variable, and if it not set (or false) - don't run integration tests; othwerwise do. No need to modify any code.
Sorry, I cannot understand how can you achieve specifying whether test should be skipped with just one variable without having different code for every test. 😅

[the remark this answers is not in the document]
THenry14: > Ok, but I'm not sure how would that be better. It just seems to me that it is more convenient to have everything checked after running ./gradlew lib:test, unless something is disabled explicitly. Specifically GettersTest are fast to run, so don't know why not include them in lib:test by default.

but this is very similar approach, the only thing is that you don't have an env variable, but you pass an argument directly to the test command. 

I know they are fast to run, but they are also prone to errors - what if rpc node is down? what if it has wrong version? what if someone does not have a rpc node, but wants to contribute with some simple PR? I jsut don't think we should run such tests by default, only on demand.

> Sorry, I cannot understand how can you achieve specifying whether test should be skipped with just one variable without having different code for every test. 😅

depends on approach, let's say for the sake of an argument that the approach does not change - everything is the same as in your PR. Then you just replace `STARKNET_JVM_SKIP_INTEGRATION_GETTERS_TEST`, `STARKNET_JVM_SKIP_INTEGRATION_ACCOUNT_TEST`, `STARKNET_JVM_SKIP_INTEGRATION_PROVIDER_TEST` variable checking in your code to one variable `SKIP_CI_TEST`, and that's it.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -30,22 +30,70 @@ enum class TransactionStatus {
...
     UNKNOWN,
+
+    @JsonNames("REVERTED")
+    REVERTED,
+}
+
+@OptIn(ExperimentalSerializationApi::class)
+@Serializable
+enum class TransactionExecutionStatus {
+    // TODO: issue #294
```
tomek0123456789, line 41: As mentioned above

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -30,22 +30,70 @@ enum class TransactionStatus {
...
+    REVERTED,
+
+    @JsonNames("REJECTED", "Rejected")
+    REJECTED,
+}
+
+@OptIn(ExperimentalSerializationApi::class)
+@Serializable
+enum class TransactionFinalityStatus {
+    // TODO: issue #294
```
tomek0123456789, line 57: As mentioned above

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -83,7 +131,16 @@ data class GatewayTransactionReceipt(
...
-    override val status: TransactionStatus = TransactionStatus.UNKNOWN,
+    val status: TransactionStatus = TransactionStatus.UNKNOWN,
+
+    @JsonNames("execution_status")
+    override val executionStatus: TransactionExecutionStatus,
+
+    @JsonNames("finality_status")
+    override val finalityStatus: TransactionFinalityStatus,
+
+    @JsonNames("revert_error", "revert_reason")
```
tomek0123456789, line 142: Does gateway still return `revert_reason`?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Messages.kt
```diff
@@ -1,15 +1,18 @@
 package com.swmansion.starknet.data.types
 
 import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.JsonNames
 
 @OptIn(ExperimentalSerializationApi::class)
 @Serializable
 data class GatewayMessageL2ToL1(
```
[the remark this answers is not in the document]
DelevoXDG: Merged gateway and rpc Message dataclasses, added tests, should now be the same as in starknet.py.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -147,35 +149,47 @@ data class BlockWithTransactionsResponse(
...
-    @SerialName("sequencer_address")
-    override val sequencerAddress: Felt,
 ) : GetBlockWithTransactionsResponse()
 
 @Serializable
 data class PendingBlockWithTransactionsResponse(
-    @SerialName("parent_hash")
-    override val parentHash: Felt,
+    // Not in RPC schema
+    // TODO: remove this when Pathfinder is updated
```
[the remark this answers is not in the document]
DelevoXDG: No, I mean the status field. It shouldn't be there, at least it's not in the schema:
```
            "PENDING_BLOCK_WITH_TXS": {
                "title": "Pending block with transactions",
                "description": "The dynamic block being constructed by the sequencer. Note that this object will be deprecated upon decentralization.",
                "allOf": [
                    {
                        "title": "Block body with transactions",
                        "$ref": "#/components/schemas/BLOCK_BODY_WITH_TXS"
                    },
                    {
                        "type": "object",
                        "title": "Block Info",
                        "properties": {
                            "timestamp": {
                                "title": "Timestamp",
                                "description": "The time in which the block was created, encoded in Unix time",
                                "type": "integer",
                                "minimum": 0
                            },
                            "sequencer_address": {
                                "title": "Sequencer address",
                                "description": "The StarkNet identity of the sequencer submitting this block",
                                "$ref": "#/components/schemas/FELT"
                            },
                            "parent_hash": {
                                "title": "Parent hash",
                                "description": "The hash of this block's parent",
                                "$ref": "#/components/schemas/BLOCK_HASH"
                            }
                        }
                    }
                ]
            }
```

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
DelevoXDG: Actually, both Gateway and JsonRpc providers seem to return `revert_error`. While in the 0.4.0 spec, there's no mention of `revert_error` and only `revert_error` is used.

For now I think we can't go wrong with accepting both.

.github/workflows/checks.yml
```diff
@@ -15,6 +15,12 @@ jobs:
         ]
 
     runs-on: ${{ matrix.os }}
+    env:
+      ENABLE_INTEGRATION_TESTS: "true"
+      ENABLE_GAS_TESTS: "false"
```
THenry14, line 20: I was under the impression that:

1. we'll be not enabling integration tests at all on CI for now (as per offline discussion - there were problems with timeouts)
2. there was a discussion in here regarding enabling/disabling tests using _**single**_ env variable OR a flag to lib:test, and I stand by this idea. I don't think we will ever enable gas tests on CI, or any tests requiring fee for that matter. All the integration tests should be runnable easily locally using (ideally) one command, and that's it.

integration_tests_config_template.env
```diff
@@ -0,0 +1,7 @@
+INTEGRATION_RPC_URL=http://example-node-url.com/rpc
```
THenry14, line 1: let's rename the file to something like: `integration_tests.env.example`, and add a general info in README about configuring and running those tests

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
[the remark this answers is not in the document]
THenry14: let's improve the comment, so it is clear what is to be removed then. Ideally, there should be an issue number stated here, under which more extensive description shall be written.

lib/src/main/kotlin/com/swmansion/starknet/signer/StarkCurveSigner.kt
```diff
@@ -16,7 +16,11 @@ class StarkCurveSigner(private val privateKey: Felt) : Signer {
     override val publicKey: Felt by lazy { StarknetCurve.getPublicKey(privateKey) }
 
     override fun signTransaction(transaction: Transaction): Signature {
-        return StarknetCurve.sign(privateKey, transaction.hash).toList()
+        if (transaction.hash == null) {
+            throw IllegalArgumentException("Transaction hash must be set before signing")
```
THenry14, line 20: Perhaps we could improve the error message here? This implies that this is something user should do, and this is probably not a case, right?

lib/src/test/kotlin/integration/provider/ProviderTest.kt
```diff
@@ -0,0 +1,352 @@
+package integration.provider
```
THenry14, line 1: these are tests not requiring gas, right? so not exactly provider tests per se. I'd rename filename to something more meaningful, like `TestsNotRequiringGas.kt`. Maybe we should rename other integration test files as well.

lib/src/test/kotlin/integration/account/AccountTest.kt
```diff
@@ -0,0 +1,262 @@
...
+        assertNotNull(receipt)
+        assertTrue(receipt.isAccepted)
+    }
+
+    @Disabled
+    // TODO: Fails with "Compiled versions older than 1.1.0 or newer than 1.3.0 are not supported. Got version 1.0.0."
+    // Can be reenabled after compiler version bump (awaiting PR #285 merge)
+    @ParameterizedTest
+    @MethodSource("getAccounts")
+    fun `sign and send declare v2 transaction`(accountAndProvider: AccountAndProvider) {
```
THenry14, line 235: Let's add more tests - ideally we should test the whole flow of declaring, deploying and invoking/calling contracts.

THenry14: Of course this does not have to be done in this PR, can be added later.
Said flow should, however, be manually tested by you at least once before merging this PR.

lib/build.gradle.kts
```diff
@@ -101,6 +101,18 @@ tasks.test {
     val poseidonPath = file("${rootDir}/crypto/poseidon/build/bindings").absolutePath
 
     systemProperty("java.library.path", "$libsSharedPath:$pedersenPath:$poseidonPath")
+    systemProperty(
+            "enableIntegrationTests",
+            project.findProperty("enableIntegrationTests")
+                    ?: System.getenv("ENABLE_INTEGRATION_TESTS")
+                    ?: "true",
```
THenry14, line 108: I think integration tests should be disabled by default. We should not require any users to set up rpc nodes just to make sure tests passes - by default we should test only with devnet, and make integration tests purely optional.

THenry14 (commented):
Great job mate 🔥 only a few smaller comments

.github/workflows/checks.yml
DelevoXDG: 1. Correct, but the reason was temporary outages of integration network which appear to be resolved as of now.

2. There was, but I was under the impression the idea was to get rid of 1+n variables. 
I believe we need at least those 3 states, which cannot be achieved with just one boolean:
1) disabling all integration tests 
2) enabling only non-gas tests
3) enabling both gas and non-gas tests

Keep in mind that AccountTest contains both tests that do require gas and those which do not (with the idea that eventually AccountTest, ProviderTest, etc. will mostly mimic the test suites for devnet).
So just passing `./gradlew :lib:test -x AccountTest` won't achieve the desired functionality.

Now, the idea is that those variables have the default values (which satisfies state 2).
Locally, expected commands are either of those:
`./gradlew :lib:test -PenableIntegrationTests=false` (1)
`./gradlew :lib:test -PenableGasTests=true` (3)

So we are still realistically passing one flag. If there's no flag, it's looking up those values in env. I think it would be nice fallback option for CI, where passing extra flags to `./gradlew :lib:koverXmlReport --info` would be less clear since the `:lib:test` is called indirectly.

We can of course tweak this to have both vars set to false by default, and we can make `PenableGasTest=true` override `PenableIntegrationTests=false` so we can be sure one flag is enough. If you insist on having one variable including defaults, we can consider something like:
`INTEGRATION_TEST_MODE` with possible values [`disabled`, `non_gas`, `all`] and explain it in readme. 

But I don't think we should complicate enabling any of those states, as even in matter of short time I found myself switching between those quite often.

lib/src/test/kotlin/integration/provider/ProviderTest.kt
DelevoXDG: As I mentioned [here](https://github.com/software-mansion/starknet-jvm/pull/281#discussion_r1296904503), the idea is that AccountTest, ProviderTest, etc. will eventually mostly mimic the test suites for devnet.
 
Whether the test case requires gas will be specified in the beginning of said test case.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -158,8 +161,10 @@ data class InvokeTransactionV0(
     @SerialName("signature")
     override val signature: Signature,
 
+    // not in RPC spec
+    // TODO: consider make it nullable so it's not included in serialized json
```
war-in, line 165: What about this TODO?

lib/src/test/kotlin/integration/account/AccountTest.kt
```diff
@@ -0,0 +1,261 @@
...
+        fun after() {}
+    }
+
+    @ParameterizedTest
+    @MethodSource("getAccounts")
+    fun `estimate fee for declare v1 transaction`(accountAndProvider: AccountAndProvider) {
+        assumeTrue(IntegrationConfig.isTestEnabled(requiresGas = false))
+
+        val (account, provider) = accountAndProvider
+        // TODO: find a better account that has a non-changing nonce
```
war-in, line 99: Is it possible to have an account with constant nonce? And if so should we even consider using such an account in tests?

war-in (approved):
Great job 🔥 Let's wait with merging until @THenry14 approves too

lib/src/test/kotlin/integration/account/AccountTest.kt
DelevoXDG: Yeah, it could be done via creating account and not using it, and only to use it for this particular kind of test. This test can fail (invalid nonce or invalid signature) because gateway has access to pending transactions and therefore transactions is sent with new nonce, while the old one was expect. 

Since gateway is to be removed anyway, I think this can temporary left as is until then and test this on gateway locally only if we need to.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
DelevoXDG: Discussed offline, to be removed.

.github/workflows/checks.yml
DelevoXDG: For now disabled integration tests on CI as per offline discussion and the fact that they are unstable today.

lib/src/test/kotlin/integration/account/AccountTest.kt
DelevoXDG: Seems like after enabling access to pending block, our node suffers from the same nonce issue as of now, in other test cases as well 😢

.github/workflows/checks.yml
THenry14: ad 1. Temporary problems are troublesome, because they can easily come back. Even though integration works now, we should not enable it for CI, probably ever.
ad 2. We'll be only running those tests locally, from time to time. There should only be two scenarios: run integration tests or do not run integration tests. 

In order to push this PR forwards we can leave the discussion at this point, but please make sure you make an issue for this and change this behaviour.

lib/src/test/kotlin/integration/provider/ProviderTest.kt
THenry14: I don't necessarily think they should mimic that behaviour - Ideally we would have a simple test suite that tests the whole flow (or at least the most important parts of it, like declaring, deploying, invoking contracts + maybe getting some info from chain to verify transactions). I don't think it is a good idea to basically duplicate Account and Provider test suites for different chain type.

lib/src/test/kotlin/integration/account/AccountTest.kt
THenry14: I understand, that an issue is created for this? 🙏

.github/workflows/checks.yml
DelevoXDG: Discussed offline, tests will stay disabled, one variable with 3 possible values will be added.

.github/workflows/checks.yml
```diff
@@ -15,6 +15,11 @@ jobs:
         ]
 
     runs-on: ${{ matrix.os }}
+    env:
+      INTEGRATION_TEST_MODE: "disabled"
+      INTEGRATION_RPC_URL: ${{ secrets.INTEGRATION_RPC_URL }}
+      INTEGRATION_ACCOUNT_ADDRESS: ${{ secrets.INTEGRATION_ACCOUNT_ADDRESS }}
+      INTEGRATION_PRIVATE_KEY: ${{ secrets.INTEGRATION_PRIVATE_KEY }}
```
THenry14, line 22: since they are disabled by default, I guess we don't need those?

THenry14 (approved):

PR #285, opened 2023-07-31
Support for Cairo 2 contracts

### LINKED ISSUE #284: Cairo 2 full support
### Feature Request

We have full support for cairo 1 contracts (calculating hashes, declaring etc)
While declaring cairo 2 contract is probably possible, we should verify everything works as expected, and try to add all the missing things that are needed for it to be on par with cairo 1.

### LINKED ISSUE #295: Reenable integration declare tests when #285 is merged
### Feature Request

> Compiled versions older than 1.1.0 or newer than 1.3.0 are not supported. Got version 1.0.0.

Reenable integration declare tests when #285 is merged.

DelevoXDG:
## Describe your changes

- Add `Cairo2ContractDefinition` as typealias of `Cairo1ContractDefinition`
- Changed `compileContracts.sh` to download cairo compilers instead of building them.
- Use the old approach for macos x64 (pending macos aarch64 support on GH actions).
- Added devnet test for contracts with new syntax (cairo compiler v2).
- Remove `cairo` submodule

## Linked issues

Closes #284 
Closes #295 

## Breaking changes


DelevoXDG:
Should we add `Cairo2ContractDefiniton` considering that `Cairo1ContractDefintion` class is already compatible with contracts with new syntax? Technically Cairo 2 ≈ Cairo 1 built with compiler 2.x.x.

THenry14:
> Should we add Cairo2ContractDefiniton considering that Cairo1ContractDefintion class is already compatible with contracts with new syntax? Technically Cairo 2 ≈ Cairo 1 with compiler 2.x.x.

In that case lets just maybe add a typealias for it?

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -302,6 +302,30 @@ class StandardAccountTest {
         assertTrue(receipt.isAccepted)
     }
 
+    @ParameterizedTest
+    @MethodSource("getAccounts")
+    fun `sign and send declare v2 transaction (cairo compiler v2)`(accountAndProvider: AccountAndProvider) {
+        val (account, provider) = accountAndProvider
+        val contractCode = Path.of("src/test/resources/compiled_v2/${provider::class.simpleName}_new_syntax_test_contract.json").readText()
```
THenry14, line 309:
```suggestion
        val contractCode = Path.of("src/test/resources/compiled_v2/${provider::class.simpleName}_contract.json").readText()
```

new sytnax is something that may soon be not true ( ͡° ͜ʖ ͡°)

THenry14: the same change for casmCode below

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -302,6 +302,30 @@ class StandardAccountTest {
...
+        val casmCode = Path.of("src/test/resources/compiled_v2/${provider::class.simpleName}_new_syntax_test_contract.casm").readText()
+
+        val contractDefinition = Cairo1ContractDefinition(contractCode)
+        val contractCasmDefinition = CasmContractDefinition(casmCode)
+        val nonce = account.getNonce().send()
+
+        val declareTransactionPayload = account.signDeclare(
```
THenry14, line 316: So let me just ask - I suppose calculating hashes is exactly the same for cairo 2 as it was for cairo 1, right?

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -77,7 +77,7 @@ class DevnetClient(
                 "--seed",
                 seed.toString(),
                 "--sierra-compiler-path",
-                "../cairo/target/debug/starknet-sierra-compile",
+                "src/test/resources/cairo_compilers_v2/cairo/bin/starknet-sierra-compile",
```
THenry14, line 80: this path could be simplified to something like

`src/test/resources/compilers/v2/cairo/bin`

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,47 +1,117 @@
 #!/bin/bash
 
+SRC_V0="src_v0"
+SRC_V1="src_v1"
+SRC_V2="src_v2"
```
THenry14, line 5: `VX_CONTRACT_PATH`, where X is version

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,47 +1,117 @@
...
+
+COMPILE_V0="compiled_v0"
+COMPILE_V1="compiled_v1"
+COMPILE_V2="compiled_v2"
```
THenry14, line 9: `VX_ARTIFACT_PATH`, where X is version

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,47 +1,117 @@
...
+
+COMPILER_BIN_V1="cairo_compilers_v1"
+COMPILER_BIN_V2="cairo_compilers_v2"
```
THenry14, line 12: It is unnecessary in my opinion - you can put all compilers to `compilers` directory, and name subdirectories by compilers versions, so eg:

`(...)/compilers/2.0.0/(...)`
`(...) /compilers/1.1.1/(...)`

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,47 +1,117 @@
...
+    curl -LsSf "$COMPILER_URL" | tar -x -C "$OUT_DIR" || exit 1
+  else
+    curl -LsSf "$COMPILER_URL" | tar -xz -C "$OUT_DIR" || exit 1
+  fi
+
+  COMPILER_PATH=$(pwd)/"$OUT_DIR"/cairo/bin/starknet-compile
+  SIERRA_PATH=$(pwd)/"$OUT_DIR"/cairo/bin/starknet-sierra-compile
+
+  echo "Done!"
+}
```
THenry14, line 73: These two should be combined, and simplified. You only need to check what system someone is running, and based on that just download file with requested version.

DelevoXDG:
> In that case lets just maybe add a typealias for it?
I was thinking the same thing, but it would require to duplicate some data classes and functions that have `Cairo1ContractDefiniton` in order for linters to actually suggest functions that have signature that has `Cairo2ContractDefiniton`. 
What do you think about just renaming it to just `CairoContractDefiniton`, which would be intended for contracts both for compilers v1 and v2.

THenry14:
> What do you think about just renaming it to just CairoContractDefiniton, which would be intended for contracts both for compilers v1 and v2.

Ok seems fine for now. if there will be some changes introduced to cairo 1/2 we will adjust accordingly then

lib/src/test/resources/compileContracts.sh
DelevoXDG: > These two should be combined, and simplified.
As for combining, it can be done although I don't see how removing function makes code better.
> You only need to check what system someone is running, as can be seen by the comment, and based on that just download file with requested version.
I tried that and actually there's a problem with that. Apparently, `macos-latest` on github actions runs macos with x64 arch, so neither of released binaries actually work on CI. It seems there's no full support for arm64 macos as of now: https://github.com/github/roadmap/issues/528 
Not sure what we should do about it, although the issue mentions that it should be in public beta now.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
DelevoXDG: I wasn't able to find any info about changes to hash calculations and the test with declaring 2.0.0 contract passes.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
DelevoXDG: That's the name in cairo repo. But the fact that the name soon won't be true is true as well.

lib/src/test/resources/compileContracts.sh
DelevoXDG: Then would have to update path `src/test/resources/cairo_compilers_v2/cairo/bin/starknet-sierra-compile` in `devnetProcessBuilder` each time, so we could agree on `compilers/v1`, `compilers/v2`?

lib/src/test/resources/compileContracts.sh
THenry14: > As for combining, it can be done although I don't see how removing function would also simplify the code.

in pseudo code:

```
SYSTEM_NAME=$(uname -s)

if [ "$SYSTEM_NAME" == "Linux" ]; then
  wget https://github.com/starkware-libs/cairo/releases/download/v$VERSION/release-x86_64-unknown-linux-musl.tar.gz
elif [ "$SYSTEM_NAME" == "Darwin" ]; then
  wget https://github.com/starkware-libs/cairo/releases/download/v$VERSION/release-aarch64-apple-darwin.tar
else
  echo "Unsupported platform"
  exit 1
fi
```

this should be enough to have the same functionality as in the marked code.

> I tried that and actually there's a problem with that. Apparently, macos-latest on github actions runs macos with x64 arch, so neither of released binaries actually work on CI. It seems there's no full support for arm64 macos as of now: https://github.com/github/roadmap/issues/528
Not sure what we should do about it, although the issue mentions that it should be in public beta now.

Yup, this is a problem. I see two possible workarounds for now, until gh supports m1s - we can either fetch cairo repo and build it when runners are macos (+ it can be cached), or we can fetch it from somewhere else.

lib/src/test/resources/compileContracts.sh
THenry14: yeah it's fine.

.github/workflows/checks.yml
```diff
@@ -19,15 +19,28 @@ jobs:
...
-      - uses: actions/setup-java@v3
+
+#        # TODO: (#310) Remove when migrating to macOS arm64 runner
+#      - name: Clone Cairo repos (macOS only)
+#        if: matrix.os == 'macos-latest'
+#        run: |
+#          git clone https://github.com/starkware-libs/cairo.git cairo1
+#          git clone https://github.com/starkware-libs/cairo.git cairo2
+#          cd cairo1 && git checkout v1.1.1 && cd ..
+#          cd cairo2 && git checkout v2.2.0 && cd ..
```
DelevoXDG, line 30: It was discussed offline, but I since then I got a strong feeling we shouldn't be making things harder for new users who might be running macOS x64 😅

.github/workflows/checks.yml
```diff
@@ -19,15 +19,28 @@ jobs:
...
-      - uses: actions/setup-java@v3
+
+      # TODO: (#310) Remove when migrating to macOS arm64 runner
+      - name: Clone Cairo repos (macOS only)
+        if: matrix.os == 'macos-latest'
+        run: |
+          git clone https://github.com/starkware-libs/cairo.git cairo1
+          git clone https://github.com/starkware-libs/cairo.git cairo2
+          cd cairo1 && git checkout v1.1.1 && cd ..
+          cd cairo2 && git checkout v2.2.0 && cd ..
```
THenry14, line 30:
```suggestion
          git clone -b v1.1.1 https://github.com/starkware-libs/cairo.git cairo1
          git clone -b v2.2.0 https://github.com/starkware-libs/cairo.git cairo2
```

lib/src/main/kotlin/com/swmansion/starknet/data/types/Contracts.kt
```diff
@@ -124,7 +127,7 @@ data class Cairo0ContractDefinition(private val contract: String) {
 }
 
 @Serializable
-data class Cairo1ContractDefinition(private val contract: String) {
+data class CairoContractDefinition(private val contract: String) {
```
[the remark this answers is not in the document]
THenry14: I think the way you've done it is ok. Let's just make sure this is properly announced as a breaking change.

lib/src/test/kotlin/integration/account/AccountTest.kt
```diff
@@ -249,6 +249,7 @@ class AccountTest {
         // If this test starts randomly falling, try recalculating class hash.
         // 2. If it fails on CI, make sure to delete the compiled contracts before running this test.
         // Chances are, the contract was compiled with a different compiler version.
+        // 3. This test sometimes fails due to getNonce receiving higher (pending) nonce than addDeclareTransaction expects
```
THenry14, line 252: is there any way we could circumvent this?

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,47 +1,134 @@
...
+
+  OS=$(uname)
+  ARCH=$(uname -m)
+
+  if [ "$OS" == "Linux" -a "$ARCH" == "x86_64" ]; then
+    curl -LsSf "https://github.com/starkware-libs/cairo/releases/download/v${VERSION}/release-x86_64-unknown-linux-musl.tar.gz" | tar -xz -C "$OUT_DIR" || exit 1
+  elif [ "$OS" == "Darwin" -a "$ARCH" == "arm64" ]; then
+    curl -LsSf "https://github.com/starkware-libs/cairo/releases/download/v${VERSION}/release-aarch64-apple-darwin.tar" | tar -x -C "$OUT_DIR" || exit 1
+  elif [ "$OS" == "Darwin" -a "$ARCH" == "x86_64" ]; then
+    build_cairo_compilers "$VERSION" "$OUT_DIR"
```
THenry14, line 80: I'd either change it to be more general - not only for macs, or remove it altogether. Would probably go with the latter, as I don't think we should add more unnecessary code for us to maintain.

If a binary for system/arch is not available to download we could allow to get the path to compiler from env/variable passed to script or throw an error.

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,47 +1,134 @@
...
-  cargo build
-  cargo run --bin starknet-compile -- --version
-  cargo run --bin starknet-sierra-compile -- --version
-  popd
-fi
+echo "Compiling v1 contracts.."
 
-COMPILER_PATH=$(pwd)/../cairo/target/debug/starknet-compile
-SIERRA_PATH=$(pwd)/../cairo/target/debug/starknet-sierra-compile
+while IFS= read -r -d '' file; do
```
THenry14, line 114: it seems v1 and v2 are pretty similar, maybe we could do this in for loop?

lib/starknet-jvm.md
```diff
@@ -577,7 +577,7 @@ fun main(args: Array<String>) {
     // Import a compiled contract
     val contractCode = Path.of("contract.json").readText()
     val casmCode = Path.of("contract.casm").readText()
-    val contractDefinition = Cairo1ContractDefinition(contractCode)
+    val contractDefinition = CairoContractDefinition(contractCode)
     val casmContractDefinition = CasmContractDefinition(casmCode)
     val nonce = account.getNonce().send()
```
THenry14, line 583: let's also check javademo, perhaps it would also need similar change

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,47 +1,134 @@
...
+
+  if ! find "$REPO_ROOT/cairo$VERSION_SHORT"; then
+    echo "Cloning cairo repo..."
+    git clone -q https://github.com/starkware-libs/cairo.git "$REPO_ROOT/cairo$VERSION_SHORT"
+    pushd "$REPO_ROOT/cairo$VERSION_SHORT" || exit 1
+    git checkout -q "v$VERSION"
+    popd
+  else
+    echo "Found existing cairo repo, skipping cloning."
+  fi
```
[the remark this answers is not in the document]
DelevoXDG: > I don't think we should cater to specific userbase, and go out of our way to ensure they have seemless experience when they want to develop.

I guess you are right, just that it was supported before and macos x64 seems like a more popular platform supporting which would be (mostly) effortless.

> we can allow to pass custom compiler to the script or something

Seems a bit complicated for a new user to understand, specifically considering it's not just one path and the correct file structure with corelib relative to compilers has to be maintained.  

> but I wouldn't include repo cloning and building the compiler just because someone someday may potentially try to use mac x86 to contribute.

I agree that it's kind of an overhead, although it's just `47M` (for reference, release is around `~107M`). It is true that building each compiler will take around `1GB`, but it was happening before and no one complained. 

But if you disagree with having it there, what do you think about including **"Supported platforms"** section in readme instead, specifying what version are officially supported, and moving instruction about git cloning there.

lib/src/test/kotlin/integration/account/AccountTest.kt
DelevoXDG: Only by adding big sleep times, preventing nonce-changing tests from parallel running, separating gateway and RPC tests from each other, also adding sleep gaps between those. 

Since those are integration tests and meant for local runs for now, I wouldn't go into it. It would make more sense to try that after gateway removal, but I'm not sure it's worse the effort, we should have devent-rs by then and probably still won't rely on integration tests for regular and CI runs since their inconsistency.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Contracts.kt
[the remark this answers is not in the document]
DelevoXDG: Come to think of it, can't we consider Cairo 2 to be an extension of Cairo 1?
Then we could still call the old dataclass `Cairo1ContractDefinition` and 
```kotlin
typealias Cairo2ContractDefinition = Cairo1ContractDefinition
``` 
And if something changes and Cairo2 contracts stop being supported via `Cairo1ContractDefinition`, we would remove typealias just redefine `Cairo2ContractDefinition`, while not affecting the old dataclass in any way, both now and in case of such change. 

I'm okay with either approach, but we could avoid making it breaking that way, while not creating any confusion by having a version-less `CairoContractDefinition`. What do you think?

lib/src/test/kotlin/integration/account/AccountTest.kt
```diff
@@ -182,14 +182,14 @@ class AccountTest {
         assertNotEquals(Felt(0), feeEstimate.overallFee)
     }
 
+    @Disabled
+    // TODO: (#311) re-enable once randomized contract source generation is supported
```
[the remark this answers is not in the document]
DelevoXDG: You mean here (#311) or in the comment? 😅

lib/src/test/resources/compileContracts.sh
DelevoXDG: They have different flags, I think making it a loop and specifying what flags go to which loop would make it less readable.

lib/starknet-jvm.md
DelevoXDG: No, currently we have Cairo0 in javademo.

lib/src/test/resources/compileContracts.sh
DelevoXDG: Are there any other platforms? This is one is popular, used by us on CI, is compatible with all of our dependencies, and effortless to "maintain". Also, it seems to me that making it more general would hide the fact that of all of the other platforms only macOS x64 is supported. 

I mentioned as to why approach with path seems problematic to me [in other comment](https://github.com/software-mansion/starknet-jvm/pull/285#discussion_r1324164073).

lib/src/test/resources/compileContracts.sh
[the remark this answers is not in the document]
THenry14: Noone complained, because there was no other way 😢  now we at least have _some_ options, and are not forced to compile ourselves.

> what do you think about including "Supported platforms" section in readme instead

this is not a matter of supported platforms, because these do not change (it refers to the platforms you can run starknet-jvm on - and not platform you can contribute from). This is rather a topic of "you want to contribute? if you're on linux x86 or apple sillicon, just do it. Otherwise, you will need to compile the cairo compiler first".  So this is rather somethign that could go to contribution guide.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Contracts.kt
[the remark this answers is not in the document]
THenry14: for now - yes. how will it look like in the future - I have no idea.

I like both ideas, and think they are both good. Both approaches seems to be future-proof. It all depends on how much you do not want to make breaking changes. Personally, I'm quite ok with this one as it is not very compilated to adjust to. But I understand a need to not introduce breaking changes, so it may make sense to do what you've just suggested. I leave the final decision to you.

lib/src/test/resources/compileContracts.sh
THenry14: >Are there any other platforms? 

Loads of :D 

>Thi one is popular, used by us on CI

but we will checkout and compile cairo on CI, because we have to, right? So I'd rather do everything that is required there. Especially so that this platform is dying, and we plan to move away from it to apple sillicon on CI eventually, right?

> Also, it seems to me that making it more general would hide the fact that of all of the other platforms only macOS x64 is supported.

supported platforms != platforms you can use to contribute to starknet-jvm from

lib/src/test/resources/compileContracts.sh
THenry14: we could loop over specific compile commands, seems to me this is the only part that varies, right?

lib/src/test/resources/compileContracts.sh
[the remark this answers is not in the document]
DelevoXDG: > and are not forced to compile ourselves.

Because of CI we are 😅 And since we are anyway for now, in my vision it's not something hard to keep available to everyone. 

But I assume we're going forward with removing it here and moving it to contribution guidelines then as `"Platforms you can contribute from"`?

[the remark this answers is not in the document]
THenry14: For now - yes. As deiscussed here https://github.com/software-mansion/starknet-jvm/pull/285#discussion_r1324295333

>But I assume we're going forward with removing it here and moving it to contribution guidelines then as "Platforms you can contribute from"?

that's the beauty - you can contribute from any platform you want :+1: the only requirement is that if you're not using linux x64 or apple sillicon, you need to compile the compiler first, and that's basically it.

lib/src/test/resources/compileContracts.sh
DelevoXDG: > supported platforms != platforms you can use to contribute to starknet-jvm from

I see, my bad using those interchangeably.

> but we will checkout and compile cairo on CI, because we have to, right? So I'd rather do everything that is required there.

Do you have moving the entire block with compiling contracts to CI? Both script and checks might get more complicated it seems to me. Certainly in such scenario no contribution guidelines can be added.

THenry14: No biggie :) 

I suppose we could just compile the compiler on CI (macos only), and then allow for scripts to use custom compilers (either by env variable or an argument passed to script, former probably makes more sense). I guess this will make sure that everything works for everyone, and is not very complicated. Contribution guide would then have a general info on where to get cairo code from, how to compile it and how to pass it to script.

Sounds good?

THenry14: Can't wait for us to finally migrate to devnet-rs and scarb :) every problem we discussed will be fixed by this

.github/workflows/checks.yml
```diff
@@ -15,19 +15,69 @@ jobs:
...
+          path: cairo2/target/release
+          key: ${{ runner.os }}-cairo2-target-release-${{ hashFiles('cairo2/Cargo.lock') }}
+
+      - name: Build Cairo compiler binaries (macOS only)
+        if: matrix.os == 'macos-latest'
+        run: |
+          if [[ ! -d "$V1_COMPILER_BUILD_PATH" || ! -d "$V2_COMPILER_BUILD_PATH" ]]; then
+            curl -sSf https://sh.rustup.rs | sh -s -- -y
+            if [[ ! -d "$V1_COMPILER_BUILD_PATH" ]]; then
+               cd ${{ github.workspace }}/cairo1
```
THenry14, line 54: I guess it's fine and it works, but I'd encourage you to use `pushd` and `popd` instead

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,47 +1,113 @@
...
+  local COMPILER_BUILD_PATH="${!COMPILER_BUILD_PATH_VAR}"
+
+  echo "Fetching compiler v$VERSION binaries..."
+
+  mkdir -p "$OUT_DIR"
+
+  OS=$(uname)
+  ARCH=$(uname -m)
+
+  if [ "$OS" == "Linux" -a "$ARCH" == "x86_64" ]; then
```
THenry14, line 38:
```suggestion
  if [ "$OS" == "Linux" && "$ARCH" == "x86_64" ]; then
```

`-a` is rather poorly defined, it is a good practice to use `&&` instead. More rationale -> https://www.shellcheck.net/wiki/SC2166

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,47 +1,113 @@
 #!/bin/bash
 
-pushd "$(dirname "$0")" || exit
-mkdir -p "compiled_v0"
+V1_COMPILER_BUILD_PATH_VAR="V1_COMPILER_BUILD_PATH"
+V2_COMPILER_BUILD_PATH_VAR="V2_COMPILER_BUILD_PATH"
```
THenry14, line 4: what does this do?

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,47 +1,113 @@
...
+        mkdir -p "$(dirname "$0")/$OUT_DIR/cairo/bin/"
+        mkdir -p "$(dirname "$0")/$OUT_DIR/cairo/corelib/"
+
+        echo "Copying binaries..."
+        rsync -aW "${COMPILER_BUILD_PATH}/" "$(dirname "$0")/$OUT_DIR/cairo/bin/" || exit 1
+        echo "Copying corelib..."
+        rsync -aW "$REPO_ROOT/cairo$VERSION_SHORT/corelib/" "$(dirname "$0")/$OUT_DIR/cairo/corelib/" || exit 1
+      else
+        echo "Your OS or architecture ($ARCH-$OS) is not supported directly."
+        echo "To proceed, please specify a valid path to a built v$VERSION compiler for ($ARCH-$OS) in the $COMPILER_BUILD_PATH_VAR environment variable."
```
THenry14, line 60: not sure if I follow, but from what I see you have to actually export a different var (eg `V1_COMPILER_BUILD_PATH_VAR` or `V2_COMPILER_BUILD_PATH_VAR`)? or am I missing something?

lib/src/test/resources/compileContracts.sh
THenry14: the same goes for all occurences of `-a` in this script

lib/src/test/resources/compileContracts.sh
DelevoXDG: Those are just names of env variables we are retrieving actual paths from.
1. I thought it would be good to have a top level declaration that specifies the names should they be changed, hence `_VAR` postfix.
2. We can later reference those names when throwing error.

lib/src/test/resources/compileContracts.sh
DelevoXDG: Are you referring to `$COMPILER_BUILD_PATH_VAR` in echo?
It is defined above and takes value of `V1_COMPILER_BUILD_PATH_VAR` or `V2_COMPILER_BUILD_PATH_VAR` depending on current version.
```
  [ "$VERSION_SHORT" == "1" ] && local COMPILER_BUILD_PATH_VAR="$V1_COMPILER_BUILD_PATH_VAR"
  [ "$VERSION_SHORT" == "2" ] && local COMPILER_BUILD_PATH_VAR="$V2_COMPILER_BUILD_PATH_VAR"
  local COMPILER_BUILD_PATH="${!COMPILER_BUILD_PATH_VAR}"
```

lib/src/test/resources/compileContracts.sh
THenry14: they probably won't, and even if so - it won't hurt us in any way, as this script is just our helper and nothing user-facing

lib/src/test/resources/compileContracts.sh
THenry14: but this variable has assigned a value of V1_COMPILER_BUILD_PATH_VAR, right? so the echo will print out the path and not the variable name (which I suppose you intend to do?)

DelevoXDG: `COMPILER_BUILD_PATH_VAR` variable is assigned the name of either `V1_COMPILER_BUILD_PATH_VAR` or `V2_COMPILER_BUILD_PATH_VAR`, not value.

And it is intended to print the name, not path, otherwise it wouldn't give any info on what variable to set.

lib/src/test/resources/compileContracts.sh
DelevoXDG: I know it probably won't, but I don't see how having them up top is bad as well. Having those names defined also helps with printing error reason with `$COMPILER_BUILD_PATH_VAR`.

Or do you have some specific change in mind?

lib/src/test/resources/compileContracts.sh
THenry14: ahh, I see now. You have a variable to hold a different variable name. I don't see ny reason why it should be done like this to be honest, it looks to me like a weird workaround. Why not just check for V1_COMPILER_BUILD_PATH or V2_COMPILER_BUILD_PATH existence, and use it as COMPILER_BUILD_PATH_VAR? the code would be less complicated, possibly even shorter

lib/src/test/resources/compileContracts.sh
THenry14: imho the more unnecessary code the worse. you can achieve exactly the same thing by just checking the variables existence directly and assigning one or the other to your general variable. the error should be then moved to this place and the script should fail fast if none of those is defined. wdyt?

DelevoXDG: I was basing this on a convention to declare constants and export env variables at the top level. Since we also need to specifically reference the name of those env variables to print an error message, at top levels we could instead have names of those variables (to avoid situation where debug info is not accurate), not their values.

I don't mind the approach you suggest though, I thought of it in the first place but then replaced it based on convention. If you don't find this convincing, I will change it.

THenry14: Sure I get this, and appreciate your idea, but I think this approach is more sensible when you actually declare some constants and not try to declare env variables. My understanding is that those variables have to be exported beforehand, so I think this just brings a bit of confusion.

so to sum up, so we're on the same page, my proposition is to:
- at the top of the script have a check for both variables
- if one of them (or maybe even both? I think we could allow this - what do you think?) is set, we can continue
- if none of them is set we print out an error that there are no cairo binaries and variables X and/or Y are not set

sounds good?

DelevoXDG: Sure.
As for compilers, they require different flags, so maybe we should require both to be set. I'm not sure v2 compilers compiles v1 contracts and v1 definitely won't compile v2 contracts.

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,47 +1,126 @@
...
+readonly V2_COMPILER_BIN_PATH="compilers/v2"
+
+readonly V1_COMPILER_VERSION="1.1.1"
+readonly V2_COMPILER_VERSION="2.2.0"
+
+if [ "$OS" == "Linux" ] && [ "$ARCH" == "x86_64" ]; then
+  true
+elif [ "$OS" == "Darwin" ] && [ "$ARCH" == "arm64" ]; then
+  true
+elif [ ! -d "$V1_COMPILER_BUILD_PATH" ] || [ ! -d "$V2_COMPILER_BUILD_PATH" ]; then
```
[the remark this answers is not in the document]
DelevoXDG: My intention here was only to check whether environment variables are required, right on start, as you previously suggested.

I assume you are referring to extracting the functionality of `fetch_compilers` and we don't wanna go back to checking
```bash
if [ ! -d "$V1_COMPILER_BUILD_PATH" ] || [ ! -d "$V2_COMPILER_BUILD_PATH" ]; then
```
inside `fetch_compilers`?
But then, we would have duplicate `curl`s, setting `OUTDIR`s, setting `COMPILER_PATH` and `SIERRA_PATH` separately. It seems that reducing duplication in one place, we will create it somewhere else 😅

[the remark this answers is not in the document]
THenry14: not really, what I had in mind was to also to refactor fetch_compilers a bit to fit into this. But then, on the second thought, maybe this is not worth it 🤔 it is a nitpick either way, so it does not make sense to block this PR, as current behaviour is passable, I guess

THenry14 (approved):

PR #289, opened 2023-08-04
Fix getEstimateFee for DeclareTransaction in RPC

### LINKED ISSUE #288: getEstimateFee for DeployAccountTransaction generates incorrect request when using RPC
Currently, `getEstimateFee` builds a request where:
- `program` contains a map when it should contain a string encoded from said map 
- there's a field `entryPointsByType` when it should be `entry_points_by_type`

DelevoXDG:
## Describe your changes

- Changed the way TransactionPayload is serialised in RPC
- Slightly modified TransactionPayload dataclasses

## Linked issues

Closes #288 

## Breaking changes


lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionPayloadSerializer.kt
```diff
@@ -0,0 +1,53 @@
...
+            else -> throw IllegalArgumentException("Invalid transaction type '${element.jsonObject["type"]?.jsonPrimitive?.content}'")
+        }
+    }
+
+    private fun deserializeDeclare(decoder: JsonDecoder, element: JsonElement): DeclareTransactionPayload =
+            when (element.jsonObject["version"]?.jsonPrimitive?.content) {
+                Felt.ONE.hexString() -> decoder.json.decodeFromJsonElement(DeclareTransactionV1Payload.serializer(), element)
+                Felt(2).hexString() -> decoder.json.decodeFromJsonElement(DeclareTransactionV2Payload.serializer(), element)
+                else -> throw IllegalArgumentException("Invalid declare transaction version")
+            }
```
DelevoXDG, line 38: Those were moved from the old `JsonRpcTransactionPayloadSerializer` but I'm not sure we ever need to deserialize `TransactionPayload` as this only something we use for building requests and never get from responses. 
There are no tests for that and I'm not sure we want to write these as this is something that is not going to be used.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionPayloadSerializer.kt
```diff
@@ -0,0 +1,53 @@
...
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializer
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.*
+
+@OptIn(ExperimentalSerializationApi::class)
+@Serializer(forClass = TransactionPayload::class)
+object JsonRpcTransactionPayloadSerializer : KSerializer<TransactionPayload> {
```
DelevoXDG, line 15: Not sure about the name, could leave it as `JsonRpcTransactionPayloadPolymorphicSerializer`, but that implies inheriting from `JsonContentPolymorphicSerializer<T>` from what I understand, which is not the case here.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionPayloadSerializer.kt
DelevoXDG, line 1: Maybe move all of these to `JsonRpcTransactionPayloadSerializer`? Added to the same file where DeclareTransactionPayload serializers were defined previously for now.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionPayloadSerializer.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    override fun serialize(encoder: Encoder, value: TransactionPayload) {
+        require(encoder is JsonEncoder)
+
+        val jsonObject = when (value) {
+            is InvokeTransactionPayload -> Json.encodeToJsonElement(InvokeTransactionPayloadSerializer, value).jsonObject
+            is DeclareTransactionV1Payload -> Json.encodeToJsonElement(DeclareTransactionV1PayloadSerializer, value).jsonObject
+            is DeclareTransactionV2Payload -> Json.encodeToJsonElement(DeclareTransactionV2PayloadSerializer, value).jsonObject
+            is DeployAccountTransactionPayload -> Json.encodeToJsonElement(DeployAccountTransactionPayloadSerializer, value).jsonObject
+            else -> throw IllegalArgumentException("Invalid transaction payload type")
```
THenry14, line 48: Let's print out what version was returnec, easeir to debug

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionPayloadSerializer.kt
```diff
@@ -65,3 +67,58 @@ object DeclareTransactionV2PayloadSerializer : KSerializer<DeclareTransactionV2P
...
+            put("type", value.type.toString())
+        }
+
+        encoder.encodeJsonElement(jsonObject)
+    }
+
+    override fun deserialize(decoder: Decoder): DeployAccountTransactionPayload {
+        throw SerializationException("Class used for serialization only.")
+    }
+}
```
THenry14, line 124: why those are needed? I though the data classes were already `@Serializable` and not much transformation is needed?

THenry14 (commented):
overall looking good :+1: @bartekryba please also take a look

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionPayloadSerializer.kt
DelevoXDG: They are, but the fail to serialise `type` field when `encodeToJsonElement` is used. 
Since we already have `DeclareTransactionV1PayloadSerializer` and `DeclareTransactionV2PayloadSerializer` I think it would be best way to solve this.

THenry14 (approved):

bartekryba (approved):
Looks good to me, good job

PR #300, opened 2023-08-21
Add basic flow examples and corresponding tests

### LINKED ISSUE #278: Add basic flow examples to docs and README
### Feature Request

- Add invoke/call example
- Add transfer token example
- Add declare cairo 1 contract example

DelevoXDG:
## Describe your changes

- Add invoke/call (transfer/balanceOf) examples
- Add const nonce account for tests
- Add invoke/call/deployAccount/estimateFee(invoke) tests
- Add revert_reason to mockedReceipt

## Linked issues

Closes #278 

## Breaking changes


lib/src/test/kotlin/integration/account/AccountTest.kt
```diff
@@ -351,4 +385,124 @@ class AccountTest {
...
+    @MethodSource("getAccounts")
+    fun `deploy account`(accountAndProvider: AccountAndProvider) {
+        assumeTrue(IntegrationConfig.isTestEnabled(requiresGas = true))
+
+        val (account, provider) = accountAndProvider
+
+        val privateKey = Felt(System.currentTimeMillis())
+        val publicKey = StarknetCurve.getPublicKey(privateKey)
+//        println(privateKey)
+//        println(publicKey)
```
THenry14, line 444: 👀

lib/src/test/kotlin/integration/account/AccountTest.kt
```diff
@@ -351,4 +385,124 @@ class AccountTest {
...
+            calldata = calldata,
+            salt = salt,
+        )
+
+        val deployedAccount = StandardAccount(
+            deployedAccountAddress,
+            privateKey,
+            provider,
+        )
+//        println(deployedAccountAddress)
```
THenry14, line 460: 👀

lib/src/test/kotlin/integration/utils/IntegrationConfig.kt
```diff
@@ -65,6 +67,8 @@ class IntegrationConfig {
                 privateKey = Felt.fromHex(
                     env.getOrElse("INTEGRATION_PRIVATE_KEY") { throw RuntimeException("INTEGRATION_PRIVATE_KEY not found in environment variables") },
                 ),
+                constNonceAccountAddress = env["INTEGRATION_CONST_NONCE_ACCOUNT_ADDRESS"]?.let { Felt.fromHex(it) },
+                constNoncePrivateKey = env["INTEGRATION_CONST_NONCE_PRIVATE_KEY"]?.let { Felt.fromHex(it) },
```
THenry14, line 71: is there any info about this in readme/else?
Why not just reuse acountAddress/privateKey? you can just export different env file (that of const nonce account) and have the same functionality, right?

THenry14 (commented):
looking good :+1:

lib/src/test/kotlin/integration/utils/IntegrationConfig.kt
DelevoXDG: > is there any info about this in readme/else?

I thought it might be better to leave for internal use only to not further confuse anyone, but I can add some basic info to readme.

>Why not just reuse acountAddress/privateKey? 

We are reusing acountAddress/privateKey it if constNonceAccountAddress/PrivateKey is not passed, but it could lead to unpredictable behaviour. I added the comment explaining that due to getNonce having access to pending block it retrieves higher nonce than other methods expect which leads to those methods failing.

> you can just export different env file (that of const nonce account) and have the same functionality, right?

Not really, because some tests change the state and some don't and the idea is to be able to run both simultaneously if needed: like fetching the eth balance and sending a transfer transaction. The need to supply a different env file each time would diminish the idea of having to use just one flag to switch between test mode and would be even harder to explain in readme. Not having info on what tests rely on constant nonce address would likely lead to someone running a increasing nonce test case, disrupting the essential function of using a constant nonce address, as the 'constant' nonce wouldn't remain constant anymore.

README.md
```diff
@@ -162,37 +163,396 @@ fun main(args: Array<String>) {
...
+        // Import a compiled contract
+        Path contractPath = Paths.get("contract.json");
+        String contractCode = String.join("", Files.readAllLines(contractPath));
+        Cairo0ContractDefinition contractDefinition = new Cairo0ContractDefinition(contractCode);
+        // Class hash is calculated using the tools you used for compilation (only for Cairo v0 contracts)
+        Felt classHash = Felt.fromHex("0x1a3b2c");
+        Felt nonce = account.getNonce().send();
+
+        // Estimate fee for declaring a contract
+        DeclareTransactionV1Payload declareTransactionPayloadForFeeEstimate = account.signDeclare(contractDefinition, classHash,new ExecutionParams(nonce, new Felt(1000000000000000L)), false);
```
tomek0123456789, line 399: 🤭 
```suggestion
        DeclareTransactionV1Payload declareTransactionPayloadForFeeEstimate = account.signDeclare(contractDefinition, classHash, new ExecutionParams(nonce, new Felt(1000000000000000L)), false);
```

README.md
```diff
@@ -162,37 +163,396 @@ fun main(args: Array<String>) {
...
+        classHash = classHash,
+        params = ExecutionParams(nonce, Felt(1000000000000000L)),
+        forFeeEstimate = true,
+    )
+    val feeEstimateRequest = provider.getEstimateFee(listOf(declareTransactionPayloadForFeeEstimate))
+    val feeEstimate = feeEstimateRequest.send().first().overallFee
+    // Make sure to prefund the account with enough funds to cover the fee for declare transaction
+
+    // Declare a contract
+    val params = ExecutionParams(nonce, Felt(feeEstimate.value.multiply(BigInteger.TEN)))
```
tomek0123456789, line 446: Even though `nonce` speaks for itself, I'd use named arguments here and in other similar classes, especially because these are code examples for documentation

README.md
```diff
@@ -162,37 +163,396 @@ fun main(args: Array<String>) {
...
+    val declareTransactionPayload = account.signDeclare(
+        sierraContractDefinition = contractDefinition,
+        casmContractDefinition = casmContractDefinition,
+        params = params,
+    )
+
+    val request = provider.declareContract(declareTransactionPayload)
+    val response = request.send()
```
tomek0123456789, line 552: Just a detail, since it is in the other code example:
```suggestion
    val request = provider.declareContract(declareTransactionPayload)
    val response = request.send()
    val hash = response.transactionHash // Hash of the transaction that declares a contract
```

tomek0123456789 (commented):
Looks great!

war-in (approved):

integration_tests.env.example
```diff
@@ -4,3 +4,5 @@ INTEGRATION_GATEWAY_URL=http://example.com/gateway
 INTEGRATION_FEEDER_GATEWAY_URL=http://example.com/feeder_gateway
 INTEGRATION_PRIVATE_KEY=0x123456789
 INTEGRATION_ACCOUNT_ADDRESS=0x123456789123456789123456789123456789123456789123456789123456789
+INTEGRATION_CONST_NONCE_PRIVATE_KEY=0x1a1a1a1a1a
+INTEGRATION_CONST_NONCE_ACCOUNT_ADDRESS=0x2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2b2
```
THenry14, line 8: let's add a newline at the end of this file

THenry14 (approved):
great job mate 🔥

PR #305, opened 2023-09-01
Add support for getBlockWithTxHashes

### LINKED ISSUE #301: Introducing starknet_getBlockWithTxHashes RPC method
### Feature Request

Introduce a `getBlockWithTxsHashes(blockId: BlockId)` method in JsonRpcProvider which would call the `starknet_getBlockWithTxHashes` rpc method. 

https://github.com/starkware-libs/starknet-specs/blob/6fadca1f1abf48523c0aa799a10061f40f973468/api/starknet_api_openrpc.json#L11

DelevoXDG:
## Describe your changes

- Add support for starknet_getBlockWithTxHashes
- Add devnet and integration tests for starknet_getBlockWithTxHashes
- Add missing integration tests for starknet_getBlockWithTxs

## Linked issues

Closes #301 

## Breaking changes


lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcGetBlockWithTransactionHashesPolymorphicSerializer.kt
DelevoXDG, line 1: Should we merge `JsonRpcGetBlockWithTransactionHashesPolymorphicSerializer.kt` and  `JsonRpcGetBlockWithTransactionsPolymorphicSerializer.kt` into a single file `JsonRpcGetBlockWithPolymorphicSerializer.kt` or `JsonRpcGetBlockPolymorphicSerializer.kt`?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -156,6 +156,7 @@ data class BlockWithTransactionsResponse(
     override val sequencerAddress: Felt,
 
     // Block body
+    // TODO (after reviews): move above block_header to align with other similar responses
```
DelevoXDG, line 159: remove once moved.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcGetBlockWithTransactionHashesPolymorphicSerializer.kt
THenry14: Yes, I think it may make sense.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
THenry14: let's just do it :+1: it may be beneficial to include all the necessary changes in reviews.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -167,7 +168,7 @@ data class BlockWithTransactionsResponse(
 
 @Serializable
 data class PendingBlockWithTransactionsResponse(
-    // Not in RPC schema
+    // Not in RPC schema, but can be returned by nodes
```
THenry14, line 171: how is that? :D

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -191,6 +192,67 @@ data class PendingBlockWithTransactionsResponse(
...
+
+@Serializable
+data class BlockWithTransactionHashesResponse(
+    @SerialName("status")
+    val status: BlockStatus,
+
+    // Block body
+
+    @SerialName("transactions")
+    override val transactionHashes: List<Felt>,
```
THenry14, line 210: shouldn't `transactions` also hold tx info? 
https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L1305

THenry14 (commented):
looking good, great job! 🔥

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: Ok, but it diff would make it hard to review.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: We should be looking at `"BLOCK_WITH_TX_HASHES"` instead:
https://github.com/starkware-libs/starknet-specs/blob/d47828436349a163e0d02cc27d7dff98a1d5b28b/api/starknet_api_openrpc.json#L1371C13-L1371C13

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -191,6 +192,67 @@ data class PendingBlockWithTransactionsResponse(
...
+@Serializable
+data class PendingBlockWithTransactionHashesResponse(
+    // Not in RPC schema, but can be returned by nodes
+    @SerialName("status")
+    val status: BlockStatus = BlockStatus.PENDING,
+
+    // Block body
+
+    @SerialName("transactions")
+    override val transactionHashes: List<Felt>,
```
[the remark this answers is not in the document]
DelevoXDG: Replied [in the previous thread](https://github.com/software-mansion/starknet-jvm/pull/305#discussion_r1321335898)

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: Just an observation, it was noticed when using our node.

THenry14: so let's clarify the comment "not in schema, but can work sometimes" actually says very little

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
THenry14: 🙈 

sorry, checked the wrong endpoint

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: Do you have something specific to add here in mind? I added this clarification so we know this is not gateway specific field and that it's probably shouldn't be removed despite not being present in RPC.

THenry14: The first part of Tomek's comment should be enough context for any future dev

DelevoXDG: I'm not sure that part actually gives any info, because it refers to the endpoint `getBlockWithTxs` in general. It's true that some nodes do not even support this endpoint for pending block, but the response for pending block is present in the spec.

So I don't see how that info is relevant in that particular case, because this dataclass is already contains `pending` in its name. 

Maybe there was some confusion, because my comment only relates to a field, not entire dataclass, dataclass is very much present in the spec. 

@tomek0123456789 can you please confirm or clarify this? 😅

THenry14: Ah I see - so the method is in spec (and with this body) expect for status field which is added by us (and that can sometimes be returned by some nodes), is that correct?

If it is correct, then:
1. we set the status so I don't think it can/should be overwritten by the node's response
2. I'd rather treat it as some "value added" by us, besides as you mentioned `pending` is already in the name and this status should never shange
3. we can mark it via comment that it is not in spec, but because of 1. I don't think "but can be returned by nodes" is relevant in our case (as the status cannot be overriden) and we should probably ignore such field in node's response (as it is not in spec - and perhaps the spec is what we want to align to, not some random node implementation)

what do you guys think?

tomek0123456789: Oh, a misunderstanding, I thought this comment was about the class itself, not the field 😅, sorry for that.

DelevoXDG: 1. Yes, the value is set no matter what node returns.
2. I kind of agree, but I wanted to specify that aside from being "value added by us", this is something that actually reflects behaviour of some RPC nodes, which means it's not a gateway specific thing and not something to get rid of when we start removing other gateway-specific fields that are marked as `Not in RPC schema` as well.
3. I agree that it generally is irrelevant information because we are ignoring that field, but the idea was to specify that our "custom value" is something that can potentially "conflict" with what node returns. 

If you think we should have cleaner comments I don't mind reverting this one.

THenry14: Sure, I understand. Whatever suits you - I would probably remove this comment, but won't oppose if you decided to leave it :+1:

PR #306, opened 2023-09-01
Convert StarkNet to Starknet

### LINKED ISSUE #302: Convert StarkNet to Starknet
### Feature Request

We should convert StarkNet to Starknet in README, docs and elsewhere.

DelevoXDG:
## Linked issues

Closes #302 

## Breaking changes


war-in (approved):
I think we can leave `StarkNetDomain` as is. We probably do not want to change the hash

PR #324, opened 2023-09-28
Adjust `simulateTransactions` (#323, #314)

### LINKED ISSUE #314: Add reverted execution_invocation to InvokeTransactoinTrace (Trace API 0.4)
### Feature Request

Currently InvokeTransactoinTrace looks like this, 
```kotlin
@Serializable
data class InvokeTransactionTrace(
    @SerialName("validate_invocation")
    val validateInvocation: FunctionInvocation?,

    @SerialName("execute_invocation")
    val executeInvocation: FunctionInvocation?,

    @SerialName("fee_transfer_invocation")
    val feeTransferInvocation: FunctionInvocation?,
) : TransactionTrace()
```

This however does not match the spec:
```json
                            "execute_invocation": {
                                "description": "the trace of the __execute__ call or constructor call, depending on the transaction type (none for declare transactions)",
                                "oneOf": [
                                    {
                                        "$ref": "#/components/schemas/FUNCTION_INVOCATION"
                                    },
                                    {
                                        "type": "object",
                                        "properties": {
                                            "revert_reason": {
                                                "name": "revert reason",
                                                "description": "the revert reason for the failed execution",
                                                "type": "string"
                                            }
                                        }
                                    }
                                ]
                            },
```

We should properly handle reverted execute invocation.

### LINKED ISSUE #323: Loosen restrictions in `JsonRpcTransactionTracePolymorphicSerializer`
### Feature Request

As of now, some nodes, don't send the `TRANSACTION_TRACE` as intended by the spec: https://github.com/starkware-libs/starknet-specs/blob/fbf8710c2d2dcdb70a95776f257d080392ad0816/api/starknet_trace_api_openrpc.json#L156 
This is fixed and the fields are marked required in RPC 0.5.0, so they should be returned by nodes then.

Still, as of now some fields are not sent by some nodes and that causes error because `JsonRpcTransactionTracePolymorphicSerializer` expects them. 

This can be temporarily resolved to support the endpoint while RPC 0.4 is still around by checking only the most important fields and not expect all of them.

DelevoXDG:
## Describe your changes

- [x] #323 
- [x] #314

## Linked issues

Closes #323 
Closes #314 

## Breaking changes

- `InvokeTransactionTrace` is renamed to `InvokeTransactionTraceBase` with derived `RevertedInvokeTransactionTrace` and `InvokeTransactionTrace`.
  - `InvokeTransactionTraceBase` currently  not have `executeInvocation` anymore.
  - Instead, `InvokeTransactionTrace` has `executeInvocation` of type `FunctionInvocation`
  - And `RevertedInvokeTransactionTrace` has `executeInvocation` of type `RevertedFunctionInvocation`

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
@@ -7,7 +7,7 @@ import kotlinx.serialization.json.*
 internal object JsonRpcTransactionTracePolymorphicSerializer :
     JsonContentPolymorphicSerializer<TransactionTrace>(TransactionTrace::class) {
     private fun selectInvokeTransactionTraceDeserializer(jsonObject: JsonObject): DeserializationStrategy<out InvokeTransactionTrace> {
-        val executeInvocation = jsonObject["execute_invocation"]?.jsonObject ?: throw IllegalStateException("Invalid INVOKE_TXN_TRACE in response: execute_invocation is missing.")
+        val executeInvocation = jsonObject["execute_invocation"]?.jsonObject ?: throw IllegalStateException("Response from node contains invalid INVOKE_TXN_TRACE: execute_invocation is missing.")
         val isReverted = "revert_reason" in executeInvocation
 
         return when (isReverted) {
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
@@ -7,7 +7,7 @@ import kotlinx.serialization.json.*
 internal object JsonRpcTransactionTracePolymorphicSerializer :
     JsonContentPolymorphicSerializer<TransactionTrace>(TransactionTrace::class) {
     private fun selectInvokeTransactionTraceDeserializer(jsonObject: JsonObject): DeserializationStrategy<out InvokeTransactionTrace> {
-        val executeInvocation = jsonObject["execute_invocation"]?.jsonObject ?: throw IllegalStateException("execute_invocation is missing.")
+        val executeInvocation = jsonObject["execute_invocation"]?.jsonObject ?: throw IllegalStateException("Invalid INVOKE_TXN_TRACE in response: execute_invocation is missing.")
         val isReverted = "revert_reason" in executeInvocation
 
         return when (isReverted) {
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
@@ -6,15 +6,25 @@ import kotlinx.serialization.json.*
 
 internal object JsonRpcTransactionTracePolymorphicSerializer :
     JsonContentPolymorphicSerializer<TransactionTrace>(TransactionTrace::class) {
+    private fun selectInvokeTransactionTraceDeserializer(jsonObject: JsonObject): DeserializationStrategy<out InvokeTransactionTrace> {
+        val executeInvocation = jsonObject["execute_invocation"]?.jsonObject ?: throw IllegalStateException("execute_invocation is missing.")
+        val isReverted = "revert_reason" in executeInvocation
+
+        return when (isReverted) {
+            true -> RevertedInvokeTransactionTrace.serializer()
+            false -> CommonInvokeTransactionTrace.serializer()
+        }
+    }
+
     override fun selectDeserializer(element: JsonElement): DeserializationStrategy<out TransactionTrace> {
         val jsonObject = element.jsonObject
 
-        return when (jsonObject.keys) {
-            setOf("validate_invocation", "execute_invocation", "fee_transfer_invocation") -> InvokeTransactionTrace.serializer()
-            setOf("validate_invocation", "fee_transfer_invocation") -> DeclareTransactionTrace.serializer()
-            setOf("validate_invocation", "constructor_invocation", "fee_transfer_invocation") -> DeployAccountTransactionTrace.serializer()
-            setOf("function_invocation") -> L1HandlerTransactionTrace.serializer()
-            else -> throw IllegalArgumentException("Invalid transaction trace type")
+        return when {
+            "execute_invocation" in jsonObject -> selectInvokeTransactionTraceDeserializer(jsonObject)
+            "constructor_invocation" in jsonObject -> DeployAccountTransactionTrace.serializer()
+            "function_invocation" in jsonObject -> L1HandlerTransactionTrace.serializer()
+            listOf("validate_invocation", "fee_transfer_invocation").any { it in jsonObject } -> DeclareTransactionTrace.serializer()
+            else -> throw IllegalStateException("Unknown transaction trace type.")
         }
     }
 }
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
@@ -6,13 +6,13 @@ import kotlinx.serialization.json.*
 
 internal object JsonRpcTransactionTracePolymorphicSerializer :
     JsonContentPolymorphicSerializer<TransactionTrace>(TransactionTrace::class) {
-    private fun selectInvokeTransactionTraceDeserializer(jsonObject: JsonObject): DeserializationStrategy<out InvokeTransactionTrace> {
+    private fun selectInvokeTransactionTraceDeserializer(jsonObject: JsonObject): DeserializationStrategy<out InvokeTransactionTraceBase> {
         val executeInvocation = jsonObject["execute_invocation"]?.jsonObject ?: throw IllegalStateException("Response from node contains invalid INVOKE_TXN_TRACE: execute_invocation is missing.")
         val isReverted = "revert_reason" in executeInvocation
 
         return when (isReverted) {
             true -> RevertedInvokeTransactionTrace.serializer()
-            false -> CommonInvokeTransactionTrace.serializer()
+            false -> InvokeTransactionTrace.serializer()
         }
     }
 
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/SimulatedTransaction.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/SimulatedTransaction.kt
@@ -65,45 +65,72 @@ data class FunctionInvocation(
 )
 
 @Serializable
-sealed class TransactionTrace()
+data class RevertedFunctionInvocation(
+    @SerialName("revert_reason")
+    val revertReason: String,
+)
 
 @Serializable
-data class InvokeTransactionTrace(
-    @SerialName("validate_invocation")
-    val validateInvocation: FunctionInvocation?,
+sealed class TransactionTrace
 
-    @SerialName("execute_invocation")
-    val executeInvocation: FunctionInvocation?,
+@Serializable
+sealed class InvokeTransactionTrace : TransactionTrace() {
+    @SerialName("validate_invocation")
+    abstract val validateInvocation: FunctionInvocation?
 
     @SerialName("fee_transfer_invocation")
-    val feeTransferInvocation: FunctionInvocation?,
-) : TransactionTrace()
+    abstract val feeTransferInvocation: FunctionInvocation?
+}
+
+@Serializable
+data class CommonInvokeTransactionTrace(
+    @SerialName("validate_invocation")
+    override val validateInvocation: FunctionInvocation? = null,
+
+    @SerialName("execute_invocation")
+    val executeInvocation: FunctionInvocation? = null,
+
+    @SerialName("fee_transfer_invocation")
+    override val feeTransferInvocation: FunctionInvocation? = null,
+) : InvokeTransactionTrace()
+
+@Serializable
+data class RevertedInvokeTransactionTrace(
+    @SerialName("validate_invocation")
+    override val validateInvocation: FunctionInvocation? = null,
+
+    @SerialName("execute_invocation")
+    val executeInvocation: RevertedFunctionInvocation? = null,
+
+    @SerialName("fee_transfer_invocation")
+    override val feeTransferInvocation: FunctionInvocation? = null,
+) : InvokeTransactionTrace()
 
 @Serializable
 data class DeclareTransactionTrace(
     @SerialName("validate_invocation")
-    val validateInvocation: FunctionInvocation?,
+    val validateInvocation: FunctionInvocation? = null,
 
     @SerialName("fee_transfer_invocation")
-    val feeTransferInvocation: FunctionInvocation?,
+    val feeTransferInvocation: FunctionInvocation? = null,
 ) : TransactionTrace()
 
 @Serializable
 data class DeployAccountTransactionTrace(
     @SerialName("validate_invocation")
-    val validateInvocation: FunctionInvocation?,
+    val validateInvocation: FunctionInvocation? = null,
 
     @SerialName("constructor_invocation")
-    val constructorInvocation: FunctionInvocation?,
+    val constructorInvocation: FunctionInvocation? = null,
 
     @SerialName("fee_transfer_invocation")
-    val feeTransferInvocation: FunctionInvocation?,
+    val feeTransferInvocation: FunctionInvocation? = null,
 ) : TransactionTrace()
 
 @Serializable
 data class L1HandlerTransactionTrace(
     @SerialName("function_invocation")
-    val functionInvocation: FunctionInvocation?,
+    val functionInvocation: FunctionInvocation? = null,
 ) : TransactionTrace()
 
 @Serializable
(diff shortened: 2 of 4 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
```diff
@@ -6,15 +6,25 @@ import kotlinx.serialization.json.*
 
 internal object JsonRpcTransactionTracePolymorphicSerializer :
     JsonContentPolymorphicSerializer<TransactionTrace>(TransactionTrace::class) {
+    private fun selectInvokeTransactionTraceDeserializer(jsonObject: JsonObject): DeserializationStrategy<out InvokeTransactionTrace> {
+        val executeInvocation = jsonObject["execute_invocation"]?.jsonObject ?: throw IllegalStateException("execute_invocation is missing.")
```
THenry14, line 10: let's improve on the exception messaging, to better point out what went wrong

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
```diff
@@ -6,15 +6,25 @@ import kotlinx.serialization.json.*
...
-            setOf("validate_invocation", "execute_invocation", "fee_transfer_invocation") -> InvokeTransactionTrace.serializer()
-            setOf("validate_invocation", "fee_transfer_invocation") -> DeclareTransactionTrace.serializer()
-            setOf("validate_invocation", "constructor_invocation", "fee_transfer_invocation") -> DeployAccountTransactionTrace.serializer()
-            setOf("function_invocation") -> L1HandlerTransactionTrace.serializer()
-            else -> throw IllegalArgumentException("Invalid transaction trace type")
+        return when {
+            "execute_invocation" in jsonObject -> selectInvokeTransactionTraceDeserializer(jsonObject)
+            "constructor_invocation" in jsonObject -> DeployAccountTransactionTrace.serializer()
+            "function_invocation" in jsonObject -> L1HandlerTransactionTrace.serializer()
+            listOf("validate_invocation", "fee_transfer_invocation").any { it in jsonObject } -> DeclareTransactionTrace.serializer()
```
THenry14, line 26: Let's remember to update this when 0.5.0 comes out - they add tx types in responses there, and it should make our life easier

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/SimulatedTransaction.kt
```diff
@@ -64,46 +64,73 @@ data class FunctionInvocation(
...
+    @SerialName("revert_reason")
+    val revertReason: String,
+)
+
 @Serializable
 sealed class TransactionTrace()
 
 @Serializable
-data class InvokeTransactionTrace(
+sealed class InvokeTransactionTrace : TransactionTrace() {
```
THenry14, line 77: what is the reason to have:

InvokeTransactionTrace
CommonInvokeTransactionTrace
RevertedInvokeTransactionTrace

?

I think just InvokeTransactionTrace and RevertedInvokeTransactionTrace should be enough?

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
DelevoXDG: I remember, I initially suggested `type` field to be added 😅 
Or do you want me to add a comment and link this to an issue?

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/SimulatedTransaction.kt
DelevoXDG: This was inspired by `TransactionReceipt`. The idea is to have interface dataclass with all the fields shared by both `Common` and `Reverted` invoke trace. The user might be unaware whether the trace in the response is reverted. It would be easier to initially cast to base invoke trace, not worrying that cast may fail, where you can already access some fields. And to access `executionInvocation`, only afterwards check whether it's common or reverted.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
THenry14: nah, if you remember about this that's fine for me :)

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/SimulatedTransaction.kt
THenry14: not sure I understand this correctly. 1. if we have a base txtrace - shouldn't reverted and not-reverted class then inherit from it? 2. what do you mean by "It would be easier to initially cast to base invoke trace, not worrying that cast may fail, where you can already access some fields"? I suppose that if cast fails and some fields are missing, there should be a response error right? we should always get all fields required by InvokeTransactionTrace OR RevertedInvokeTransactionTrace. or am I misunderstanding something?

DelevoXDG: 1. > shouldn't reverted and not-reverted class then inherit from it?

Yes, but we could also have intermediate dataclass that inherits from txtrace, that contains fields that are shared by reverted and non-reverted class, like we do with invoke v1 and v2 txs for example.

2. > I suppose that if cast fails and some fields are missing, there should be a response error right?

What I mean is that when `simulateTransactions` request is sent, the response is just raw `List<TransactionTrace>`. The user might be unaware that simulated invoke transaction is reverted and cast to `InvokeTransactionTrace`. If the current approach is used, the cast is safe and we can access all of the fields expect for `executionInvocation`. Otherwise, the cast will fail if transaction turns out to be reverted.

> we should always get all fields required by InvokeTransactionTrace OR RevertedInvokeTransactionTrace.

The problem is there's no strictly defined reverted invoke transaction trace in the spec:
```json
{
    "name": "INVOKE_TXN_TRACE",
    "type": "object",
    "description": "the execution trace of an invoke transaction",
    "properties": {
        "validate_invocation": {
            "$ref": "#/components/schemas/FUNCTION_INVOCATION"
        },
        "execute_invocation": {
            "description": "the trace of the __execute__ call or constructor call, depending on the transaction type (none for declare transactions)",
            "oneOf": [
                {
                    "$ref": "#/components/schemas/FUNCTION_INVOCATION"
                },
                {
                    "type": "object",
                    "properties": {
                        "revert_reason": {
                            "name": "revert reason",
                            "description": "the revert reason for the failed execution",
                            "type": "string"
                        }
                    }
                }
            ]
        },
        "fee_transfer_invocation": {
            "$ref": "#/components/schemas/FUNCTION_INVOCATION"
        }
    }
},
```
so forcing user to know our specific implementation in order to do a safe cast to a dataclass `InvokeTransactionTrace`, which user knowing the spec might assume works both for reverted and non-reverted trace, is suboptimal I think.

One other  potential solution could be to have optional `revert_reason` in `function_invocation` dataclass. This is however problematic because:
1.  `function_invocation` in all the other tx traces doesn't have `revert_reason`.
2. This deviates from the spec even more.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
DelevoXDG: What other details do you have in mind? 

The trace should already contain the info that:
1. `JsonRpcTransactionTracePolymorphicSerializer` failed.
2. Specifically, `selectInvokeTransactionTraceDeserializer` failed.
3. Specifically, because `execute_invocation` field is missing.

THenry14: `"execute_invocation is missing."`

let's add at least that's from simulate transaction endpoint; possibly we could print out the response we got and state it may be node's fault, that should help identify the real problem

DelevoXDG: Well technically the trace could be used elsewhere. And the stack trace would already contain info that `simulateTransactions` endpoint was used.

But I agree we could add info indicating that it may be the node's fault, I'm just not sure we ever print out the full response, or in that case, jsonObject when using serializers.

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/SimulatedTransaction.kt
THenry14: still not sure I follow:

> The user might be unaware that simulated invoke transaction is reverted and cast to InvokeTransactionTrace

why the user has to cast? shouldn't it be something that sdk is doing automatically? when I use simulate transaction, I assume that I will get a data class containing response, and that I won't have to cast it manually

> The problem is there's no strictly defined reverted invoke transaction trace in the spec:

if I understand the snippet you've sent correctly, you have `execute_invocation` that can be either `FUNCTION_INVOCATION` (succeeded) OR have a `revert_reason` (reverted). Doesn't that mean we have a strict definition? is it possible for succeded transaction to have `revert_reason` field?

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
THenry14: > Well technically the serializer could be used elsewhere.

do you have any specific place in mind? it's hard for me to imagine where someone would use trace transaction serializer, apart from trace transactions 🤔 

>  I'm just not sure we ever print out the full response, or in that case, jsonObject when using serializers.

sure, that was just an idea, not a requirement :+1: we should stick to conventions we have; if we do not do it anywhere, let's stick to it. Still, the message could be improved imo

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/SimulatedTransaction.kt
DelevoXDG: > when I use simulate transaction, I assume that I will get a data class containing response, and that I won't have to cast it manually

You will get List<SimulatedTransaction>, where `SimulatedTransaction` is:
```kotlin
@Serializable
data class SimulatedTransaction(
    @SerialName("transaction_trace")
    @Serializable(with = JsonRpcTransactionTracePolymorphicSerializer::class)
    val transactionTrace: TransactionTrace,

    @SerialName("fee_estimation")
    val feeEstimation: EstimateFeeResponse,
)
```
So, this field could be any `TransactionTrace` and we have to cast to a specific tx trace anyway to access fields specific to the derived dataclass.

> is it possible for succeded transaction to have revert_reason field?

No, what I meant unlike `PENDING_TXN_RECEIPT`, the's no explicitly defined something like `REVERTED_INVOKE_TXN_TRACE`, so the users might assume `InvokeTransactionTrace` is always valid dataclass to cast to because it is the only one present in the spec.

THenry14: but we can also serialize those transaction traces right? 

> So, this field could be any TransactionTrace and we have to cast to a specific tx trace anyway to access fields specific to the derived dataclass.

by "we" you mean sdk? I agree. and we should do it automatically, and when there is invoke tx trace - to reverted or non reverted one. but if by "we" you mean users, I don't think I agree.

> No, what I meant is that unlike PENDING_TXN_RECEIPT, the's no explicitly defined something like REVERTED_INVOKE_TXN_TRACE, so the users might assume InvokeTransactionTrace is always valid dataclass to cast to because it is the only that has spec counterpart INVOKE_TXN_TRACE.

Sure I get that, but the thing is we can easily determine if something is a reverted tx or non-reverted, right? then why can't we just have 2 dataclasses for it, but need to introduce the third (common) one? this part I don't quite understand, sorry.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcTransactionTracePolymorphicSerializer.kt
THenry14: > I meant other endpoints that are not simulateTransactions. We currently don't support those, but it could be used in starknet_traceTransaction or starknet_traceBlockTransactions endpoints.

Ah ok, so futureproofing. Sure, that makes sense then.

> I will add the info that the error might be caused by the node-specific implementation and response of simulateTransactions. We are unlikely to add any other trace api endpoints to SDK before 0.5.0 anyway?

let's keep the messaging simple, and just state that node sent over a wrong response (maybe let's not mention any node-specific implementations? my point is - we actually don't know, the potential error users encounters may be not connected to node implementation at all).

Yes, probably not before 0.5.0. Maybe after it is released, we'll see :+1:

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/SimulatedTransaction.kt
DelevoXDG: > by "we" you mean sdk? I agree. and we should do it automatically, and when there is invoke tx trace - to reverted or non reverted one. but if by "we" you mean users, I don't think I agree.

Kind of both. We have the interface `TransactionTrace`, and the SDK already will know the exact type of `TransactionTrace` by the time response is received, but not earlier than that. Meaning we cannot write any code that uses said fields because we don't know the exact types of tx traces in response at that time. Meaning something like that would have to be done:

```kotlin
if (trace is DeployAccountTransactionTrace){
    val deployAccountTrace = trace as DeployAccountTransactionTrace
    println(deployAccountTrace.constructorInvocation)
}
``` 
otherwise `.constructorInvocation` cannot be accessed.

Similarly for invoke tx trace like in the test:
```kotlin
val trace = simulationResult.first().transactionTrace
assertTrue(trace is InvokeTransactionTrace)
assertTrue(trace is RevertedInvokeTransactionTrace)
val revertedTrace = trace as RevertedInvokeTransactionTrace
assertNotNull(revertedTrace.executeInvocation)
assertNotNull(revertedTrace.executeInvocation?.revertReason)
```

> then why can't we just have 2 dataclasses for it, but need to introduce the third (common) one? 

Two reasons:
1. Because the user who uses the SDK for the first time would only be aware of `InvokeTransactionTrace` (`INVOKE_TXN_TRACE `) from the spec. So they do something like:
```
val trace = simulationResult.first().transactionTrace
val invokeTrace = trace as InvokeTransactionTrace
println(invokeTrace.executeInvocation?.callType)
```
being unaware `RevertedInvokeTransactionTrace` exists (no strict counterpart in the spec) and this fails.
2. `CommonInvokeTransactionTrace` and `InvokeTransactionTrace` share common fields. So something like this could be done without needing to know whether the trace is reverted or not:
```
val trace = simulationResult.first().transactionTrace
val invokeTrace = trace as InvokeTransactionTrace
println(invokeTrace.validateInvocation?.calldata)
```  
If we don't have that interface, something like this has to be done for all the common fields:
```kotlin
val trace = simulationResult.first().transactionTrace
val validateCalldata = when (trace){
    is InvokeTransactionTrace -> trace.validateInvocation?.calldata
    is RevertedInvokeTransactionTrace -> trace.validateInvocation?.calldata
    else -> throw IllegalStateException("Unexpected trace type.")
}
println(validateCalldata)
```
Seems less convenient to me.

THenry14: >     val deployAccountTrace = trace as DeployAccountTransactionTrace

this shouldn't be necessary - the cast should be done automatically, and `simulationResult` (the List) should hold already casted objects. We should just hold serialized datacalasses, ready to use. Something like this should be doable:

```
val trace = simulationResult.first().transactionTrace
if (trace is DeployAccountTransactionTrace){
    println(trace.constructorInvocation)
}
```

> Because the user who uses the SDK for the first time would only be aware of InvokeTransactionTrace (INVOKE_TXN_TRACE ) from the spec.

Sorry but I still don't see how this is relevant - everything should be done automnatically; if user do not have to cast, one doesn't have to be bothered with this. This is something a plymorphic serializer should do for us.

> CommonInvokeTransactionTrace and InvokeTransactionTrace share common fields

I believe I already said this - but yes, they share common fields, that's why I believe they should inherit both from thebase class. But manual casting, imo, is not a good way. Please consider `getTransaction` endpoint - you just call it and get ready, proper dataclass; no need to manually cast. I think this is a way to go.

If there is still some confusion, or perhaps you feel I still misunderstand some of your ideas, lets discuss offline as I feel this discussion is running circles at this point :+1:

DelevoXDG:
``` 
val trace = simulationResult.first().transactionTrace
if (trace is DeployAccountTransactionTrace){
    println(trace.constructorInvocation)
}
```
You're correct, this is doable.
But this isn't if there's no base class:
``` 
val trace = simulationResult.first().transactionTrace
if (trace is InvokeTransactionTrace){
    println(trace.validateInvocation)
}
```
We don't get the simple way to access common fields that could belong to either of classes at the point where we don't know which one we received in response, casting or no casting.

THenry14: But we can have a base class, that other classes inherit from, right? And then we can cast automatically - right after we get the response - all the items from the response to their respective dataclasses. Then we know exactly what we have, and the classes are final. Then we can use all of them straight away (exactly how it works for getTransaction), and both snippets are doable.

DelevoXDG: Discussed offline, changed naming to show purpose of classes more clearly. Will address similar naming in #329 .

THenry14 (approved):

PR #349, opened 2023-11-10
Support merkletree, selector, string in TypedData

### LINKED ISSUE #340: Support more types in `TypedData`
### Feature Request

Add support for:
- `string`
- `selector`
- `merkletree`

DelevoXDG:
## Describe your changes

- Added `MerkleTree` (public for now, could be made internal)?
- Support `merkletree`, `selector`, `string` base type names in TypedData
- Update example in description
- Add tests

## Linked issues

Closes #340 

## Breaking changes


.github/workflows/checks.yml
```diff
@@ -76,9 +76,9 @@ jobs:
 
       - name: Install sncast 0.6.0
         run: |
-          curl --proto '=https' --tlsv1.2 -LsSf https://raw.githubusercontent.com/foundry-rs/starknet-foundry/master/scripts/install.sh | sh -s -- -v 0.6.0
-          source /Users/runner/.bashrc || true
-          echo "$(dirname $(which sncast))" >> $GITHUB_PATH    
+          asdf plugin add starknet-foundry
+          asdf install starknet-foundry 0.6.0
+          asdf global starknet-foundry 0.6.0
```
THenry14, line 81: let's use:

- snfoundry gh action
- newe(r/st) version of snfoundry

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
```diff
@@ -11,35 +13,60 @@ import kotlinx.serialization.json.*
...
- *             "name": "Cow",
- *             "wallet": "0xCD2a3d9F938E13CD947Ec05AbC7FE734Df8DD826",
- *         },
- *         "to": {
- *             "name": "Bob",
- *             "wallet": "0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB",
- *         },
- *         "contents": "Hello, Bob!",
- *     },
+ *    "types": {
```
THenry14, line 16: How was this data obtained? it would be cool to have such info (command) somewhere in here as a comment

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TypedDataTypeBaseSerializer.kt
```diff
@@ -0,0 +1,18 @@
...
+import kotlinx.serialization.json.*
+
+internal object TypedDataTypeBaseSerializer : JsonContentPolymorphicSerializer<TypeBase>(TypeBase::class) {
+    override fun selectDeserializer(element: JsonElement): DeserializationStrategy<out TypeBase> {
+        val type = element.jsonObject["type"]?.jsonPrimitive?.content
+
+        return when (type) {
+            "merkletree" -> TypedData.MerkleTreeType.serializer()
+            is String -> TypedData.Type.serializer()
+            else -> throw IllegalArgumentException("Input element does not contain mandatory field 'type'")
```
THenry14, line 15: It may contain this field, but we are just not checking for it. " .. contains unknown type field: <type>" may be a better message

THenry14 (commented):
lgtm, I have tried to compare it with the python code from cairo-lang and it seems the implementation's looking good

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
DelevoXDG: It was obtained from starknet.js 😄, same as when TypedData was initially added.

THenry14: XD

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TypedDataTypeBaseSerializer.kt
DelevoXDG: > but we are just not checking for it.

Do you mean that that the field is present but has a different name? This message meant to show that the field named `type` is missing.

.github/workflows/checks.yml
DelevoXDG: The first version that outputs `snforge x.y.z` instead of `forge x.y.z` is 0.9.1, meaning every version before that is incompatible with gh action.

And snforge 0.9.1 is breaking and will require rewriting some devnet code. I would rather avoid complicating this PR and migrate to 10.0.2+ when it's out (assuming declare is fixed by then), since this is not priority and both setup-snfoundry and asdf approach take 2 seconds.

THenry14: what do you mean by breaking exactly?

THenry14:
I appreciate you improve tests (and other things), but please do so before you request a review. it is a bit confusing and increasingly difficult to re-review changes after you add commits not related to discussion.

I would really appreciate if you requested a review only after all the intended changes are here, and each commit that is added after the review is only related to the discussion raised in this PR. thanks 🙏

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TypedDataTypeBaseSerializer.kt
DelevoXDG:
```
val type = element.jsonObject["type"]?.jsonPrimitive?.content
```
`type` here is always `String?`, so the only way the else condition is reached is by `type` being null. I can add an extra check if you think it should be there though.

.github/workflows/checks.yml
THenry14: Can you elaborate a bit more on it? perhaps I miss some importatnt piece of information here, but I don't think anything in cast is this breaking

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TypedDataTypeBaseSerializer.kt
THenry14: perhaps, but just modifying the message will be enough imo, to include an information that it either does not exist or is unknown. you never know when allowed types change, my thinking here is maybe it is better to be more explicit with the error message

.github/workflows/checks.yml
DelevoXDG: https://github.com/software-mansion/starknet.swift/pull/114
Like I said, some changes to commands outputs, not much, but be changed when removing workaround is possible.

THenry14: ok but it should be possible now, right? this change is already included in 0.10.1 if I remember correctly

DelevoXDG: I can add changes to deserialization now, but removing workaround will have to come later as declare is not working properly as of 0.10.1

THenry14: ok, so you mean declaring multiple contracts as one package, and not commands outputs? 
Yes this will be a part of 0.11 probably. if this is blocking, then sure - let's wait.

THenry14 (approved):

Issue #326, opened 2023-10-02
Remove feeder_gateway


DelevoXDG:
### Feature Request

- Remove all the feeder_gateway methods from `GatewayProvider`
- `StandardAccount` should only accept `JsonRpcProvider` as a provider

PR #358, opened 2023-11-22
Support RPC 0.6.0

### LINKED ISSUE #356: Support Starknet 0.13.0
### Feature Request

Support RPC 0.6.0, with Starknet 0.13.0 limitations.
The rc1 of RPC 0.6.0 is already avaliable: https://github.com/starkware-libs/starknet-specs/compare/v0.5.1...v0.6.0-rc1

DelevoXDG:
## Linked issues

Closes #356 

## Breaking changes


lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -33,6 +33,15 @@ enum class TransactionType(val txPrefix: Felt) {
     L1_HANDLER(Felt.fromHex("0x6c315f68616e646c6572")), // encodeShortString('l1_handler')
 }
 
+@Serializable
+enum class DAMode {
```
DelevoXDG, line 37: It's `DA_MODE` in the spec, but naming this `DAMode` feels very non-java, so maybe we should deviate from the spec a bit and rename it to `DataAvailabilityMode`?

DelevoXDG:
Blocked by #326

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -41,18 +44,44 @@ data class DeployAccountResponse(
...
+        // Because of this, the L2 gas values are hardcoded to 0
+        return ResourceBoundsMapping(
+            l1Gas = ResourceBounds(maxAmount = maxAmount, maxPricePerUnit = maxPricePerUnit),
+            l2Gas = ResourceBounds(maxAmount = Uint64.ZERO, maxPricePerUnit = Uint128.ZERO),
+        )
+    }
+    private fun addOverhead(value: BigInteger, overhead: Double): BigInteger {
+        val multiplier = ((1 + overhead) * 100).roundToInt().toBigInteger()
+        return value.multiply(multiplier).divide(BigInteger.valueOf(100))
+    }
```
DelevoXDG, line 83: We typically don't define much methods inside files with just dataclasses.
However, I think the way it was done (just plain function), is sub-optimal, since we shouldn't really get max fee or resource bounds without having estimate response.

We could:
- move this particular response to separate file
- define those methods as extensions

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -396,11 +425,64 @@ data class PendingStateUpdateResponse(
...
+    val maxPricePerUnit: Uint128,
+) {
+    companion object {
+        @field:JvmField
+        val ZERO = ResourceBounds(Uint64.ZERO, Uint128.ZERO)
+    }
+
+    fun toMaxFee(): Felt {
+        return maxAmount.value.multiply(maxPricePerUnit.value).toFelt
+    }
```
DelevoXDG, line 456: Similar as above, convenience methods that could be changed to extensions, moved to another file or left as is.

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
```diff
@@ -14,59 +14,111 @@ interface Account {
...
-     * Sign a transaction.
+     * Sign a version 1 invoke transaction.
      *
      * Sign a transaction to be executed on Starknet.
      *
      * @param call a call to be signed
      * @param params additional execution parameters for the transaction
      * @param forFeeEstimate when set to `true`, it changes the version to `2^128+version` so the signed transaction can only be used for fee estimation
-     * @return signed invoke function payload
+     * @return signed invoke transaction version 1 payload
```
THenry14, line 24: is this v1 or v1/v2 ?

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
```diff
@@ -85,7 +137,27 @@ interface Account {
         maxFee: Felt,
         nonce: Felt,
         forFeeEstimate: Boolean,
-    ): DeployAccountTransactionPayload
+    ): DeployAccountTransactionV1Payload
+
+    /**
+     * Sign version 3 deploy account transaction.
+     *
+     * Sign a deploy account transaction that requires prefunding deployed address.
```
THenry14, line 145:
```suggestion
     * Sign a deploy account transaction that requires prefunding said account.
```
Not sure deployed is the best wording

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
```diff
@@ -85,7 +137,27 @@ interface Account {
...
+     *
+     * @param classHash hash of the contract that will be deployed. Has to be declared first!
```
THenry14, line 147:
```suggestion
     * @param classHash hash of the contract that will be deployed
```

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
```diff
@@ -103,10 +175,35 @@ interface Account {
...
+     */
+    fun signDeployAccount(
+        classHash: Felt,
+        calldata: Calldata,
+        salt: Felt,
+        l1ResourceBounds: ResourceBounds,
+        forFeeEstimate: Boolean = false,
+    ): DeployAccountTransactionV3Payload {
+        val params = DeployAccountParamsV3(
+            nonce = Felt.ZERO,
```
THenry14, line 201: are you sure it is safe to set nonce to 0 here?

DelevoXDG: Yes, as it was done previously in v1 deploy account. 
Nonce cannot be increased without deployed account as far as I am aware.

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
DelevoXDG: Only declare has v2 transactions.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/NumAsHexSerializers.kt
```diff
@@ -0,0 +1,48 @@
...
+) : KSerializer<T> {
+    override fun deserialize(decoder: Decoder): T {
+        val hex = decoder.decodeString()
+        return fromHex(hex)
+    }
+
+    override fun serialize(encoder: Encoder, value: T) {
+        encoder.encodeString(hexString(value))
+    }
+}
```
THenry14, line 48: isn't FeltSerializer missing here? (I mean deserialize/serialize)

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
THenry14: let's make an issue and think about it at later date

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
THenry14: Let's leave it as is

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/NumAsHexSerializers.kt
DelevoXDG: This is a generic serializer, it only encodes and decodes strings, not Felts. It's a base class for all hex serializers, including FeltSerializer

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: Sure, #375

THenry14 (approved):
we'll fix some of design issues etc in consecutive PRs

PR #359, opened 2023-11-23
[WIP] Remove GatewayProvider and LegacyDevnetClient 

DelevoXDG:
## Linked issues

Closes

## Breaking changes


lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
```diff
@@ -4,9 +4,8 @@ import com.swmansion.starknet.provider.Request
 import java.util.concurrent.CompletableFuture
 import java.util.function.Function
 
-// Whole HttpResponse is required for deserialization instead of just
-// body while gateway is supported, because gateway errors are returned
-// as http errors with json body.
+// Gateway errors were returned as http errors with json body.
+// Now that gateway is removed, this can be refactored to only use the response body instead of the whole HttpResponse.
 typealias HttpResponseDeserializer<T> = Function<HttpResponse, T>
```
DelevoXDG, line 9: I think we can remove this comment altogether and stick to the current approach. 

I'm pretty sure RPC nodes can fall prey to network errors which could result in HTTP error instead of JSON-RPC one. Changing this could hinder our ability to properly handle such situation.

Current usage of `HttpResponseDeserializer`:
```kotlin
@JvmSynthetic
internal fun <T> buildJsonHttpDeserializer(deserializationStrategy: KSerializer<T>): HttpResponseDeserializer<T> {
    return Function { response ->
        if (!response.isSuccessful) {
            throw RequestFailedException(
                payload = response.body,
            )
        }

        val jsonRpcResponse =
            Json.decodeFromString(
                JsonRpcResponse.serializer(deserializationStrategy),
                response.body,
            )

        if (jsonRpcResponse.error != null) {
            throw RpcRequestFailedException(
                code = jsonRpcResponse.error.code,
                message = jsonRpcResponse.error.message,
                revertError = jsonRpcResponse.error.data?.revertError,
                payload = response.body,
            )
        }

        if (jsonRpcResponse.result == null) {
            throw RequestFailedException(message = "Response did not contain a result", payload = response.body)
        }

        jsonRpcResponse.result
    }
}
```

.github/workflows/checks.yml
```diff
@@ -41,7 +39,7 @@ jobs:
           python-version: 3.9.12
           cache: 'pip'
 
-      - name: Install legacy devnet
+      - name: Install cairo-lang
```
THenry14, line 42: this is needed for v0 compilation, right? what do you think about removing this dependency, and storing compiled contracts in repo? we will probably drop cairo0 support soon-ish, and only focus on cairo1/2

lib/src/test/resources/compileContracts.sh
```diff
@@ -1,87 +1,14 @@
...
-}
-
-# TODO: (#345) Remove once legacy devnet is removed or CI migrated to macOS arm64 runnerecho "Fetching Cairo compilers to support legacy devnet..."
-pushd "$(dirname "$0")" || exit 1
-fetch_compilers "$V2_COMPILER_VERSION" "$V2_COMPILER_BIN_PATH"
-popd
-
 echo "Installing scarb..."
 asdf plugin add scarb || true
 echo "Done!"
```
THenry14, line 14: I'd remove this from this script; locally we should assume users already have this, and on CI we should use scarb action

.github/workflows/checks.yml
```diff
@@ -88,36 +86,10 @@ jobs:
...
-            popd
-          else
-            echo "Found existing binaries, skipping compilation."
-          fi
-
       - name: Run tests and generate coverage report
         env:
           DEVNET_PATH: ${{ github.workspace }}/starknet-devnet-rs/target/release/starknet-devnet
-          NETWORK_TEST_MODE: "non_gas"
+          NETWORK_TEST_MODE: "disabled"
```
THenry14, line 92: why disable? devnet started working?

lib/src/main/kotlin/com/swmansion/starknet/deployercontract/Deployer.kt
```diff
@@ -35,8 +36,43 @@ interface Deployer {
...
+     * @throws RequestFailedException
+     * @throws SaltGenerationFailedException
+     */
+    fun deployContract(classHash: Felt, constructorCalldata: Calldata, maxFee: Felt): Request<ContractDeployment> {
+        val random = SecureRandom()
+        val salt = random.longs(1, 1, Long.MAX_VALUE).findFirst().orElseThrow { SaltGenerationFailedException() }
+        val feltSalt = Felt(salt)
+        return deployContract(classHash, true, feltSalt, constructorCalldata, maxFee)
+    }
+
```
THenry14, line 75: let's keep implementations away from this file. btw, do we really need this one?

lib/src/main/kotlin/com/swmansion/starknet/deployercontract/Deployer.kt
```diff
@@ -35,8 +36,43 @@ interface Deployer {
...
+    /**
+     * Deploy a contract through Universal Deployer Contract (UDC) with random generated salt and
+     * unique parameter set to true
+     *
+     * @param classHash a class hash of the declared contract
+     * @param constructorCalldata constructor calldata
+     * @param maxFee maximum fee that account will use for the deployment
+     * @throws RequestFailedException
+     * @throws SaltGenerationFailedException
```
THenry14, line 67:
```suggestion
     * @param classHash a class hash of the declared contract
     * @param constructorCalldata constructor calldata
     * @param maxFee maximum fee that account will use for the deployment
     *
     * @throws RequestFailedException
     * @throws SaltGenerationFailedException
```

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -113,39 +116,95 @@ class JsonRpcProvider(
...
+     * Get a value of storage var.
+     *
+     * Get a value of a storage variable of contract at the provided address.
+     *
+     * @param contractAddress an address of the contract
+     * @param key an address of the storage variable inside contract
+     * @param blockTag The tag of the requested block.
+     *
+     * @throws RequestFailedException
+     */
```
THenry14, line 129: shouldn't this be in Provider.kt?

(applies to other doc-comments in this file as well)

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
THenry14: >I'm pretty sure RPC nodes can fall prey to network errors which could result in HTTP error instead of JSON-RPC one. Changing this could hinder our ability to properly handle such situation.

that's true. I think we can remove the comment altogether, as well.

THenry14 (commented):
The changes look good
I have only glanced through tests though, but I don't think then changes there were breaking in any way

Let's also make sure android and java demos are not using anything gateway related, and that they are still working.

overall great work 💪

lib/src/main/kotlin/com/swmansion/starknet/deployercontract/Deployer.kt
DelevoXDG: Just below, the same approach was used. 
But I agree we should keep implementations away from interfaces for consistency.

As for the reasons why we need deployContract with maxFee parameter, I tried to outline those in #361

THenry14: I meant this specific implementation - you have already added max_fee to deployContract at the top - why create new variation (with implied uniqueness and random salt)?

DelevoXDG: There were previously two implementations, one with all the all the parameters, and the other one without the parameters that can be assumed, so I thought it would be useful to allow passing maxFee in both cases.

I can imagine someone wanting to not bother with generating random salt while also passing maxFee at the same time.

THenry14: meh, maybe. that's a detail though, if you want to leave this that's fine

PR #366, opened 2023-12-04
Support any data in JsonRpcError

### LINKED ISSUE #362: Error response mismatch
### What happened

From tg:

For every rpc error response the library is throwing JsonDecodingException when using it with the new Pathfinder v0.9.5+.
kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 76: Encountered an unknown key 'error' at path: $.error.data
Use 'ignoreUnknownKeys = true' in 'Json {}' builder to ignore unknown keys.
JSON input: {\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32603,\"message\":\"Internal error\",\"data\":{\"error\":\"Invalid message selector\"}},\"id\":0}

It seems like the lib is not adapted for the new error model introduced with 0.5.1 rpc spec. Basically the rpc errors are now wrapped in the data object. Could you please adapt the lib to the new error model or at least make it more relaxed (and not throw) when there is an unexpected field in the json?

### Stack trace

.

### Steps to reproduce

1. 

### SDK Version

latest

### Language

Kotlin

### Language

n/a

### Language version

whatever we are using in repo

### What operating system are you using?

Linux, Mac

### Is there an existing issue for this?

- [X] I have searched the existing issues and verified no issue exits for this problem.

DelevoXDG:
## Describe your changes

Until RPC 0.7.0 comes out, with, hopefully, limitations on internal errors, should accept any data as an error.

- Separate `JsonRpcError` into `JsonRpcStandardError` and `JsonRpcContractError`
- Add tests

## Linked issues

Closes #362 

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcErrorPolymorphicSerializer.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcErrorPolymorphicSerializer.kt
@@ -1,0 +1,60 @@
+package com.swmansion.starknet.data.serializers
+
+import com.swmansion.starknet.provider.rpc.JsonRpcContractError
+import com.swmansion.starknet.provider.rpc.JsonRpcError
+import com.swmansion.starknet.provider.rpc.JsonRpcStandardError
+import kotlinx.serialization.DeserializationStrategy
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.*
+
+internal object JsonRpcErrorPolymorphicSerializer : JsonContentPolymorphicSerializer<JsonRpcError>(JsonRpcError::class) {
+    override fun selectDeserializer(element: JsonElement): DeserializationStrategy<JsonRpcError> {
+        val jsonElement = element.jsonObject
+
+        val isContractError = "data" in jsonElement &&
+            jsonElement["data"]!! is JsonObject &&
+            "revert_error" in jsonElement["data"]!!.jsonObject &&
+            jsonElement["data"]!!.jsonObject.size == 1
+        return when {
+            isContractError -> JsonRpcContractError.serializer()
+            else -> JsonRpcStandardError.serializer()
+        }
+    }
+}
+
+internal object JsonRpcStandardErrorSerializer : KSerializer<JsonRpcStandardError> {
+    override fun deserialize(decoder: Decoder): JsonRpcStandardError {
+        val input = decoder as? JsonDecoder ?: throw SerializationException("Expected JsonInput for ${decoder::class}")
+
+        val jsonObject = input.decodeJsonElement().jsonObject
+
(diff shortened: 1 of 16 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcErrorPolymorphicSerializer.kt
```diff
@@ -0,0 +1,60 @@
...
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.*
+
+internal object JsonRpcErrorPolymorphicSerializer : JsonContentPolymorphicSerializer<JsonRpcError>(JsonRpcError::class) {
+    override fun selectDeserializer(element: JsonElement): DeserializationStrategy<JsonRpcError> {
+        val jsonElement = element.jsonObject
+
+        val isContractError = "data" in jsonElement &&
```
DelevoXDG, line 20: Should base on code instead?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -19,26 +21,54 @@ private data class JsonRpcResponse<T>(
...
+    @SerialName("message")
+    override val message: String,
+
+    @SerialName("data")
+    override val data: String? = null,
+) : JsonRpcError()
+
 @Serializable
-private data class JsonRpcError(
+internal data class JsonRpcContractError(
```
DelevoXDG, line 52: Theoretically, it could be as well accepted as StandardError, but then error.toString() would look differently:
```
40 : "Contract error" : {"revert_error": "Revert error"}
```

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcErrorPolymorphicSerializer.kt
THenry14: if possible that would probably be a better approach

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcErrorPolymorphicSerializer.kt
```diff
@@ -0,0 +1,60 @@
+package com.swmansion.starknet.data.serializers
```
THenry14, line 1: I suppose the file name could not include JsonRpc?

THenry14 (approved):

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcErrorPolymorphicSerializer.kt
DelevoXDG: I think it should, since the naming convention is {class_name}Serializer, and we probably don't wanna rename `JsonRpcError` to just `Error`

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/JsonRpcErrorPolymorphicSerializer.kt
DelevoXDG: Discussed offline, decided to leave the current approach to avoid unnecessary errors associated with invalid contract errors returned by nodes.

Issue #384, opened 2023-12-28
Test v3 transactions on real Sepolia network


DelevoXDG:
### Feature Request

Once possible and needed, test support testing v3 transactions on Sepolia (network tests)

PR #378, opened 2023-12-22
Support Sepolia; Support `starknet_chainId`

DelevoXDG:
## Describe your changes
- Update `StarknetChainId`:
  - Add `StarknetChainId.SEPOLIA_TESTNET` and `SEPOLIA_INTEGRATION`  
  - Rename `StarknetChainId.TESTNET` -> `StarknetChainId.GOERLI`
- `Provider`, `StandardProvider`
  - Add `getChainId` (support `starknet_chainId`)
  - removed `chainId` property in favor of `getChainId` method
  - `StandardProvider` constructor no longer takes `chainId` as an argument
- Update relevant documentation and network test config
- Bump cairo-lang to 0.13.0  

Added basic devnet and network tests to test `starknet_chainId` and Sepolia chain ids. Due to current Sepolia network limitations and migration time considerations, I think support for Sepolia in other network tests should be addressed separately: #379.

## Linked issues

Closes #357 
Closes #376 

## Breaking changes

- [x] This issue contains breaking changes

- Renamed `StarknetChainId.TESTNET` -> `StarknetChainId.GOERLI`
- `Provider`: removed `chainId` property; Use `getChainId` method
- `StandardProvider` constructor no longer takes `chainId` as an argument

PR #386, opened 2023-12-29
Support sepolia network tests

### LINKED ISSUE #379: Support running network tests on Sepolia
### Feature Request

Follow up to #357 

We will eventually need to migrate to Sepolia in network tests, since Goerli is getting deprecated.

DelevoXDG:
## Describe your changes

Follow up to #378 

- Add `ACCOUNT_CAIRO_VERSION` to network test config. Allows tests to work properly with Cairo 1 accounts.
- Support running most tests on Sepolia testnet and integration. v3 tx tests will be addressed in the future once we get some STRK tokens: #384
- Add map contracts for reference
- Update README

## Linked issues

Closes #379 

## Breaking changes


lib/src/test/resources/contracts/src/map.cairo
DelevoXDG, line 1: The Cairo 1 and Cairo 0 map contract is not currently used anywhere, but I think we might wanna keep them for reference and/or if we ever want to declare them again.

lib/src/test/kotlin/network/provider/ProviderTest.kt
```diff
@@ -54,94 +54,40 @@ class ProviderTest {
...
-        }
+
         assertNotNull(transactionStatus2.executionStatus)
         assertEquals(TransactionExecutionStatus.REVERTED, transactionStatus2.executionStatus)
     }
 
-    @Disabled
-    @Test
-    fun `estimate message fee`() {
-        assumeTrue(NetworkConfig.isTestEnabled(requiresGas = false))
```
DelevoXDG, line 86: I decided there's no good reason to spend time updating this test, so I removed it altogether.
Network tests are meant for basic standard flow operations and new features. This test was created when there was no devnet with new rpc version, now this functionality is tested fine on devnet. If there are changes to it, it would be easier to write a new test case from scratch than redo this one.

THenry14 (commented):
What is the mid-to-long term plan for tests? It seems to me we could get rid of some of the cairo 0 tests, maybe even goeril tests, and a few things could be simplified. Also, it seems that some hashes (eg 0x1e04294ce4f7dd489ba5b5618dc112b37f9a7e82e2ded5691fb3083839dd3b5) are repeated in multiple places, so maybe they could be stored in a variable and reused that way, which would make maintaining the code a bit easier?

THenry14:
> Do you mean all tests or only the network ones? I think devnet seems to be more stable, so keep testing most of functionalities there, while network tests are reserved for standard flow functionality and new features.

All tests. 

> As for declaring cairo 0, I think we could say it's still core functionality, and for now there's almost no maintance required for those tests.

perhaps, I wonder if we could drop cairo 0 altogether though. I'm not sure it is core functionality - getting receipts, statuses etc - yes, but for example declaring - not so much. Let's make an issue and discuss this at later date.

> Mid-long term, I think we can stop testing new features on Goerli, and only use Sepolia. Eventually, we can remove Goerli tests altogether.

I think we can migrate everything to sepolia sooner. This is not an urgent thing though.

THenry14 (approved):

PR #405, opened 2024-02-05
Support ignoring unknown json keys in responses and contract codes

### LINKED ISSUE #399: Add `ingnoreUnknownKeys = true` for responses parsing
### Feature Request

from tg: 

would it make sense to use ignoreUnknownKeys = true for the parsing of the responses to prevent issues <with missing fields> in the future?

DelevoXDG:
## Describe your changes

- Support using `Json` instance with `ignoreUnknownKeys = true` on demand for deserialization of responses in `JsonRpcProvider` and Cairo 1 contract codes.
- Reuse `Json` instances where possible
- Rename  `HttpErrorHandlerTest` to `JsonRpcResponseTest`

## Linked issues

Closes #399 

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/Cairo1ClassHashCalculator.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/Cairo1ClassHashCalculator.kt
@@ -3,10 +3,14 @@ package com.swmansion.starknet.data
 import com.swmansion.starknet.crypto.Poseidon
 import com.swmansion.starknet.crypto.starknetKeccak
 import com.swmansion.starknet.data.types.*
+import kotlinx.serialization.json.Json
 
 object Cairo1ClassHashCalculator {
+    private val jsonWithIgnoreUnknownKeys by lazy { Json { ignoreUnknownKeys = true } }
+
     fun computeSierraClassHash(contract: Cairo1ContractDefinition): Felt {
-        val contractClass = contract.deserializationJson.decodeFromJsonElement(ContractClass.serializer(), contract.toJson())
+        val json = if (contract.ignoreUnknownJsonKeys) jsonWithIgnoreUnknownKeys else Json
+        val contractClass = json.decodeFromJsonElement(ContractClass.serializer(), contract.toJson())
 
         val sierraVersion = Felt.fromShortString("CONTRACT_CLASS_V" + contractClass.contractClassVersion)
         val externalEntryPointHash = Poseidon.poseidonHash(getSierraEntryPointsArray(contractClass.entryPointsByType.external))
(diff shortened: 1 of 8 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/Cairo1ClassHashCalculator.kt
```diff
@@ -6,9 +6,11 @@ import com.swmansion.starknet.data.types.*
 import kotlinx.serialization.json.Json
 
 object Cairo1ClassHashCalculator {
+    private val jsonWithIgnoreUnknownKeys by lazy { Json { ignoreUnknownKeys = true } }
 
     fun computeSierraClassHash(contract: Cairo1ContractDefinition): Felt {
-        val contractClass = Json.decodeFromJsonElement(ContractClass.serializer(), contract.toJson())
+        val json = if (contract.ignoreUnknownJsonKeys) jsonWithIgnoreUnknownKeys else Json
+        val contractClass = json.decodeFromJsonElement(ContractClass.serializer(), contract.toJson())
```
DelevoXDG, line 13: Do we want to additionally test this? We would have to probably keep altered compiled codes of Cairo 1 contracts, and we previously tried to avoid that.

THenry14: nah, should be fine without explicit test

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -27,13 +27,21 @@ import kotlinx.serialization.json.*
...
-    private val jsonWithDefaults = Json { encodeDefaults = true }
+    constructor(url: String, ignoreUnknownJsonKeys: Boolean) : this(url, OkHttpService(), ignoreUnknownJsonKeys)
+    constructor(url: String, httpService: HttpService) : this(url, httpService, false)
+    constructor(url: String) : this(url, OkHttpService(), false)
 
-    private val defaultFeeEstimateSimulationFlags: Set<SimulationFlagForEstimateFee> by lazy {
-        setOf(SimulationFlagForEstimateFee.SKIP_VALIDATE)
+    companion object {
+        private val jsonWithDefaults = Json { encodeDefaults = true }
+        private val jsonWithIgnoreUnknownKeys by lazy { Json { ignoreUnknownKeys = true } }
```
THenry14, line 40: shouldn't this be one val?

DelevoXDG: I'm not sure which one are you referring to, both of the vals in the snippet are vals 😄

THenry14: I refer to both. Should these be two separate json objects, or one json object? if the former, why is it so?

DelevoXDG: `jsonWithDefaults` is only used for encoding. 
`jsonWithIgnoreUnknownKeys` will be only used:
- for decoding responses
- when `ignoreUnknownJsonKeys` is true

Using one json object would cause the following problems:
- Inability to keep json instances static and available to all providers (creating new `Json` instances is a bit costly, so I think we should avoid it if possible)
- Use of `encodeDefaults` and `ignoreUnknownKeys` when these are not needed

THenry14 (approved):

PR #406, opened 2024-02-07
Use Felt instead of StarknetChainId

### LINKED ISSUE #404: Remove StarknetChainId
### Feature Request

There's no longer need to keep `StarknetChainId` enum. 

`Provider.getChainId` can just return `Felt` value instead.

We can assume users know which node they're using. Alternatively, users may use `Felt.toShortString` to verify the chain id of the node.

DelevoXDG:
## Describe your changes

- Use `Felt` instead of `StarknetChainId` in arguments, return values and fields
- Remove `StarknetChainId`

## Linked issues

Closes #404 

## Breaking changes

- Removed `StarknetChainId` enum; Use `Felt.toShortString` and `Felt.fromShortString` to convert to and from readable name
- `Provider.getChainId()` now returns `Felt` instead of `StarknetChainId`
- `TransactionHashCalculator` methods now accept `chainId` as `Felt`, not `StarknetChainId`
- `TransactionFactory`  methods now accept `chainId` as `Felt`, not `StarknetChainId`


lib/src/test/kotlin/starknet/data/types/TransactionsTest.kt
```diff
@@ -13,10 +13,12 @@ import java.math.BigInteger
 internal class TransactionsTest {
     @Test
     fun getHash() {
+        val chainId = Felt.fromShortString("SN_GOERLI")
```
DelevoXDG, line 16: This got me thinking a bit.

`TransactionFactory.makeXYZTransaction` is something we expect users to barely use, but it's still public. Naturally, users should use `Provider.getChainId()` before passing one to `make` methods, but I wonder if it would make sense to keep the enum available, but still accept `Felt` in all methods, e.g.: 
```
val tx1 = TransactionFactory.makeInvokeV1Transaction(
            senderAddress = Felt.fromHex("0x2a"),
            calldata = listOf(),
            chainId = StarknetChainId.SEPOLIA.feltValue,
            ...
```

That would still allow use of custom networks, but maybe(?) provide a slight accommodation to the newcomers. This will however introduce a slight inconsistency.

Is this a valid concern or should we just go ahead and remove the enum altogether?

THenry14: I would remove the enum, this is pretty easy to overcome. To make things even easier, we can perhaps add encoded network names somewhere in the docs? and of course, let's show how to use this in our demos - I think doing fromShortString there should be enough to point users in the correct direction. sounds good?

THenry14 (approved):

PR #409, opened 2024-02-09
Support RPC 0.7.0

### LINKED ISSUE #408: Support RPC 0.7.0
### Feature Request

The [rc0](https://github.com/starkware-libs/starknet-specs/releases/tag/v0.7.0-rc0) is now available

DelevoXDG:
## Describe your changes

- Fee-related changes
    - In `StandardAccount`, add `executeV1` and `executeV3` methods that allow passing overhead that should be applied to auto-estimated fee
  - Add `dataGasConsumed`, `dataGasPrice` to `EstimateFeeResponse`
  - Update fee-calculation `EstimateFeeResponse` methods
    - Change `maxAmount` calculation logic in `toResourceBounds` 
    - Use `0.5` overhead in `toResourceBounds` by default
    - Add docstrings for `toMaxFee` and `toResourceBounds`
- Block-related changes
  -  Add support for `starknet_getBlockWithReceipts` endpoint
     - Add `Provider.getBlockWithReceipts` methods
     - ~~Add `PendingDeployTransactionReceipt` as a temporary workaround **(pending final decision on the response structure in the upcoming updates to the spec)**~~
     - Add `BlockWithReceiptsResponse` and `PendingBlockWithReceiptsResponse`
     - #431 
     - #432 
  - Add new fields  (`l1DataGasPrice`, `l1DataAvailabilityMode`) to all `BlockWith[..]` dataclasses 
  - Refactor all `BlockWith[..]` dataclasses 
    - Add `GetBlockWithResponse`, `ProcessedBlock`, `PendingBlock` intermediate interfaces 
    - Remove `status` from `PendingBlockWith[..]` dataclasses
- Resources-related changes
  - Add `ComputationResources`
  - Remove `ExecutionResourcesSerializer` workaround serializer
  - Add `executionResources` to `InvokeTransactionTrace`, `RevertedInvokeTransactionTrace`, `DeclareTransactionTrace`, `DeployAccountTransactionTrace`
  - Replace `executionResources` with `computationResources` in `FunctionInvocation`
- Bump devnet
  - Use higher `maxFee` for declare methods by default in `DevnetClient`
- Bump version to **0.11.0**

## Linked issues

Closes #408 

## Breaking changes

  - Not compatible with JSON-RPC 0.6.0 spec
  - Removed `executionResources` from `FunctionInvocation`; Use `computationResources` instead
  - Removed `status` from `PendingBlockWith[..]` dataclasses


lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -161,15 +161,19 @@ data class SyncingResponse(
 ) : Syncing()
 
 @Serializable
-sealed class GetBlockWithTransactionsResponse {
-    abstract val transactions: List<Transaction>
+sealed class GetBlockWithResponse {
```
DelevoXDG, line 164: Should we rename it to `GetBlockWithResponseBase` or something else?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -240,21 +263,94 @@ data class PendingBlockWithTransactionsResponse(
 
     @SerialName("starknet_version")
     override val starknetVersion: String,
-) : GetBlockWithTransactionsResponse()
+) : GetBlockWithTransactionsResponse(), PendingBlock
+
+@Serializable
+sealed class GetBlockWithTransactionsReceiptsResponse : GetBlockWithResponse {
```
DelevoXDG, line 269: I think we should get the naming in line, cause these are barely readable. I suggest the `TransactionReceipt` approach:

`GetBlockWithTransaction[...]Response` -> `BlockWithTransaction[...]`
`BlockWithTransaction[...]Response` -> `ProcessedBlockWithTransaction[...]`
`PendingBlockWithTransaction[...]Response` -> `PendingBlockWithTransaction[...]`


where `[...] = [s/Hashes/Receipts]`

This is breaking, but I think it will make interaction with these easier in the future.

WDYT?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -160,20 +161,36 @@ data class SyncingResponse(
     override val highestBlockNumber: Int,
 ) : Syncing()
 
+sealed interface GetBlockWithResponse {
```
DelevoXDG, line 164: Also, we can probably move all of these to `lib/src/main/kotlin/com/swmansion/starknet/data/types/Block.kt`

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
```diff
@@ -319,6 +319,52 @@ interface Account {
...
+     * Execute a list of calls using version 3 invoke transaction with automatically estimated fee
+     * and specified overheads that will be added to the estimated fee when resource bounds are calculated.
+     *
+     * @param calls a list of calls to be executed.
+     * @param estimateAmountOverhead how big overhead should be added to the estimated amount.
+     * @param estimateUnitPriceOverhead how big overhead should be added to the estimated unit price.
+     *
+     * @return Invoke function response, containing transaction hash.
+     */
+    fun executeV3(calls: List<Call>, estimateAmountOverhead: Double, estimateUnitPriceOverhead: Double): Request<InvokeFunctionResponse>
```
DelevoXDG, line 343: Should we change the logic a bit and change `estimate[..]Overhead` to `estimate[..]Multiplier`, or rename them altogether? It might be more intuitive now that these methods are added.

WDYT?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -238,23 +292,119 @@ data class PendingBlockWithTransactionsResponse(
...
+    @SerialName("transaction")
+    val transaction: Transaction,
+
+    @Serializable(with = TransactionReceiptPolymorphicSerializer::class)
+    @SerialName("receipt")
+    val receipt: TransactionReceipt,
+)
+
+@Serializable
+sealed class GetBlockWithReceiptsResponse : GetBlockWithResponse {
```
DelevoXDG, line 317: I think we might want to improve naming for these, seems to me they are becoming less and less readable 😅 

I suggest the `TransactionReceipt` approach:

`GetBlockWith[..]Response` -> `BlockWith[..]`
`BlockWith[..]Response` -> `ProcessedBlockWith[..]`
`PendingBlockWith[..]Response` -> `PendingBlockWith[..]`


where `[...] = [Transactions/TransactionHashes/Receipts]`

This is breaking, but I think it will make interaction with these easier in the future.

WDYT?

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/GetBlockWithPolymorphicSerializer.kt
```diff
@@ -24,11 +18,22 @@ internal object GetBlockWithTransactionsPolymorphicSerializer : JsonContentPolym
...
+    override fun selectDeserializer(element: JsonElement): DeserializationStrategy<GetBlockWithReceiptsResponse> {
+        return when (isPendingBlock(element.jsonObject)) {
+            true -> PendingBlockWithReceiptsResponse.serializer()
+            false -> BlockWithReceiptsResponse.serializer()
+        }
+    }
+}
+
+private fun isPendingBlock(element: JsonObject): Boolean {
+    return listOf("block_hash", "block_number", "new_root").any { it !in element }
```
ddoktorski, line 38: Does it make sense to check that all these keys are missing, not only one of them?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
ddoktorski: Sounds good to me

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
```diff
@@ -320,6 +320,52 @@ interface Account {
...
+    /**
+     * Execute single call using version 1 invoke transaction with automatically estimated fee
+     * and specified overheads that will be added to the estimated fee when max fee is calculated.
+     *
+     * @param call a call to be executed.
+     * @param estimateFeeOverhead how big overhead should be added to the estimated fee.
+     *
+     * @return Invoke function response, containing transaction hash.
+     */
+    fun executeV1(call: Call, estimateFeeOverhead: Double): Request<InvokeFunctionResponse>
```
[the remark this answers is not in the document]
DelevoXDG: I understand the concern, however, we have docs for that, and the methods have different signatures, `maxFee` is passed as `Felt`, while overhead / multiplier is `Double`.

There's unfortunately no named arguments in Java, so we can only hope parameter name hints in IDEs and docs do their job, unless we want to create intermediate classes like `AutoEstimateParams`.

.github/workflows/checks.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     env:
-      DEVNET_SHA: 55191ee549b33ccbb0bc9d20dd929e39832a5ea5
+      DEVNET_SHA: 5eddfa62918aac86751ed93b09827e9b79ffb100
```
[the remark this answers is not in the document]
DelevoXDG: Incomplete [PRC 0.7.0](https://github.com/0xSpaceShard/starknet-devnet-rs/pull/364)

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/GetBlockWithPolymorphicSerializer.kt
DelevoXDG: This will return `true` if `any` of the keys is missing.

ddoktorski: Yup, so my question was that maybe we should return `true` iff all the keys are missing.

DelevoXDG: Why though?

If we actually consider the case when there is only one of the values missing in the node's response (which is not something we really should have to account for), and `ignoreUnknownKeys` is set to `true`, the exception will be avoided thanks to the current approach. If not set, deserialization will fail anyway. 

Assuming normal behavior, using `any` just slightly reduces the number of checks we make here 😅

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -160,20 +192,33 @@ data class SyncingResponse(
...
+    val timestamp: Int
+    val sequencerAddress: Felt
+    val parentHash: Felt
+    val l1GasPrice: ResourcePrice
+    val l1DataGasPrice: ResourcePrice
+    val l1DataAvailabilityMode: L1DAMode
+    val starknetVersion: String
+}
+
+sealed interface ProcessedBlock : GetBlockWithResponse {
```
THenry14, line 205: nit: let's add a comment that this is in fact a `BlockHeader`, or change the name altogether. 
The same goes for PendingBlock

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
THenry14: Ideally we should just have the same names as in spec.

That being said, I agree that having a `Response` in the class name is abit too verbose, since it's already defined in Responses.kt, but I do wonder if this change won't be too troublesome to users - now, if I'm not mistaken, there will be (eg) GetBlockWithTransactions classes defined in both Responses.kt and Payloads.kt, right? (just to be consistent). Does that mean, that when we want to import both classes, we must import them with an alias? Doesn't that defeat the idea of getting rid of verbose namings?

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Resources.kt
```diff
@@ -0,0 +1,98 @@
...
+    abstract val ecdsaApplications: Int?
+    abstract val bitwiseApplications: Int?
+    abstract val keccakApplications: Int?
+    abstract val segmentArenaApplications: Int?
+}
+
+@Serializable
+data class ComputationResources(
+    @SerialName("steps")
+    override val steps: Int,
```
THenry14, line 25: technically this cannot be 0. do you think we should make a serializer that ensures that, or do we yolo it and can just depend on the node's response?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: > if I'm not mistaken, there will be (eg) GetBlockWithTransactions classes defined in both Responses.kt and Payloads.kt, right?

No, in `Payloads.kt` the classes are named `GetBlockWith[..]Payload`, so we shouldn't expect any collisions.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: Makes sense, do you think it should be docstring or a comment for internal use? 

Just a little bit unsure about `BlockHeader` name, seems like this is valid way to define it in the spec, but is a bit less appropriate for an interface name, as we are not dealing with responses per se (where having "header" makes sense), but with objects that are converted from those responses

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Resources.kt
DelevoXDG: I thought about that and it seems to me yolo is the best approach right now 😅 

We already go in the direction of putting less restrictions than in the spec: `Felt` instead of `ContractAddress`, non-optional fields because of the nodes' deviations, allowing `ingoreUnknownKeys` etc. 

I suppose we can put faith into nodes being correct, and if they aren't, at least entire operation won't fail because of an incorrect zero.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
THenry14: That is my point - if we want to be consistent, shouldn't we rename these as well?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
THenry14: This is an interface, and it's not unreasonable to expect at least some users to have a look at the spec first, so perhaps this should be in the docstring?

DelevoXDG: Yeah, I agree, but the problem is that `ProcessedBlock` does not fully correspond to `BLOCK_HEADER`, as there's a `status` property that is not in the header, but is always included alongside it. 

Ofc I can mention this in the dosctring. Just not sure dosctring that contains

```
Corresponds to `BLOCK_HEADER` with additional `status` property"
```
is really helpful 😅

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: No, not really.

We already use the naming I suggest now in several places, e.g. 
- `GetTransactionStatusPayload` for payload and `TransactionStatus` for response
- `GetTransactionReceiptPayload` for payload and `TransactionReceipt` for response

THenry14: wait, let's take a step back. 

- If I understand correctly, you propose to change the naming for Responses classes - basically ditch the `Response` from the name
- I then ask if we want to be consistent, and drop verbose naming from other classes as well (Payloads); if yes, wouldn't this lead to classes named the same for Payloads and Responses, which would then mean we must import them with aliases
- I get that right now we have *Payload classes for payloads, but by consistent I mean to the Responses approach - why name something a *Payload, if it already is in a Payload.kt, and we do not name classes *Response if they reside in Responses.kt. this seems incosistent.

so my question is - if we want to do it, shoulnd't we do it for Payloads as well? What's with readability then?

DelevoXDG: > If I understand correctly, you propose to change the naming for Responses classes - basically ditch the Response from the name

Not really all of them. I suppose we could also rename `EstimateFeeResponse` -> `EstimateFee` or `FeeEstimate`, but leave classes like `DeployAccountResponse` as is. With logic being that we drop `Response` only if it's not necessary to understand the purpose of the class.

But there's no reason to introduce that many breaking changes at once. Since the spec already introduces changes to the blocks, I thought it would be reasonable to apply changes to them right now, as these are the most problematic with their verbose names. Rest can be done in a separate PR, unless you think we should introduce that change at once.

> if yes, wouldn't this lead to classes named the same for Payloads and Responses, which would then mean we must import them with aliases

This shouldn't really matter, as we use all of the `Payload` classes, at least those located in `Payloads.kt`, only internally. I'm not sure why aren't they `internal`?

> I then ask if we want to be consistent, and drop verbose naming from other classes as well (Payloads);

> I get that right now we have *Payload classes for payloads, but by consistent I mean to the Responses approach - why name something a *Payload, if it already is in a Payload.kt,

I think it introduces unnecessary collision and makes us rely on the source of the import instead of just the name. Besides, the name without `Payload` postfix just doesn't work for `Payload.kt` classes as it doesn't reflect what they are:
- `TransactionReceipt` is a good name for object converted from the response, we don't really need additional `Response` postfix. `TransactionReceipt`, `GetTransactionReceipt`, however, are not good names for the payload with tx hash as only field
- Similarly, `BlockWith[..]` is good name for response, while `BlockWith[..]` as a name for a payload does not reflect what the class is

My intuition was to drop postfix when it's not necessary, not drop it everywhere because the postfix is in the file name, so I don't think we should do it for `Payloads` as it only harms readability.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: Continued there: https://github.com/software-mansion/starknet-jvm/pull/409#discussion_r1507605842

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/GetBlockWithPolymorphicSerializer.kt
ddoktorski: I overlooked the fact that deserialization will only work if all keys are missing (without using `ignoreUnknownKeys`). Given this behavior, I am fine with the current approach.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
THenry14: > Rest can be done in a separate PR, unless you think we should introduce that change at once.

Yes, let's do just that to avoid confusion etc. This could be a part of some minor, post 0.7.0 refactor.

> This shouldn't really matter, as we use all of the Payload classes, at least those located in Payloads.kt, only internally. I'm not sure why aren't they internal?

I think they were once public, but am not sure. if that's the case, I'm fine with making them internal and naming them differently to responses.

> I think it introduces unnecessary collision and makes us rely on the source of the import instead of just the name. 

yup, that's my point. let's make sure all payloads are internal, and if they are we can leave them as they are as this won't lead to interface inconsistencies. Otherwise, we should think about a correct approach. sounds good?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
THenry14: wait, so the nodes return some additional field that is not in spec? 
Just something like "The class reflects a BLOCK_HEADER object in starknet-spec, with an additional status field as returned by some of the nodes" ? Although just having an information that this is indeed a BLOCK_HEADER object in spec shoudl be sufficient enough imo

DelevoXDG: > wait, so the nodes return some additional field that is not in spec?

No, it is in the spec, just that it's returned not as a part of `BLOCK_HEADER`, but alongside it, i.e.
```
"allOf": [
    {
        "title": "Block status",
        "type": "object",
        "properties": {
            "status": {
                "title": "Status",
                "$ref": "#/components/schemas/BLOCK_STATUS"
            }
        },
        "required": [
            "status"
        ]
    },
    {
        "title": "Block header",
        "$ref": "#/components/schemas/BLOCK_HEADER"
    },
    {
        "title": "Block body with transaction hashes",
        "$ref": "#/components/schemas/BLOCK_BODY_WITH_TX_HASHES"
    }
]
```

> Although just having an information that this is indeed a BLOCK_HEADER object in spec should be sufficient enough imo

I suppose we can avoid minor inconsistency here

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: >  let's make sure all payloads are internal

Should we make the change as a part of this PR or separately?

THenry14: as you prefer

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
THenry14: wait, so this is not a BLOCK_HEADER, but BLOCK_WITH_TX_HASHES right?

DelevoXDG: If you're referring to the interface in the sdk, you could say it's `BLOCK_WITH_TX_HASHES` without `BLOCK_BODY_WITH_TX_HASHES` or `BLOCK_HEADER` + `BLOCK_STATUS`

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: #430

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
THenry14: ok, so what is the final plan?

DelevoXDG: I went ahead and added reference to `BLOCK_HEADER` to the docstring. I don't think an extra status property does any harm. I'm not sure we necessary have to document it, as it seems to me it might only add more confusion.

THenry14 (approved):

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: #432

PR #413, opened 2024-02-09
Add `TransactionVersion` enum

### LINKED ISSUE #412: Use enums instead of Felt for tx versions
### Feature Request

Starting with RPC 0.7.0, all transaction versions are represented as enum in the spec. Considering that, it might be reasonable to use enums in all transaction classes on our side.

DelevoXDG:
## Describe your changes

Related to #409

- Add `TransactionVersion` enum
- Use it instead of `Felt` in tx, tx payload dataclasses and related functions

## Linked issues

Closes #412 

## Breaking changes

- `version` argument in `TransactionFactory` functions is now of type `TransactionVersion` instead of `Felt`
- `version` argument in `TransactionHashCalculator` functions is now of type `TransactionVersion` instead of `Felt`
- `TransactionPayload.version` is now of type `TransactionVersion` instead of `Felt`

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/account/StandardAccount.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/account/StandardAccount.kt
@@ -48,10 +48,7 @@ class StandardAccount(
 
     override fun signV1(calls: List<Call>, params: ExecutionParams, forFeeEstimate: Boolean): InvokeTransactionV1Payload {
         val calldata = AccountCalldataTransformer.callsToExecuteCalldata(calls, cairoVersion)
-        val signVersion = when (forFeeEstimate) {
-            true -> estimateVersion(Felt.ONE)
-            false -> Felt.ONE
-        }
+        val signVersion = if (forFeeEstimate) TransactionVersion.V1_QUERY else TransactionVersion.V1
         val tx = TransactionFactory.makeInvokeV1Transaction(
             senderAddress = address,
             calldata = calldata,
@@ -68,10 +65,7 @@ class StandardAccount(
 
     override fun signV3(calls: List<Call>, params: InvokeParamsV3, forFeeEstimate: Boolean): InvokeTransactionV3Payload {
         val calldata = AccountCalldataTransformer.callsToExecuteCalldata(calls, cairoVersion)
-        val signVersion = when (forFeeEstimate) {
-            true -> estimateVersion(Felt(3))
-            false -> Felt(3)
-        }
+        val signVersion = if (forFeeEstimate) TransactionVersion.V3_QUERY else TransactionVersion.V3
         val tx = TransactionFactory.makeInvokeV3Transaction(
             senderAddress = address,
             calldata = calldata,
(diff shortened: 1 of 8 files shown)


lib/src/main/kotlin/com/swmansion/starknet/account/StandardAccount.kt
```diff
@@ -48,10 +48,7 @@ class StandardAccount(
 
     override fun signV1(calls: List<Call>, params: ExecutionParams, forFeeEstimate: Boolean): InvokeTransactionV1Payload {
         val calldata = AccountCalldataTransformer.callsToExecuteCalldata(calls, cairoVersion)
-        val signVersion = when (forFeeEstimate) {
-            true -> estimateVersion(Felt.ONE)
-            false -> Felt.ONE
-        }
+        val signVersion = if (forFeeEstimate) TransactionVersion.V1_QUERY else TransactionVersion.V1
```
THenry14, line 51: Maybe we could change the name `v1_query` to something more meaningful? I guess this version can be used only for estimates, right? so. perhaps something connected to it? (unless this is how it's called in spec or something, then 🤷 )

DelevoXDG: It's not called that way in the spec, but it is always referred to as "query version". We can of course rename it to `V1_FOR_FEE_ESTIMATE`, but it might be a bit misleading since it's also used in `simulateTransactions`.

Maybe we should instead write a description of the enum for the docs, explaining the purpose of query versions?

THenry14: > Maybe we should instead write a description of the enum for the docs, explaining the purpose of query versions?

yes, let's

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
```diff
@@ -28,6 +29,40 @@ enum class TransactionType(val txPrefix: Felt) {
...
 
+/**
+ * The version of the transaction.
+ *
+ * The version is used to determine the transaction format and is used for transaction signing.
+ *
+ * Standard versions include [V0], [V1], [V2], and [V3]. These are utilized for regular transaction execution.
+ * Query versions [V1_QUERY], [V2_QUERY], and [V3_QUERY] should only be used for creating transactions to be used
+ * in queries that do not alter the chain state, as in methods like [Provider.simulateTransactions] and [Provider.getEstimateFee].
+ * Avoid using query versions for transactions intended for execution.
```
THenry14, line 40: > Avoid using query versions for transactions intended for execution.

I'm pretty sure this should not even be possible, and such tx would fail

THenry14 (approved):

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/Transaction.kt
DelevoXDG: It's not, hence the warning 😅

THenry14: But the warning says "Avoid using query versions for transactions intended for execution.", which implies it is possible :P

DelevoXDG: Which should convey to the user "it's possible from the SDK-side", but should not be done to .

DelevoXDG: Which should convey to the user "it's possible from the SDK-side, but should not be done".
Should I rephrase it to:

```
Sending transaction with a query version for execution will result in a failure.
```

THenry14: Yes, explicit is better than implicit :+1:

THenry14 (approved):

PR #417, opened 2024-02-13
Support custom chain IDs

### LINKED ISSUE #415: Allow custom chain IDs in `StarknetChainId`
### Feature Request

self-explanatory

DelevoXDG:
## Describe your changes

- Convert `StarknetChainId` from enum to data class
  - Support creating `StarknetChainId` from `Felt`
  - Add helper methods `fromHex`, `fromNetworkName`
- Rename `StarknetChainId.MAINNET` -> `StarknetChainId.MAIN` for consistency with naming convention for other chain ids.

## Linked issues

Closes #415 

## Breaking changes

- `StarknetChainId` is no longer an enum; it will not have enum specific methods
- Renamed `StarknetChainId.MAINNET` -> `StarknetChainId.MAIN`

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetChainId.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetChainId.kt
@@ -16,8 +16,7 @@ data class StarknetChainId(val value: Felt) {
         val SEPOLIA = StarknetChainId(Felt.fromHex("0x534e5f5345504f4c4941")) // encodeShortString('SN_SEPOLIA'),
 
         @field:JvmField
-        val INTEGRATION_SEPOLIA =
-            StarknetChainId(Felt.fromHex("0x534e5f494e544547524154494f4e5f5345504f4c4941")) // encodeShortString('SN_INTEGRATION_SEPOLIA'),
+        val INTEGRATION_SEPOLIA = StarknetChainId(Felt.fromHex("0x534e5f494e544547524154494f4e5f5345504f4c4941")) // encodeShortString('SN_INTEGRATION_SEPOLIA'),
 
         @JvmStatic
         fun fromNetworkName(shortString: String): StarknetChainId {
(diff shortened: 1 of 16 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetChainId.kt
DelevoXDG, line 1: Alternative approach can be checked out at 5d080307b2bf9319e6e129eb6b2f7ada6a94cd75.
For now, I think data class approach is better, since it provides all the useful methods. I was particularly interested 
in `.equals()` and `.hashCode() as they ensure equality between predefined objects and manually defined objects with the same values as predefined ones without the need to use any extra tricks.

lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetChainId.kt
```diff
@@ -1,20 +1,31 @@
...
+
+        @JvmStatic
+        fun fromNetworkName(shortString: String): StarknetChainId {
+            return StarknetChainId(Felt.fromShortString(shortString))
+        }
+
+        @JvmStatic
+        fun fromHex(hex: String): StarknetChainId {
+            return StarknetChainId(Felt.fromHex(hex))
+        }
```
DelevoXDG, line 29: Should we keep those? Hard for me to definitively say whether those are useful or redundant.

THenry14: I think they may come in handy, I'd leave those

THenry14 (approved):
💪

lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetChainId.kt
DelevoXDG: Added just one more (`toNetworkName`) for consistency

PR #420, opened 2024-02-14
Update `StarknetChainId` docs

DelevoXDG:
## Describe your changes

Related to #417 

## Linked issues

Closes

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetChainId.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetChainId.kt
@@ -18,16 +18,33 @@ data class StarknetChainId(val value: Felt) {
         @field:JvmField
         val INTEGRATION_SEPOLIA = StarknetChainId(Felt.fromHex("0x534e5f494e544547524154494f4e5f5345504f4c4941")) // encodeShortString('SN_INTEGRATION_SEPOLIA'),
 
+        /**
+         * Create [StarknetChainId] from a network name represented as ASCII string.
+         *
+         * @param networkName string transformed to [StarknetChainId]
+         */
         @JvmStatic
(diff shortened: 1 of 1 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetChainId.kt
```diff
@@ -18,16 +18,33 @@ data class StarknetChainId(val value: Felt) {
         @field:JvmField
         val INTEGRATION_SEPOLIA = StarknetChainId(Felt.fromHex("0x534e5f494e544547524154494f4e5f5345504f4c4941")) // encodeShortString('SN_INTEGRATION_SEPOLIA'),
 
+        /**
+         * Create [StarknetChainId] from a network name represented as ASCII string.
+         *
+         * @param networkName string transformed to [StarknetChainId]
```
THenry14, line 24: we generally did not informa about return value in other docstrings, right? (provider springs to my mind)
should we start doing so now?

THenry14 (approved):
my only comment is a nitpick - please address it before merge. otherwise, lgtm

lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetChainId.kt
DelevoXDG: We normally do, but sometimes we don't (in `Felt` for example).

I just assumed 
```
* @return [StarknetChainId]
```
is redundant in the context.

THenry14: I stand corrected, indeed we have this for signer as well. We might want to think about adding the information to provider (and other relevant places) as well. thanks for clarification, this can be resolved.

THenry14 (approved):

PR #428, opened 2024-03-02
Update TypedData in line with SNIP-12

### LINKED ISSUE #423: Update `TypedData` in line with `SNIP-12`
### Feature Request

We should make the necessary version to support 1st revision of the spec.

Related info:
- [SNIP-12](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md)
- [Additional clarifications](https://github.com/starknet-io/SNIPs/pull/77)
- starknet.js [implementation](https://github.com/starknet-io/starknet.js/pull/920)

DelevoXDG:
## Describe your changes

This PR is part of the stack:

-- Update TypedData in line with SNIP-12 (#428) 
-- Use sealed classes for basic and preset types (#440)
-- Add separate methods for primitives (#442)
-- Nest TypedData-related test cases (#434)

### TypedData Revision 0 and Revision 1
- Support `bool` basic type
- Introduce `Domain` class; Use it for `domain` instead of `JsonObject` 
- Introduce `Revision` enum
- Remove `raw` basic type; Don't allow `merkletree` with `raw` leaves type
- Rename type-related classes
  - `TypeBase`->`Type`
  - `Type`->`StandardType`

### TypedData Revision 1
- Support new types
  - Basic types: `enum`, `string` (long string), `u128`, `i128`, `ContractAddress`, `ClassHash`, `timestamp`, `shortstring` 
    - Add `EnumType`
  - Preset types: `TokenAddress`, `NtfId`, `u256`
- ~~Rename `types`->`customTypes`; Add `types` to handle `customTypes` + `presetTypes`~~ Add `allTypes` to handle `types` + `presetTypes`
- Update type verification in line with [these suggestions](https://github.com/starknet-io/SNIPs/pull/77)
- Update internal logic in line with revision 1 standards

### Other
- Expand `HashMethod` enum to hash multiple elements

## Linked issues

Closes #423 

## Breaking changes

- `TypedData.domain` is now of type `TypedData.Domain` instead of `JsonObject`
- Renamed `TypedData.Type`->`TypedData.StandardType`, `TypedData.TypeBase`->`TypedData.Type`
- Removed `raw` basic type

## References
(might come in handy for reviews)

- [SNIP-12](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md)
- [Additional clarifications re. types](https://github.com/starknet-io/SNIPs/pull/77)
- starknet.js [implementation](https://github.com/starknet-io/starknet.js/pull/920)
- [Serialization of Cairo types](https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/serialization_of_Cairo_types)


lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
```diff
@@ -80,48 +85,135 @@ import kotlinx.serialization.json.*
...
+                is EnumType -> extractEnumTypes(it.type) + it.contains
+                is MerkleTreeType -> listOf(it.contains)
+                is StandardType -> listOf(stripPointer(it.type))
+            }
+        }.distinct() + domain.separatorName + primaryType
+
+        customTypes.keys.forEach {
+            require(it.isNotEmpty()) { "Types cannot be empty." }
+            require(!it.isArray()) { "Types cannot end in *. [$it] was found." }
+            require(!it.startsWith("(") && !it.endsWith(")")) { "Types cannot be enclosed in parentheses. [$it] was found." }
```
DelevoXDG, line 150: https://github.com/starknet-io/SNIPs/pull/77/files

Not sure if this should be interpreted as "cannot be enclosed in ()" or should also apply to a singular parenthesis

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
```diff
@@ -136,20 +236,32 @@ data class TypedData private constructor(
...
+            params.forEach { param ->
+                val extractedTypes = when {
+                    param is EnumType -> {
+                        require(revision == Revision.V1) { "'enum' basic type is not supported in revision ${revision.value}." }
+                        listOf(param.contains)
+                    }
+                    param.type.isEnum() -> {
+                        require(revision == Revision.V1) { "Enum types are not supported in revision ${revision.value}." }
+                        extractEnumTypes(param.type)
+                    }
```
DelevoXDG, line 248: Should we have these just in case, or would it be enough for us to check `verifyTypes()`

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
DelevoXDG, line 1: I also considered to make `TypedData` a sealed class, where there would be unified logic, while parts that are revision-specific:
- `basicTypes`, `presetTypes`
- `verifyTypes()`
- `encodeValue()`
- `getDependencies()`
- `encodeDependency()`

would be placed in their respective data classes. This would reduce `require()` checks based on revision, remove revision-specific variables and functions, and probably improve overall readability

This however has some caveats, not necessary problems, but something to take into account:
- There could be slight code duplication re. similar, but diferent logic between rev 0 and 1
- Would not be able to directly instantiate `TypedData`, only `TypedDataRevX`/`TypedDataVX`; Could use factory method though
- We previously more or less followed sn.js approach, that would move us away from it

This of course something that would likely involve moving a lot of things around, so I decided to not include this in the current PR. I wonder if this is something we want to change before releasing the SNIP support, or leave for later / leave as is for now.

WDYT?

THenry14: Imo this is a nice to have, but we probably should not focus on this atm. We can create an issue with low prio for this though

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
```diff
@@ -80,52 +84,142 @@ import kotlinx.serialization.json.*
 @Suppress("DataClassPrivateConstructor")
 @Serializable
 data class TypedData private constructor(
-    val types: Map<String, List<TypeBase>>,
+    @SerialName("types")
+    val customTypes: Map<String, List<Type>>,
+
```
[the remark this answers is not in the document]
THenry14: I don't have a strong opinion about this, so whatever you think is best. I generally like the idea of keeping the naming as close to json as possible though. it may have been a bit less descriptive at first, but it follows closer the json which is a big plus imo.

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
THenry14: I understand it as inside `()`

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
```diff
@@ -81,52 +84,143 @@ import kotlinx.serialization.json.*
...
-        types = types,
+        customTypes = customTypes,
         primaryType = primaryType,
-        domain = Json.parseToJsonElement(domain).jsonObject,
+        domain = Json.decodeFromString(domain),
         message = Json.parseToJsonElement(message).jsonObject,
     )
 
+    @Transient
+    private val revision = domain.revision ?: Revision.V0
```
THenry14, line 109: what are the reasons to have v0 as default?

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
THenry14: I'd leave them

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
```diff
@@ -159,34 +259,62 @@ data class TypedData private constructor(
...
+
+            return try {
+                Felt.fromHex(primitive.content)
+            } catch (e: Exception) {
+                Felt.fromShortString(primitive.content)
             }
         }
+
+        throw IllegalArgumentException("Unsupported primitive type: $primitive")
+    }
```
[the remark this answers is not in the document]
THenry14: it would be a very good idea to have multiple methods per type indeed

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
```diff
@@ -276,29 +438,81 @@ data class TypedData private constructor(
...
+                ),
+                "TokenAmount" to listOf(
+                    StandardType("token_address", "ContractAddress"),
+                    StandardType("amount", "u256"),
+                ),
+                "NftId" to listOf(
+                    StandardType("collection_address", "ContractAddress"),
+                    StandardType("token_id", "u256"),
+                ),
+            )
```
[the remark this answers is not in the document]
THenry14: I was thinking the same tbh, we could really use those. Can be done in this PR but only if made in one-two commits (to make reviewing easier). otherwise, different PR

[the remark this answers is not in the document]
DelevoXDG: #440

THenry14 (approved):
💪

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
DelevoXDG: #441

PR #431, opened 2024-03-04
Refactor receipts and block with receipts

DelevoXDG:
## Describe your changes

Simplify receipts classes to adjust for receipt and `starknet_getBlockWithReceipts` changes in RPC 0.7.0 spec

- Fix `BlockWithReceipts.transactionWithReceipts` ->  `BlockWithReceipts.transactionsWithReceipts`
- Remove `ProcessedTransactionReceipt`, `PendingTransactionReceipt`
- Merge `Pending[..]TransactionReceipt` and `ProcessedTransactionReceipt[..]TransactionReceipt` into `[..]TransactionReceipt` classes
- Push `blockHash` and `blockNumber` as optional fields to `TransactionReceipt`
- Add `hasBlockInfo` and `isPending` helper properties
- Add `isPending` helper method in `Block`

## Linked issues

Closes

## Breaking changes

- Remove `ProcessedTransactionReceipt`, `PendingTransactionReceipt`
- Rename `ProcessedInvokeTransactionReceipt` -> `InvokeTransactionReceipt`, rename `ProcessedDeployAccountTransactionReceipt` -> `DeployAccountTransactionReceipt`, rename `ProcessedDeclareTransactionReceipt` -> `DeclareTransactionReceipt`, rename `ProcessedL1HandlerReceipt` -> `L1HandlerTransactionReceipt`, rename `ProcessedTransactionDeployReceipt` -> `TransactionDeployReceipt`
- Remove `PendingInvokeTransactionReceipt`, `PendingDeployAccountTransactionReceipt`, `PendingDeclareTransactionReceipt`, `PendingL1HandlerTransactionReceipt`, `PendingDeployTransactionReceipt`
- `blockHash` and `blockNumber` fields are now optional in all `TransactionReceipt` classes


lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -70,29 +71,38 @@ sealed class TransactionReceipt {
...
+    /**
+     * Checks if the transaction is pending.
+     *
+     * Returns correct result only for receipts obtained using [Provider.getTransactionReceipt].
+     * For receipts obtained using [Provider.getBlockWithReceipts], use [BlockWithReceipts.isPending].
+     *
+     * @return true if the transaction is pending, false otherwise.
+     */
+    val isPending: Boolean
+        get() = !hasBlockInfo && finalityStatus == TransactionFinalityStatus.ACCEPTED_ON_L2
```
DelevoXDG, line 101: Considering Argent is interacting with receipts from top-level, and other users probably do to, I assumed we shouldn't fully eliminate `isPending` property. 

It, however, is problematic in the 0.7 spec, as the receipt in `ProcessedBlockWithReceipts` (`BLOCK_BODY_WITH_RECEIPTS`) can satisfy all the conditions.

My suggestions are:
1. Add `blockHash` and `blockNumber` from `ProcessedBlockWithReceipts` to each of the receipts. That will however involve costly `copy` operations. It can be checkout at a471b657eac9c516bb831c621dd098769d232b03.
2. Current approach:
  - Add `Block.isPending` method
  - Add info to the docs that `TransactionReceipt.isPending` shouldn't really be used when the receipt is obtained from `BlockWithReceipts.isPending`

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
```diff
@@ -70,29 +71,39 @@ sealed class TransactionReceipt {
...
+    /**
+     * Checks if the transaction is in a pending block.
+     *
+     * This method provides accurate results for receipts obtained through [Provider.getTransactionReceipt].
+     * For receipts acquired using [Provider.getBlockWithReceipts], refer to [BlockWithReceipts.isPending].
+     *
+     * @return `true` if the transaction is pending; `false` otherwise.
+     */
+    val isPending: Boolean
+        get() = !hasBlockInfo && finalityStatus == TransactionFinalityStatus.ACCEPTED_ON_L2
```
THenry14, line 92: what if we get accepted_on_l1 ?

DelevoXDG: It will not be a considered a pending receipt, as per the spec pending receipt should always have `ACCEPTED_ON_L2` finality status

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -322,7 +328,7 @@ data class ProcessedBlockWithReceipts(
     override val status: BlockStatus,
 
     @SerialName("transactions")
-    override val transactionWithReceipts: List<TransactionWithReceipt>,
+    override val transactionsWithReceipts: List<TransactionWithReceipt>,
```
ddoktorski, line 331: nit: I see the point with differentiating between a list of transactions and transactions with receipts, but it might be easier for everyone if we just stick with calling it `transactions`.

ddoktorski (approved):
LGTM

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
ddoktorski: I've already approved, but on second thoughts, the current approach is indeed not ideal and we should think about something better. Do we have any rough estimates of the impact of the first approach?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
DelevoXDG: It seems more descriptive and we also we have `transactionHashes` field just below. I don't have a strong say on this though, @THenry14 WDYT?

THenry14: these are different types basically, transactions (hashes) and receipts. do not have a strong opinion as well, but I'd lean towards leaving this as is? it is indeed more descriptive

lib/src/main/kotlin/com/swmansion/starknet/data/types/transactions/TransactionReceipt.kt
DelevoXDG: There are also discussions to add `block_number` to pending receipt. Then, we could potentially make `isPending` throw error or return `null` if it's obtained using `Provider.getBlockWithReceipts`. I think this is all however a topic for a separate discussion when the decision is made and the new version of the spec is released. As for now I suggest we pick one of the currently available solutions to not further delay the release

> Do we have any rough estimates of the impact of the first approach?

I further read about `.copy()` and it seems like it's not as costly as I thought. Do you think we should go for it instead?

ddoktorski: If there's no premise to suggest a significant performance impact, I believe we should go with the first approach for now. Though it's a bit of a hack, it assures a consistent user experience, ensuring identical behavior regardless of the method used to get a transaction receipt.

PR #437, opened 2024-03-13
Update MerkleTree to allow Poseidon hashing

DelevoXDG:
## Describe your changes

Related to and is needed for #428 
- Add `HashMethod` enum
- Add optional `hashFunction` parameter to `MerkleTree` constructor
- Add mandatory `hashFunction` parameter to `MerkleTree.hash()` method

## Linked issues

Closes

## Breaking changes

- `MerkleTree.hash()` static method now requires a `hashFunction` argument to be passed


lib/src/main/kotlin/com/swmansion/starknet/crypto/HashMethod.kt
```diff
@@ -0,0 +1,18 @@
+package com.swmansion.starknet.crypto
+
+import com.swmansion.starknet.data.types.Felt
+
+enum class HashMethod {
```
DelevoXDG, line 5: I'm not sure whether `HashMethod` or `HashFunction` is a more suitable name for this enum

THenry14: Maybe `HashType` instead? method/function kind of indicates we are dealking with a function, and not an enum

lib/src/main/kotlin/com/swmansion/starknet/crypto/HashMethod.kt
```diff
@@ -0,0 +1,18 @@
...
+    PEDERSEN {
+        override fun hash(first: Felt, second: Felt): Felt {
+            return StarknetCurve.pedersen(first, second)
+        }
+    },
+    POSEIDON {
+        override fun hash(first: Felt, second: Felt): Felt {
```
THenry14, line 12: Won't we need it to be able to calculate hashes on bigger number of arguments? (pedersen as well. this is probably a nitpick, since this is more of a future-problem, probably)

lib/src/main/kotlin/com/swmansion/starknet/data/types/MerkleTree.kt
```diff
@@ -12,9 +13,9 @@ data class MerkleTree(
 
     companion object {
         @JvmStatic
-        fun hash(a: Felt, b: Felt): Felt {
+        fun hash(a: Felt, b: Felt, hashFunction: HashMethod): Felt {
             val (aSorted, bSorted) = if (a < b) Pair(a, b) else Pair(b, a)
-            return StarknetCurve.pedersen(aSorted, bSorted)
+            return hashFunction.hash(aSorted, bSorted)
```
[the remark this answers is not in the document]
THenry14: hmm, I'm leaning towards ditching the defaults altogether tbh. I'm worried the defaults could cause us unnecessary trouble, by ditching we give all the power (and responsibility) to users. They will have to check which hash they need, but at least this will be an informed choice they will de before using the function (and not after).

lib/src/main/kotlin/com/swmansion/starknet/crypto/HashMethod.kt
DelevoXDG: Yes, it is gonna be needed and will be added in #428. Just wanted to keep this PR focused on MerkleTree changes.

lib/src/main/kotlin/com/swmansion/starknet/crypto/HashMethod.kt
DelevoXDG: > Maybe `HashType` instead? 

I thought about that one as well. Just seems to me that it doesn't convey the function/purpose of this enum.

> indicates we are dealking with a function, and not an enum

Maybe, although I think it's already defined as an enum, so that might be enough?

THenry14: yeah, maybe you're right. But definitely Method, not Function imo

THenry14 (approved):

Issue #389, opened 2024-01-15
Bump demos to latest version


DelevoXDG:
### Feature Request

Bump demos to latest version

PR #445, opened 2024-03-19
Fix build for android

DelevoXDG:
## Describe your changes

 
Fix errors and warnings observed when running `:android:compileDebugKotlin` that caused 0.11.1 release to fail. 
 
- Don't use `BigInteger.TWO`: not supported in Java 1.8
- `android` deps
  - Add `kotlin-reflect`
  - Bump and format deps
- Update `NativeLoader` to handle `System.getProperty()` returning null

## Linked issues

Closes

## Breaking changes

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -57,8 +57,8 @@ dependencies {
     testImplementation 'junit:junit:4.13.2'
     androidTestImplementation 'androidx.test.ext:junit:1.1.5'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
-    implementation 'org.bouncycastle:bcprov-jdk18on:1.76'
-    implementation("com.squareup.okhttp3:okhttp:4.11.0")
+    implementation 'org.bouncycastle:bcprov-jdk18on:1.77'
+    implementation("com.squareup.okhttp3:okhttp:4.12.0")
     implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.5.1")
     implementation 'org.jetbrains.kotlin:kotlin-reflect:1.9.23'
 }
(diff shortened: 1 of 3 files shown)


lib/src/main/kotlin/com/swmansion/starknet/crypto/NativeLoader.kt
```diff
@@ -57,6 +57,8 @@ internal object NativeLoader {
     class UnsupportedPlatform(system: String, architecture: String) :
         RuntimeException("Unsupported platfrom $system:$architecture")
 
+    class UnknownOS : RuntimeException("Unknown OS")
```
THenry14, line 60:
```suggestion
    class UnknownOS : RuntimeException("Failed to fetch OS name")
```
nit

android/build.gradle
```diff
@@ -57,7 +57,8 @@ dependencies {
     testImplementation 'junit:junit:4.13.2'
     androidTestImplementation 'androidx.test.ext:junit:1.1.5'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
-    implementation 'org.bouncycastle:bcprov-jdk18on:1.76'
-    implementation("com.squareup.okhttp3:okhttp:4.11.0")
-    implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.5.1")
+    implementation 'org.bouncycastle:bcprov-jdk18on:1.77'
```
THenry14, line 60: have you tried to build it locally + run the demo? I vaguely remember we had some problems last time we tried to bump some deps

THenry14 (approved):

android/build.gradle
DelevoXDG: Yeah,  it now build locally with no errors. 

As for the demo, we'd need to do #389 and as of now, it would be much easier to test demo once we actually release the new version.

PR #449, opened 2024-03-21
Allow integer `revision` in `TypedData.Domain`

### LINKED ISSUE #448: `revision` value in typed data revision 1 should be an integer
### Feature Request

https://github.com/starknet-io/SNIPs/pull/79

DelevoXDG:
## Describe your changes

- `TypedData.Domain` can now be decoded from the following json, where `revision` is an integer:
```
"domain": {
    "name": "Starknet Example",
    "version": "1",
    "chainId": "SN_MAIN",
    "revision" : 1
  },
```
- Add `domain v0 object with string revision` in `TypedDataTest`

## Linked issues

Closes #448 

## Breaking changes

- `TypedData.Domain.Revision` is now of type `JsonPrimitive?` instead of `TypedData.Revision?`. Use `TypedData.Domain.resolvedRevision` to access revision as `TypedData.Revision`


DelevoXDG:
Even though  this change is not breaking for the typed data where `revision` is a string, I suppose it's best if we wait for https://github.com/starknet-io/SNIPs/pull/79 to be merged.

lib/src/main/kotlin/com/swmansion/starknet/data/TypedData.kt
```diff
@@ -175,12 +175,22 @@ data class TypedData private constructor(
...
         val version: JsonPrimitive,
         val chainId: JsonPrimitive,
-        val revision: Revision? = null,
+        val revision: JsonPrimitive? = null,
     ) {
-        internal val separatorName = when (revision ?: Revision.V0) {
+        @Transient
+        val resolvedRevision = revision?.let {
+            // Allow both string and integer values for revision
+            lenientJson.decodeFromJsonElement<Revision>(it)
```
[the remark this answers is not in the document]
franciszekjob: Yes, `decodeFromJsonElement` is able to handle that. Let's remove `lenientJson` as we discussed offline.

franciszekjob (approved):
LGTM

lib/src/test/kotlin/starknet/data/TypedDataTest.kt
```diff
@@ -548,7 +548,7 @@ internal class TypedDataTest {
                 "name": "DomainV1",
                 "version": "1",
                 "chainId": "2137",
-                "revision": "1"
```
THenry14, line 551: do we have a test with "1" as well? I suppose it may still be possible to have such response, right?

franciszekjob: Correct, considering `"1"` may occur in response, then these tests should be also added.

THenry14 (approved):

lib/src/test/kotlin/starknet/data/TypedDataTest.kt
```diff
@@ -524,7 +524,7 @@ internal class TypedDataTest {
     }
 
     @Nested
-    inner class InvalidTypesTest {
+    inner class TypesTest {
```
franciszekjob, line 527: A there was need to test resolved revision correctness (from domain object bot as string and int), I've decided to rename `InvalidTypesTest` to `TypesTest` and broaden its scope.
There is also a possibility to keep `InvalidTypesTest` and create separate eg. `ValidTypesTest` class. This way we will keep current test score and create add a new one. Wdyt? @DelevoXDG @THenry14

lib/src/test/kotlin/starknet/data/TypedDataTest.kt
```diff
@@ -551,6 +551,14 @@ internal class TypedDataTest {
                 "revision": 1
             }
         """.trimIndent()
+        val domainObjectV1WithStringRevision = """
```
franciszekjob, line 554: I've decided to place domain object with revision here. It could be also placed directly in the test, but wanted to maintain consistency with the place where `domainObjectV1` is defined.

lib/src/test/kotlin/starknet/data/TypedDataTest.kt
DelevoXDG: I don't think `"1"` is possible per the https://github.com/starknet-io/SNIPs/pull/79

lib/src/test/kotlin/starknet/data/TypedDataTest.kt
```diff
@@ -703,6 +711,34 @@ internal class TypedDataTest {
...
+                types = mapOf(
+                    domainTypeV1,
+                ),
+                primaryType = "StarknetDomain",
+                domain = domainObjectV1,
+                message = "{\"StarknetDomain\": 1}",
+            ).domain.resolvedRevision
+
+            assertEquals(Revision.V1, resolvedRevision)
+        }
```
DelevoXDG, line 727: This case is already covered by all the other tests that are dealing with regular revision 1 typed data.
```suggestion
```

lib/src/test/kotlin/starknet/data/TypedDataTest.kt
DelevoXDG: I think we should move it directly to the test

lib/src/test/kotlin/starknet/data/TypedDataTest.kt
```diff
@@ -703,6 +711,34 @@ internal class TypedDataTest {
...
+                types = mapOf(
+                    domainTypeV1,
+                ),
+                primaryType = "StarknetDomain",
+                domain = domainObjectV1WithStringRevision,
+                message = "{\"StarknetDomain\": 1}",
+            ).domain.resolvedRevision
+
+            assertEquals(Revision.V1, resolvedRevision)
+        }
```
DelevoXDG, line 741: Nit: 
```suggestion
        @Test
        fun `domain v1 object with string revision`() {
            val td = TypedData(
                types = mapOf(
                    domainTypeV1,
                ),
                primaryType = "StarknetDomain",
                domain = domainObjectV1WithStringRevision,
                message = "{\"StarknetDomain\": 1}",
            )

            assertEquals(Revision.V1, resolvedRevision.domain.resolvedRevision)
        }
```

Just makes catching td creation errors more easy (hypothetically, very unlikely situation 😅)

lib/src/test/kotlin/starknet/data/TypedDataTest.kt
DelevoXDG: > A there was need to test resolved revision correctness (from domain object bot as string and int), I've decided to rename InvalidTypesTest to TypesTest and broaden its scope.

`InvalidTypesTests` is a scope to deal with **types** specifically, as in types (basic/preset/custom) used in typed data. So tests related to `revision` don't really fit there.

We should instead:
1. Remove `domain v1 object with integer revision`
2. Move `domainObjectV1WithStringRevision` inside  `domain v1 object with string revision` test
3. Move `domain v1 object with string revision` outside of `InvalidTypesTest`

> There is also a possibility to keep InvalidTypesTest and create separate eg. ValidTypesTest class.

I think this should be just its own separate test. You can look for another scope to put it, but definitely not anything with `Types` in it.

lib/src/test/kotlin/starknet/data/TypedDataTest.kt
DelevoXDG: > but this is atm limited to revision 0 right? 

Well yeah, but also:

> - **revision (optional)**: the revision of the specification to be used. If the value is omitted it will default to `0` .
>     - Revision `0`: Represents the de facto spec before this SNIP was published. The purpose is to help with backwards compatibility. It’s not recommended to use it.

We can add the test just for the sake of it, but we're testing something very unlikey

> I find this a bit confusing :D

It is, but apparently handling shortstrings as ints or vice versa is a typical thing for typed data as I previously discovered trying to introduce some additional type checks which resulted in tests failing for existing examples

lib/src/test/kotlin/starknet/data/TypedDataTest.kt
franciszekjob: Sure, that's why I've added remark about it. I will move it to separate test then.

lib/src/test/kotlin/starknet/data/TypedDataTest.kt
```diff
@@ -734,4 +734,32 @@ internal class TypedDataTest {
...
+                "chainId": "2137",
+                "revision": "1"
+            }
+        """.trimIndent()
+        val domainTypeV1 = "StarknetDomain" to listOf(
+            TypedData.StandardType("name", "shortstring"),
+            TypedData.StandardType("version", "shortstring"),
+            TypedData.StandardType("chainId", "shortstring"),
+            TypedData.StandardType("revision", "shortstring"),
+        )
```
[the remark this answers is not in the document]
franciszekjob: I think we can't - `domainTypeV1` used in `InvalidTypesTest` is accessible only in its scope, therefore it cannot be accessed in in `domain v1 object with string revision` test.

Its possible to create a field directly in `TypedDataTest`, but I don't think its a good solution. Wdyt? @DelevoXDG

DelevoXDG (commented):
LGTM

PR #455, opened 2024-04-24
Support batching multiple RPC requests

### LINKED ISSUE #443: Support batching multiple RPC requests
### Feature Request

Allow batching multiple RPC calls within the same request.
The nodes are accepting the array of RPC calls and they execute them all in parallel.

**from tg:**
> Here is an example on goerli calling balanceOf (for 2 accounts) on the DAI token contract
> ```json
> [
>  {
>  "jsonrpc": "2.0",
>  "id": "1",
>  "method": "starknet_call",
>  "params": [
>   {
>    "contract_address": "0x123",
>    "entry_point_selector": "0x456",
>    "calldata": [
>     "0x1"
>    ]
>   },
>    "pending"
>   ]
>  },
>  {
>   "jsonrpc": "2.0",
>   "id": "2",
>   "method": "starknet_call",
>   "params": [
>    {
>     "contract_address": "0x123",
>     "entry_point_selector": "0x456",
>     "calldata": [
>      "0x2"
>     ]
>    },
>    "pending"
>   ]
>  }
> ]
> 
> ```

franciszekjob:
## Describe your changes

- add `batchRequests` method (vararg and list versions)
- add `JsonRpcRequest` data class
- add `BatchHttpRequest` class 
- add test to `ProviderTest`:
    * `batch call contract with block hash and block tag`
    * `batch get transactions`
- add tests to `JsonRpcResponseTest`:
   * `rpc provider parses batch getTransactionStatus`
   * `rpc provider parses batch callContract`
   * `rpc provider parses batch response with incorrect order`
- implement `JsonRpcRequest` data class representing single RPC request
- update `starknet-jvm.md` readme with proper examples of `batchRequests` usage
## Linked issues

Closes #443 

## Breaking changes


lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -52,6 +52,8 @@ internal fun <T> buildJsonHttpDeserializer(
                 response.body,
             )
 
+        // TODO: In case of batch request, exception should not be thrown while checking
+        // each single response. Instead, it should be thrown when accessing specific call with an error.
```
THenry14, line 56: what do you mean by 'checking'?
I generally agree that exception should be thrown when accessing specific call with an error, unless there was some network error/node error - then we should throw it straight away (eg when node returned malformed response, or is unreachable)

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -68,6 +66,28 @@ class JsonRpcProvider(
...
+    ): BatchHttpRequest<T> {
+        val httpRequests = calls.map { it as HttpRequest<T> }
+        val updatedJsonObjects = httpRequests.mapIndexed { index, request ->
+            val jsonMap = Json.parseToJsonElement(request.payload.body.toString()).jsonObject.toMutableMap()
+            jsonMap["id"] = JsonPrimitive(index)
+            JsonObject(jsonMap)
+        }
+
+        val jsonArray = JsonArray(updatedJsonObjects)
+        val deserializer = httpRequests.first().deserializer
```
THenry14, line 80: can you explain this to me, since I'm not sure I follow - why do we take .first() here?
can `calls` be ever empty? what happens then?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -68,6 +66,28 @@ class JsonRpcProvider(
         return HttpRequest(payload, buildJsonHttpDeserializer(responseSerializer, deserializationJson), httpService)
     }
 
+    fun <T> combineBatchRequest(
+        calls: List<Request<T>>,
+    ): BatchHttpRequest<T> {
+        val httpRequests = calls.map { it as HttpRequest<T> }
```
THenry14, line 72: To be a bit more type safe, maybe we could first verify the elements are indeed HttpRequest<T>, or at least handle the possible exception?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -68,6 +66,28 @@ class JsonRpcProvider(
...
+            JsonObject(jsonMap)
+        }
+
+        val jsonArray = JsonArray(updatedJsonObjects)
+        val deserializer = httpRequests.first().deserializer
+        val payload = HttpService.Payload(
+            url,
+            "POST",
+            emptyList(),
+            jsonArray.toString(),
```
THenry14, line 85: we first made if json object, and then converted to string. I wonder if there is a way to make it more efficient. this might be a nitpick.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
franciszekjob: To be precisely, by checking I meant deserializing each response. I updated todo comment to make it more clear.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
DelevoXDG: It seems to me every request should have it's own `deserializer`

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
DelevoXDG: I'm not sure there's much room for improvement. We use the same approach in original `buildRequest`. 

We can use `Json.encodeToString` on `List<JsonRpcRequest>` if we opt for the approach I suggested above, but I doubt there would be much difference anyway.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
franciszekjob: Correct, I've changed the type from `Request` to `HttpRequest` 👍

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
```diff
@@ -21,3 +23,26 @@ class HttpRequest<T>(
...
+
+class BatchHttpRequest<T>(
+    val payload: HttpService.Payload,
+    val deserializer: HttpResponseDeserializer<T>,
+    val service: HttpService,
+) : Request<List<T>> {
+    private fun parseResponse(response: HttpResponse): List<T> {
+        val responseList = Json.parseToJsonElement(response.body).jsonArray
+        return responseList.map { responseJson ->
+            val deserializedValue = deserializer.apply(HttpResponse(response.isSuccessful, response.code, responseJson.toString()))
```
DelevoXDG, line 35: 1. We should probably refactor this logic to not create fake `HttpResponse`, as this way we're repeating the same checks multiple times. We should instead handle http deserialization once, and then handle each json deserialization separately.
2. As per the json-rpc [spec](https://www.jsonrpc.org/specification#batch):
> The Response objects being returned from a batch call MAY be returned in any order within the Array.
So we probably want to reorder the responses correctly based on their `id`'s

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
franciszekjob: Here we need to determine the deserializer for our response. Since `httpRequests` have the same generic type, we retrieve deserializer from the first one. I've added prevention from retrieving first element's deserializer when `calls` is empty.

lib/starknet-jvm.md
```diff
@@ -393,6 +413,39 @@ fun main(args: Array<String>) {
         low = response[0],
         high = response[1],
     )
+
+    // Create multiple calls
```
DelevoXDG, line 417: I don't have a strong opinion on this, but I think:
1. This should be a separate example
2. We should send calls to separate addresses, or better, to separate endpoints

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -68,41 +65,65 @@ class JsonRpcProvider(
         return HttpRequest(payload, buildJsonHttpDeserializer(responseSerializer, deserializationJson), httpService)
     }
 
-    override fun getSpecVersion(): Request<String> {
+    fun <T> combineBatchRequest(
+        calls: List<HttpRequest<T>>,
```
DelevoXDG, line 69: 1. 
```suggestion
        requests: List<HttpRequest<T>>,
```
2. I'm not sure I understand why are we batching only the requests of the same type.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
franciszekjob: Yes, calling `joinToString` on `updatedJsonObjects` with specifying `prefix` and `postifx` params is handy here, updated 👍

franciszekjob: Ok, can we leave it then with the original implementation?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -68,6 +66,28 @@ class JsonRpcProvider(
         return HttpRequest(payload, buildJsonHttpDeserializer(responseSerializer, deserializationJson), httpService)
     }
 
+    fun <T> combineBatchRequest(
+        calls: List<Request<T>>,
+    ): BatchHttpRequest<T> {
```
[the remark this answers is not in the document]
franciszekjob: 1. Let's choose `batchRequests` then
2. Good idea, it removes the need to wrap the calls in list while calling `batchRequests`.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
franciszekjob: I've updated this part, now every request has it's own deserializer.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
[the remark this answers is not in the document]
franciszekjob: In we want to keep information about occurring errors without directly throwing an exception, we have to return a list with success/error results. We can use existing `JsonRpcResponse` as it provides both result and error fields, but this causes a misalignment between reading data from a single and batch request. Batch request would then provide us with type `JsonRpcResponse<T>` and single request with `T`. Wdyt?

lib/starknet-jvm.md
franciszekjob: I moved it to separate example and changed the addresses. Example of separate methods usage can't be covered, as we currently don't support multiple methods yet.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -68,6 +66,28 @@ class JsonRpcProvider(
...
+        val httpRequests = calls.map { it as HttpRequest<T> }
+        val updatedJsonObjects = httpRequests.mapIndexed { index, request ->
+            val jsonMap = Json.parseToJsonElement(request.payload.body.toString()).jsonObject.toMutableMap()
+            jsonMap["id"] = JsonPrimitive(index)
+            JsonObject(jsonMap)
```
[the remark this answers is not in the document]
franciszekjob: Reasonable idea, I've created `JsonRpcRequest` data class, and reorganized passing and creating payload references.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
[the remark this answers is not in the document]
franciszekjob: This case will be addresses in following up pull request, as we discussed offline 👍

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
franciszekjob: I've added responses ordering and reorganized deserialization. Now `buildJsonBatchHttpDeserializer` is responsible for deserializing batch response.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -52,19 +69,37 @@ internal fun <T> buildJsonHttpDeserializer(
...
         }
 
-        if (jsonRpcResponse.result == null) {
-            throw RequestFailedException(message = "Response did not contain a result", payload = response.body)
+        val jsonArray = Json.parseToJsonElement(response.body).jsonArray
+        val jsonRpcResponses = deserializationStrategies.mapIndexed { index, strategy ->
+            deserializationJson.decodeFromString(
+                JsonRpcResponse.serializer(strategy),
+                jsonArray[index].toString(),
+            )
```
DelevoXDG, line 96: 1. Let's use `decodeFromJsonElement` instead to avoid repeat conversions
2. Mapping strategies to responses seems a little unintuitive to me, maybe we could map `jsonArray` instead?
3. I was wrong during our offline discussion. We do need to either access response `"id"` using raw json object or sort by it to ensure correct match between deserializer and response:
```suggestion
        val responses = jsonArray.map {
            deserializationJson.decodeFromJsonElement(
                JsonRpcResponse.serializer(deserializationStrategies[it.jsonObject["id"]!!.jsonPrimitive.int]), it,
            )
        }
```
Or since we don't support multiple types for now, we could go back to using a single deserializer.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -8,15 +8,29 @@ import com.swmansion.starknet.data.serializers.TransactionReceiptPolymorphicSeri
...
+import com.swmansion.starknet.service.http.*
 import kotlinx.serialization.KSerializer
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
 import kotlinx.serialization.builtins.ListSerializer
 import kotlinx.serialization.builtins.serializer
 import kotlinx.serialization.json.*
 
+@Serializable
+data class JsonRpcRequest(
```
DelevoXDG, line 20: Should this be public?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -44,65 +58,85 @@ class JsonRpcProvider(
...
-        val payload =
-            HttpService.Payload(url, "POST", emptyList(), requestJson.toString())
+        return HttpRequest(url, jsonRpcRequest, responseSerializer, deserializationJson, httpService)
+    }
 
-        return HttpRequest(payload, buildJsonHttpDeserializer(responseSerializer, deserializationJson), httpService)
+    fun <T> batchRequests(requests: List<HttpRequest<T>>): BatchHttpRequest<T> {
+        if (requests.isEmpty()) {
+            throw IllegalArgumentException("Please provide requests while creating a batching requests")
+        }
```
DelevoXDG, line 79:
```suggestion
        require(requests.isNotEmpty()) { "Please provide requests while creating a batching requests" }
```

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -44,65 +58,85 @@ class JsonRpcProvider(
...
 
-        val payload =
-            HttpService.Payload(url, "POST", emptyList(), requestJson.toString())
+        return HttpRequest(url, jsonRpcRequest, responseSerializer, deserializationJson, httpService)
+    }
 
-        return HttpRequest(payload, buildJsonHttpDeserializer(responseSerializer, deserializationJson), httpService)
+    fun <T> batchRequests(requests: List<HttpRequest<T>>): BatchHttpRequest<T> {
+        if (requests.isEmpty()) {
+            throw IllegalArgumentException("Please provide requests while creating a batching requests")
```
DelevoXDG, line 78: Nit:
```suggestion
            throw IllegalArgumentException("Cannot create a batch request from an empty list of requests.")
```

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -44,65 +58,85 @@ class JsonRpcProvider(
...
+    fun <T> batchRequests(vararg requests: HttpRequest<T>): BatchHttpRequest<T> {
+        if (requests.isEmpty()) {
+            throw IllegalArgumentException("Please provide requests while creating a batching requests")
+        }
+
+        return buildBatchRequest(requests.toList())
+    }
+
+    private fun <T> buildBatchRequest(requests: List<HttpRequest<T>>): BatchHttpRequest<T> {
+        val updatedJsonRpcRequests = requests.mapIndexed { index, request ->
```
DelevoXDG, line 93: Nit: could we have a more readable name, perhaps `orderedRequests`?

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
```diff
@@ -1,23 +1,81 @@
...
-    val payload: HttpService.Payload,
-    val deserializer: HttpResponseDeserializer<T>,
+    val url: String,
+    val jsonRpcRequest: JsonRpcRequest,
+    val serializer: KSerializer<T>,
+    val deserializationJson: Json,
     val service: HttpService,
 ) : Request<T> {
 
+    private fun getPayload(): HttpService.Payload {
```
DelevoXDG, line 22: 1. I think it should be `preparePayload` or `buildPayload`
2. Nit: We might want to make this a lazy property to cache payload for repeat requests, @THenry14 wdyt?

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
```diff
@@ -1,23 +1,81 @@
...
+        val body = Json.encodeToString(JsonRpcRequest.serializer(), jsonRpcRequest)
```
DelevoXDG, line 23: Nit:
```suggestion
        val body = Json.encodeToString(jsonRpcRequest)
```
we can also inline it

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
```diff
@@ -1,23 +1,81 @@
...
+    val deserializationJson: Json,
+    val service: HttpService,
+) : Request<List<T>> {
+    private fun parseResponse(response: HttpResponse): List<T> {
+        val results = buildJsonBatchHttpDeserializer<T>(responseSerializers, deserializationJson).apply(response)
+        return results
+    }
+
+    private fun getPayload(): HttpService.Payload {
+        val body = jsonRpcRequests.map { Json.encodeToString(JsonRpcRequest.serializer(), it) }.toString()
```
DelevoXDG, line 60: Nit:
```suggestion
        val body = Json.encodeToString(jsonRpcRequests)
```

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
```diff
@@ -1,23 +1,81 @@
...
+        val deserializer = buildJsonHttpDeserializer(serializer, deserializationJson)
+        val payload = getPayload()
         return service.sendAsync(payload).thenApplyAsync(deserializer)
     }
 }
+
+class BatchHttpRequest<T>(
+    val url: String,
+    val jsonRpcRequests: List<JsonRpcRequest>,
+    val responseSerializers: List<KSerializer<T>>,
```
DelevoXDG, line 50: Nit: `responseDeserializers` or `deserializers`

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -1200,4 +1200,53 @@ class ProviderTest {
...
+                call = call1,
+                blockHash = blockHash,
+            ),
+            provider.callContract(
+                call = call2,
+                blockTag = BlockTag.LATEST,
+            ),
+        )
+        val request = provider.batchRequests(providerCalls)
+        org.junit.jupiter.api.assertAll({ request.send() })
```
DelevoXDG, line 1230: Let's import this instead.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -1200,4 +1200,53 @@ class ProviderTest {
...
+
+        val response = request.send()
```
DelevoXDG, line 1232: What is it exactly we are asserting here? Why send another same request right away?

lib/src/test/kotlin/starknet/provider/response/JsonRpcResponseTest.kt
```diff
@@ -190,4 +189,90 @@ class JsonRpcResponseTest {
...
+
+        val ethContractAddress = Felt.fromHex("0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7")
+
+        val httpServiceMock = mock<HttpService> {
+            on { send(any()) } doReturn HttpResponse(true, 200, mockResponse)
+        }
+        val provider = JsonRpcProvider("", httpServiceMock)
+
+        val call = Call(
+            ethContractAddress,
```
DelevoXDG, line 224: Nit: we can use `DevnetClient.ethErc20ContractAddress` or inline the value since this is a mock test anyway.

lib/src/test/kotlin/starknet/provider/response/JsonRpcResponseTest.kt
DelevoXDG, line 1: Maybe we should add a test / tests with:
- Response array that is not correctly ordered
- Deals with polymormic serializers (tx or tx receipt)

lib/starknet-jvm.md
```diff
@@ -396,6 +396,107 @@ fun main(args: Array<String>) {
...
+        // Specify the contract addresses, in this example ETH ERC20 contracts are used
+        Felt contractAddress1 = Felt.fromHex("0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7");
+        Felt contractAddress2 = Felt.fromHex("0x05dcf20d87407f0f7a6bd081a2cee990b19c77ebf31c8b0ddd5a6f931af21e0c");
+        
+        // Create multiple calls
+        List<Felt> calldata = List.of(account.getAddress());
+
+        Call call1 = new Call(contractAddress1, "balanceOf", calldata);
+        Call call2 = new Call(contractAddress2, "balanceOf", calldata);
+        Request<List<List<<Felt>>> request = provider.batchRequests(
```
DelevoXDG, line 434: Nit:
```suggestion

        // Create a batch request
        Request<List<List<<Felt>>> request = provider.batchRequests(
```

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -3,13 +3,14 @@ package com.swmansion.starknet.provider.rpc
...
 import kotlinx.serialization.*
-import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.*
 import java.util.function.Function
 
 @Serializable
-private data class JsonRpcResponse<T>(
+internal data class JsonRpcResponse<T>(
     @SerialName("id")
     val id: Int,
```
DelevoXDG, line 15: Since we're dealing with batch requests now, we should consider whether we ever wanna handle request with null `id`, as it is _technically_ possible per the [spec](https://www.jsonrpc.org/specification#response_object)

lib/src/main/kotlin/com/swmansion/starknet/provider/Provider.kt
```diff
@@ -19,7 +20,7 @@ interface Provider {
      *
      * @throws RequestFailedException
      */
-    fun getTransaction(transactionHash: Felt): Request<Transaction>
+    fun getTransaction(transactionHash: Felt): HttpRequest<Transaction>
```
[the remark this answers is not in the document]
franciszekjob: I agree, `Provider` has higher level of abstraction, and in `HttpRequest` has rpc-related references.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -3,13 +3,14 @@ package com.swmansion.starknet.provider.rpc
...
+import com.swmansion.starknet.service.http.HttpResponse
 import com.swmansion.starknet.service.http.HttpResponseDeserializer
 import kotlinx.serialization.*
-import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.*
 import java.util.function.Function
 
 @Serializable
-private data class JsonRpcResponse<T>(
+internal data class JsonRpcResponse<T>(
```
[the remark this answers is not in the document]
franciszekjob: Updated.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
franciszekjob: In fact my intention was also to test `send` method here, but if so:
1. It should be a separate test (do we actually need it, wdyt?)
2. Usage of `assertAll` is not appropriate here
I will address the change.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -35,6 +36,22 @@ internal data class JsonRpcError(
     val data: String? = null,
 )
 
+internal fun <T> extractResult(jsonRpcResponse: JsonRpcResponse<T>, response: HttpResponse): T {
+    if (jsonRpcResponse.error != null) {
+        throw RpcRequestFailedException(
+            code = jsonRpcResponse.error.code,
+            message = jsonRpcResponse.error.message,
+            data = jsonRpcResponse.error.data,
+            payload = response.body,
```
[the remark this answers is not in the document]
franciszekjob: Good point, let's change the payload from whole list of requests to only the failed one.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
franciszekjob: 1. Fine
2. Agree, it may be more meaningful
3. Ok, good that we are both on the same page.

Regarding usage of single deserializer - it's actually more intuitive to pass deserializer for each response, plus incoming feature will require it. Do you agree we can keep list of deserializers?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
franciszekjob: `JsonRpcRequest` was used in other file before, I didn't change its visibility later. Now it's private 👍

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
franciszekjob: It think it's better to make it a lazy property. Updated.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
franciszekjob: Updated, it was completely removed (explanation in following comment).

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
franciszekjob: In case of batch request, we shouldn't handle request with null `id` at all. Here's a quick example:
1. We send batch request 
2. We receive responses
3. Two or more of them have null `id`
4. We cannot identify them

Technically, if only one response has null `id`, we can identify it by elimination of others which are successful. However, this is a dirty solution and I don't think it's the way to go.

Wdyt @DelevoXDG ?

lib/src/test/kotlin/starknet/provider/response/JsonRpcResponseTest.kt
franciszekjob: Sure, I've added them.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
DelevoXDG: I don't really understand the first point. Do we need what?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
DelevoXDG: Ok, I suppose since we're always sending `id` in our requests, we should always be getting `id` in the responses and this doesn't apply:

> If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
franciszekjob: I meant, do we also need to test sending `send` method here.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -1200,4 +1200,65 @@ class ProviderTest {
...
+        )
+
+        val response = request.send()
+
+        assertNotNull(response[0])
+        assertNotNull(response[1])
+    }
+
+    @Test
+    fun `batch get transactions with polymorphic serializers`() {
```
DelevoXDG, line 1252:
```suggestion
    fun `batch get transactions`() {
```

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -1200,4 +1200,65 @@ class ProviderTest {
...
+        val blockNumber = provider.getBlockNumber().send()
+        val request = provider.batchRequests(
+            provider.getTransactionByBlockIdAndIndex(blockNumber, 0),
+            provider.getTransactionByBlockIdAndIndex(blockNumber, 0),
+        )
+
+        val response = request.send()
+
+        assertNotNull(response[0])
+        assertNotNull(response[1])
```
DelevoXDG, line 1248: Responses are not nullable, so `assertNotNull` will always pass. 

It's better to test against expected values, or at least against zero if we don't know those.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -1200,4 +1200,65 @@ class ProviderTest {
...
+    }
+
+    @Test
+    fun `batch get transactions with polymorphic serializers`() {
+        val blockNumber = provider.getBlockNumber().send()
+        val request = provider.batchRequests(
+            provider.getTransactionByBlockIdAndIndex(blockNumber, 0),
+            provider.getTransaction(invokeTransactionHash),
```
DelevoXDG, line 1256: We can probably delete `batch get transactions by block number and index` since it can be the part of this test.

lib/src/test/kotlin/starknet/provider/response/JsonRpcResponseTest.kt
```diff
@@ -190,4 +190,132 @@ class JsonRpcResponseTest {
...
+
+        val httpServiceMock = mock<HttpService> {
+            on { send(any()) } doReturn HttpResponse(true, 200, mockResponse)
+        }
+        val provider = JsonRpcProvider("", httpServiceMock)
+
+        val call = Call(
+            DevnetClient.ethErc20ContractAddress,
+            Felt.fromHex("0x2e4263afad30923c891518314c3c95dbe830a16874e8abc5777a9a20b54c76e"),
+            listOf(Felt.fromHex("0x07f6331182b9bcbf9c1a5943e309c05399a935d170f7f07494cdf7a174cd7527")),
```
DelevoXDG, line 225: We should use something more similar to the other tests where entrypoint is a string. No reason for mock test to have complicated values so it's not clear what it would do if it was a regular test.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
franciszekjob: Let's make it clear - usage of extra `send` and assertion here was intended to emphasize test of sending the request. As you pointed out, `send` is used anyway, so there's no point to make it redundant and overcomplicate the test. Neither an assertion nor an additional usage of `send` is needed here. They have been removed.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
franciszekjob: Change applied.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -35,6 +36,38 @@ internal data class JsonRpcError(
...
+            payload = payload,
+        )
+    }
+
+    if (jsonRpcResponse.result == null) {
+        throw RequestFailedException(message = "Response did not contain a result", payload = response.body)
+    }
+    return jsonRpcResponse.result
+}
+
```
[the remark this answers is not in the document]
franciszekjob: I merged it into one function and changed its parameters `fun <T> extractResult(jsonRpcResponse: JsonRpcResponse<T>, totalPayload: String, payload: String = totalPayload): T`. 
`totalPayload` is body of the response while `payload` is specific part of the body.
This way we will either call `extractResult(jsonRpcResponse, response.body)` or `extractResult(jsonRpcResponse, response.body, jsonResponse.toString())`.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -52,19 +85,38 @@ internal fun <T> buildJsonHttpDeserializer(
...
+    return Function { response ->
+        if (!response.isSuccessful) {
+            throw RequestFailedException(
                 payload = response.body,
             )
         }
 
-        if (jsonRpcResponse.result == null) {
-            throw RequestFailedException(message = "Response did not contain a result", payload = response.body)
+        val jsonArray = Json.parseToJsonElement(response.body).jsonArray
```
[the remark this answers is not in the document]
franciszekjob: Changed applied.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -52,19 +85,38 @@ internal fun <T> buildJsonHttpDeserializer(
...
+                jsonArray[index].toString(),
+            )
         }
 
-        jsonRpcResponse.result
+        val results = jsonRpcResponses.sortedBy { it.id }.zip(jsonArray)
+            .map { (jsonRpcResponse, json) ->
+                val payload = json.toString()
+                extractResult(jsonRpcResponse, response, payload)
+            }
```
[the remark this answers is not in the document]
franciszekjob: Change applied.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -52,19 +85,38 @@ internal fun <T> buildJsonHttpDeserializer(
...
+    }
+}
+
+internal fun <T> buildJsonBatchHttpDeserializer(
+    deserializationStrategies: List<KSerializer<T>>,
+    deserializationJson: Json,
+): HttpResponseDeserializer<List<T>> {
+    // TODO: In case of batch request, exception should not be thrown.
+    // Instead, we want to return wrapped responses.
+    // This enables access to successful results or encountered errors.
```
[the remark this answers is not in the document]
franciszekjob: Ok, will open new PR for that.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
franciszekjob: Name change applied.

lib/src/test/kotlin/starknet/provider/response/JsonRpcResponseTest.kt
franciszekjob: I've changed it to `Felt.ZERO`.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -1200,4 +1200,65 @@ class ProviderTest {
...
+        assertEquals(response[0].first(), expectedBalance)
+        assertEquals(response[1].first(), expectedBalance)
+    }
+
+    @Test
+    fun `batch get transactions by block number and index`() {
+        val blockNumber = provider.getBlockNumber().send()
+        val request = provider.batchRequests(
+            provider.getTransactionByBlockIdAndIndex(blockNumber, 0),
+            provider.getTransactionByBlockIdAndIndex(blockNumber, 0),
```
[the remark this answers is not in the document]
franciszekjob: 1. I've added checking different transaction hashes (`invokeTransactionHash`, `declareTransactionHash` and `deployAccountTransactionHash`).
2. I've increased requests number to 4 in `batch get transactions` test.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
franciszekjob: As we discussed offline, our requests will always include correct `id`, so we can assume that properly working server will also return responses with corresponding `id` values. In case response has `id`, the error is caused by the server and in such situation we don't want to handle the response.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
franciszekjob: Yes, I marked it by accident, all the points should be addresses now.

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
```diff
@@ -1,23 +1,78 @@
...
+
+class BatchHttpRequest<T>(
+    val url: String,
+    val jsonRpcRequests: List<JsonRpcRequest>,
+    val responseDeserializers: List<KSerializer<T>>,
+    val deserializationJson: Json,
+    val service: HttpService,
+) : Request<List<T>> {
+    private fun parseResponse(response: HttpResponse): List<T> {
+        val results = buildJsonBatchHttpDeserializer<T>(responseDeserializers, deserializationJson).apply(response)
```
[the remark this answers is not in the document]
franciszekjob: I've updated the code - `deserializer` and `payload `are now created once and reused. Also, I've added lazy modifier to payload in `HttpRequest`.

franciszekjob:
I've moved update of `Cmake` version to separate [PR](https://github.com/software-mansion/starknet-jvm/pull/460) as it is not in the scope of current task.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -44,65 +58,81 @@ class JsonRpcProvider(
...
+
+        return buildBatchRequest(requests)
+    }
 
-        return HttpRequest(payload, buildJsonHttpDeserializer(responseSerializer, deserializationJson), httpService)
+    fun <T> batchRequests(vararg requests: HttpRequest<T>): BatchHttpRequest<T> {
+        require(requests.isNotEmpty()) { "Please provide requests while creating a batching requests" }
+
+        return buildBatchRequest(requests.toList())
+    }
```
DelevoXDG, line 86: Can avoid some minor duplication here:
```suggestion
    fun <T> batchRequests(vararg requests: HttpRequest<T>): BatchHttpRequest<T> {
        return batchRequests(requests.toList())
    }
```
or
```suggestion
    fun <T> batchRequests(vararg requests: HttpRequest<T>): BatchHttpRequest<T> = batchRequests(requests.toList())
```

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -44,65 +58,81 @@ class JsonRpcProvider(
...
 
-        val payload =
-            HttpService.Payload(url, "POST", emptyList(), requestJson.toString())
+        return HttpRequest(url, jsonRpcRequest, responseSerializer, deserializationJson, httpService)
+    }
+
+    fun <T> batchRequests(requests: List<HttpRequest<T>>): BatchHttpRequest<T> {
+        require(requests.isNotEmpty()) { "Please provide requests while creating a batching requests" }
+
+        return buildBatchRequest(requests)
```
DelevoXDG, line 79: Seems like it can be inlined if you refactor vararg version of the method?

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -1200,4 +1200,55 @@ class ProviderTest {
...
+        )
+        val blockHash = provider.getBlockHashAndNumber().send().blockHash
+
+        val call2 = Call(
+            contractAddress = balanceContractAddress,
+            entrypoint = "get_balance",
+            calldata = emptyList(),
+        )
+
+        val providerCalls = listOf(
```
DelevoXDG, line 1219: Nit:
```suggestion
        val callRequests = listOf(
```

lib/src/test/kotlin/starknet/provider/response/JsonRpcResponseTest.kt
```diff
@@ -190,4 +190,132 @@ class JsonRpcResponseTest {
         assertEquals("Internal error", exception.message)
         assertEquals("[\"Invalid message selector\",\"0x1234\"]", exception.data)
     }
+
+    @Test
+    fun `rpc provider parses batch callContract`() {
```
DelevoXDG, line 195: Nit:
```suggestion
    fun `rpc provider parses batch callContract response`() {
```

lib/src/test/kotlin/starknet/provider/response/JsonRpcResponseTest.kt
DelevoXDG, line 1: Are you sure this many JsonRpcResponseTests for batching are necessary? 

Just seems to me that it's enough to have a normal test and incorrect order test, or even just incorrect order test would be sufficient. The rest can be tested through normal tests without mocks. 

WDYT?

lib/starknet-jvm.md
```diff
@@ -396,6 +396,109 @@ fun main(args: Array<String>) {
...
+import com.swmansion.starknet.provider.Request;
+import com.swmansion.starknet.provider.rpc.JsonRpcProvider;
+
+import java.math.BigInteger;
+import java.util.List;
+
+public class Main {
+    public static void main(String[] args) {
+        // Create a provider for interacting with Starknet
+        Provider provider = new JsonRpcProvider("https://example-node-url.com/rpc");
```
DelevoXDG, line 417: Just making sure, have you tried running this example code? 

It seems to me that this might not work, as `provider` is of `Provider` interface, and `Provider` interface (I assume intentionally) does not support `batchRequests`.

DelevoXDG (commented):
Some minor things remain to be addressed, but looking good overall

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
franciszekjob: Let's choose first suggestion 👍

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
```diff
@@ -1,23 +1,86 @@
...
-class HttpRequest<T>(
-    val payload: HttpService.Payload,
-    val deserializer: HttpResponseDeserializer<T>,
-    val service: HttpService,
+class HttpRequest<T> private constructor(
+    val jsonRpcRequest: JsonRpcRequest,
+    val serializer: KSerializer<T>,
+    private val payload: Lazy<HttpService.Payload>,
+    private val deserializer: HttpResponseDeserializer<T>,
+    private val service: HttpService,
```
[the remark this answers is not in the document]
franciszekjob: `jsonRpcRequest` and `serializer` aren't private on purpose. They are both accessed from `HttpRequest` object  inside [buildBatchRequest](https://github.com/software-mansion/starknet-jvm/blob/6406e1c50e92d7881a5d8583804b1f307e304894/lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt#L86) method.
I've changed them both to internal in order to keep them less visible. Wdyt?

lib/src/test/kotlin/starknet/provider/response/JsonRpcResponseTest.kt
franciszekjob: I agree 👍 - keeping only incorrect order test seems reasonable, because the rest of tests is in fact present in `JsonRpcProviderTest`. Furthermore, `JsonRpcResponseTest` is dedicated to test different scenarios of incoming responses (different errors etc).

I've removed redundant tests.

lib/starknet-jvm.md
franciszekjob: Correct, `batchRequests` was is not supported by `Provider`. It is a custom method, and is not an equivalent of real Starknet endpoint. That's why I decided not to include it in `Provider`.

However, it may be potentially confusing for users that they cannot use `batchRequests` on an instance of `Provider`.

Wdyt?

Regarding the test - I've updated sample code and replaced `Provider` with `JsonRpcProvider` 👍

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
franciszekjob: Could you explain more about what you mean by "refactoring the vararg version of the method"? 😄

lib/starknet-jvm.md
DelevoXDG: > However, it may be potentially confusing for users that they cannot use batchRequests on an instance of Provider.

DelevoXDG: >  It is a custom method and not an equivalent of specified Starknet endpoint. That's why I decided not to include it in Provider.

I would say it's JSON-RPC specific method. Previously, we also had `GatewayProvider` which likely would not support batching. However, currently afaik there are no ways to communicate with Starknet other than JSON-RPC, so theoretically all providers are "JSON-RPC" providers, and the original naming was retained mainly for backward compatibility.

The way I see it, we either:
1. Leave batching only in `JsonRpcProvider`, because it's JSON-RPC specific.
2. Push batching up to `Provider` interface to allow users to keep using the `Provider` as an interface.

It seems to me 2. offers broader utility, however, it slightly breaks abstraction by introducing some underlying request and response mechanics to the provider interface. 

WDYT?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
DelevoXDG: > Could you explain a bit more what you mean 😄 ? 

`buildBatchRequest ` can be removed with its contents inlined here, as it's not used anywhere else

lib/src/main/kotlin/com/swmansion/starknet/service/http/HttpRequest.kt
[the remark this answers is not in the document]
DelevoXDG: > They are both accessed from `HttpRequest` object inside [buildBatchRequest](https://github.com/software-mansion/starknet-jvm/blob/6406e1c50e92d7881a5d8583804b1f307e304894/lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt#L86) method.

That's what I meant 😄 

We can probably now make `JsonRpcRequest` internal as well: https://github.com/software-mansion/starknet-jvm/pull/455/files#r1581247674

lib/starknet-jvm.md
franciszekjob: While I generally uphold the principle of not violating the level of abstraction, the deviation in this case is rather minimal, making the 2. option possibly acceptable. Consequently, implementing `Provider` interface in `JsonRpcpProvider` currently seems unnecessary and doesn't add significant value. Do we want to keep it, if all providers are "JSON-RPC" providers?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
franciszekjob: Got it. I've applied proper refactor. 🫡

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -88,18 +92,18 @@ internal fun <T> buildJsonBatchHttpDeserializer(
...
             )
+            orderedResults[id] = extractResult(jsonRpcResponse, response.body, jsonElement.toString())
         }
 
-        val results = responses.sortedBy { it.id }.zip(jsonResponses)
-            .map { (jsonRpcResponse, jsonResponse) ->
-                extractResult(jsonRpcResponse, response.body, jsonResponse.toString())
-            }
-        results
+        orderedResults.filterNotNull()
```
DelevoXDG: I get the idea. But I'm not sure the linear ordering here is worth introducing mutability and breaking null-safety, as these batch requests are very unlikely to have considerably big number of requests. 

Are you sure there's no solution for this to maintain these two?

lib/starknet-jvm.md
DelevoXDG: >  Do we want to keep it, if all providers are "JSON-RPC" providers?

Discussed offline.

> the deviation in this case is rather minimal, making the 2. option possibly acceptable

The only thing that concerns me is that we'd introduce `HttpRequest` to the method signature of an interface, which is JSON-RPC specific.

I don't have a strong say on what's the best approach here, @THenry14 wdyt?

THenry14: No strong opinion either, but I guess it makes a bit more sense to me to keep it in JsonRpcProvider

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
DelevoXDG: Discussed offline, decided to leave the current approach 👍🏻

DelevoXDG (approved):
LGTM

PR #461, opened 2024-05-06
Support batching multiple RPC requests of different types

franciszekjob:
## Describe your changes

Add support for batching requests of different type
   - add `JsonRpcProvider.batchRequestsAny()`
   - add `HttpBatchRequest` class
   - add `Response` as common return type for `JsonRpcProvider` calls
   - add `IntResponse`, `StringResponse`
   - add inline classes as wrappers for built-in types returned by `JsonRpcProvider` methods and their corresponding serializers
   - add `FeltArraySerializer`
   
Allow proceeding with failure of particular requests while using `batchRequests` and `batchRequestsAny`
- wrap values returned by `send()` and `sendAsync()` methods in `Result`

Update tests:
- add `batch requests any` in `ProviderTest`
- add `rpc provider parses batch response with error` in `JsonRpcResponseTest`
- update `JsonRpcResponseTest` due to changes of `JsonRpcProvider` methods

## Linked issues

Closes

## Breaking changes

- Change returned type in `JsonRpcProvider` methods:
   - `getSpecVersion(): HttpRequest<String>` -> `getSpecVersion(): HttpRequest<StringResponse>`
   - `callContract(...): HttpRequest<List<Felt>>` -> `callContract(...): HttpRequest<FeltArray>`
   - `getBlockNumber(): HttpRequest<Int>` -> `getBlockNumber(): HttpRequest<StarknetInt>`
   - `getBlockTransactionCount(...): HttpRequest<Int>` -> `getBlockTransactionCount(...): HttpRequest<IntResponse>`
   - `getEstimateFee(...): HttpRequest<List<EstimateFeeResponse>>` -> `getEstimateFee(...): HttpRequest<EstimateFeeResponseList>`
   - `simulateTransactions(...): HttpRequest<List<SimulatedTransaction>>` -> `simulateTransactions(...): HttpRequest<SimulatedTransactionList>`


.github/workflows/checks.yml
```diff
@@ -29,10 +29,10 @@ jobs:
           java-version: 11
 
       - name: Setup cmake
-        uses: jwlawson/actions-setup-cmake@4f73d30c2fc44a9466a3ed890fb8db7a7b554302
```
THenry14, line 32: let's make sure all the changes to master/upstream branch are present

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/IntWrapperSerializer.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import com.swmansion.starknet.data.types.IntWrapper
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+
+object IntWrapperSerializer : KSerializer<IntWrapper> {
```
THenry14, line 11: not sure about the name tbh

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/StringWrapperSerializer.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import com.swmansion.starknet.data.types.StringWrapper
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+
+object StringWrapperSerializer : KSerializer<StringWrapper> {
```
THenry14, line 11: the same remark about the name, not sure StringWrapper is the best choice

lib/src/main/kotlin/com/swmansion/starknet/data/types/FeltArray.kt
```diff
@@ -1,11 +1,30 @@
...
+import kotlinx.serialization.encoding.Encoder
 
-data class FeltArray(private val list: MutableList<Felt>) : ConvertibleToCalldata, MutableList<Felt> by list {
+data class FeltArray(private val list: MutableList<Felt>) : ConvertibleToCalldata, MutableList<Felt> by list, HttpBatchRequestType {
     constructor(vararg elements: Felt) : this(elements.toMutableList())
     constructor(collection: Collection<Felt>) : this(collection.toMutableList())
     constructor() : this(emptyList())
 
-    override fun toCalldata(): List<Felt> = list.toList()
+    override fun toCalldata(): List<Felt> = listOf(size.toFelt) + list
```
THenry14, line 15: this is not relevant to this PR, right?

lib/src/main/kotlin/com/swmansion/starknet/data/types/FeltArray.kt
```diff
@@ -1,11 +1,30 @@
...
+}
+
+object FeltArraySerializer : KSerializer<FeltArray> {
```
THenry14, line 18: also not sure about FeltArray, the name may be a bit vague. Maybe FeltList would be better?

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpBatchRequestOfDifferentTypes.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.swmansion.starknet.provider.Request
+import com.swmansion.starknet.provider.rpc.JsonRpcRequest
+import com.swmansion.starknet.provider.rpc.buildJsonHttpBatchDeserializerOfDifferentTypes
+import com.swmansion.starknet.service.http.HttpService
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+import java.util.concurrent.CompletableFuture
+
+class HttpBatchRequestOfDifferentTypes private constructor(
```
THenry14, line 13: I think there should be no distinction between batch reuest and barch request of different types. there should be one class with distinct constructors.

THenry14 (commented):
partial review

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpBatchRequestOfDifferentTypes.kt
franciszekjob: Your suggestion is reasonable, but we have to keep in mind that the main difference between these two is the type inference of corresponding function results. `HttpBatchRequest` has a type parameter, while `HttpBatchRequestOfDifferentTypes` does not. Because of that, `batchRequests` results will be fully inferred, while the other ones not. This lets for comfortable usage if we know that we will make requests only of one type.
Does that make more sense to you?

lib/src/main/kotlin/com/swmansion/starknet/data/types/FeltArray.kt
franciszekjob: `List<Felt>` was returned in some methods of `Provider` (eg. [here](https://github.com/software-mansion/starknet-jvm/blob/895dfc81eb51d0e15d8fa72c12f04596c8eb106c/lib/src/main/kotlin/com/swmansion/starknet/provider/Provider.kt#L157)) and I had to change it to a wrapper class. `FeltArray` has been already implemented before, so I decided it can be utilized and there's no need to create another wrapper. `FeltArray` didn't have serializer, that's why I've added it. Should I add separate wrapper, eg. `FeltList` and use it instead of existing `FeltArray`, wdyt @THenry14 ?

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/IntWrapperSerializer.kt
franciszekjob: Maybe `HttpBatchRequestString`? It's less general then current `IntWrapper`.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/StringWrapperSerializer.kt
franciszekjob: Wdyt of `HttpBatchRequestString`?

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/IntWrapperSerializer.kt
THenry14: This name is not intuitive, let's remember this is a type that we will return no matter if someone uses batched request or not. Perhaps something like `StarknetInt` would be a bit better?

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/StringWrapperSerializer.kt
THenry14: as per comment with Int

lib/src/main/kotlin/com/swmansion/starknet/data/types/FeltArray.kt
THenry14: ok, we can leave it as is then I suppose

lib/src/main/kotlin/com/swmansion/starknet/data/types/IntWrapper.kt
```diff
@@ -0,0 +1,8 @@
+package com.swmansion.starknet.data.types
+
+import com.swmansion.starknet.data.serializers.IntWrapperSerializer
+import kotlinx.serialization.Serializable
+
+@JvmInline
+@Serializable(with = IntWrapperSerializer::class)
+value class IntWrapper(val value: Int) : HttpBatchRequestType
```
THenry14, line 8: If I remember correctly, value classes are not interoperable with underlaying types? (so they are completely new types, unlike type aliases). How will that impact users? If they call a method that returns IntWrapper type, they will need to smehow convert it to int in order to do something more with it, right?

lib/src/main/kotlin/com/swmansion/starknet/data/types/StringWrapper.kt
```diff
@@ -0,0 +1,8 @@
+package com.swmansion.starknet.data.types
+
+import com.swmansion.starknet.data.serializers.StringWrapperSerializer
+import kotlinx.serialization.Serializable
+
+@JvmInline
+@Serializable(with = StringWrapperSerializer::class)
+value class StringWrapper(val value: String) : HttpBatchRequestType
```
THenry14, line 8: the same question as per IntWrapper value class

lib/src/main/kotlin/com/swmansion/starknet/data/types/Transaction.kt
```diff
@@ -382,7 +382,7 @@ data class DeclareTransactionV2(
     val contractDefinition: Cairo1ContractDefinition? = null,
 ) : DeclareTransaction(), DeprecatedTransaction {
     @Throws(ConvertingToPayloadFailedException::class)
-    internal fun toPayload(): DeclareTransactionV2Payload {
```
THenry14, line 385: why this function will be no longer internal, what's the rationale?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -57,36 +41,60 @@ internal fun <T> buildJsonHttpDeserializer(
...
+
+        val jsonResponses = Json.parseToJsonElement(response.body).jsonArray
+        val responses = jsonResponses.map {
+            deserializationJson.decodeFromJsonElement(
+                JsonRpcResponse.serializer(deserializationStrategies[it.jsonObject["id"]!!.jsonPrimitive.int]),
+                it,
             )
         }
 
+        val results = responses.sortedBy { it.id }.zip(jsonResponses)
```
THenry14, line 83: is there a possible performance hit with that?

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpBatchRequestOfDifferentTypes.kt
THenry14: I think there should be a way to merge those two, and it would be a lot simpler for users to just be able to rely on one class with multiple constructors

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/IntWrapperSerializer.kt
franciszekjob: Correct, `StarknetInt` seems better and is suitable to all types of requests.

lib/src/main/kotlin/com/swmansion/starknet/data/types/IntWrapper.kt
franciszekjob: In order to access integer value from `StarknetInt` wrapper (`IntWrapper` before), we have to access its `value`  property. For example having such function `getBlockNumber(): HttpRequest<StarknetInt>` we will use it this way:
```kotlin
val response: StarknetInt = provider.getBlockNumber().send()
val result: Int = response.value
```

lib/src/main/kotlin/com/swmansion/starknet/data/types/StringWrapper.kt
franciszekjob: Retrieving string value from `StarknetString` will be the same as from `StarknetInt` - we have to access `value` property.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Transaction.kt
franciszekjob: The change was unintentional and has been removed in the newest push.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
franciszekjob: I can surely prepare some performance tests on that. Wdyt?

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpBatchRequestOfDifferentTypes.kt
franciszekjob: `HttpBatchRequestOfDifferentTypes` has been removed in favor of enhancing simplicity. Two constructors have been introduced in `HttpBatchRequest` - one for  single and different types of requests respectively.
It caused a little issue with type erasure in JVM. I've introduced `containsMultipleRequestTypes` boolean flag . It helps to differentiate the overloaded constructors without altering the logic of our code, solving the possible signature clash.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
franciszekjob: Update: sorting has been completely removed as it was unnecessary. Correct assignment of deserializers to json responses is done by accessing element with adequate index from deserializers list - index is the id coming from json. Full implementation [here](https://github.com/software-mansion/starknet-jvm/blob/a8cb6a320165a1c53e389f90e30615e618f00623/lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt#L81).

franciszekjob: Update: I've improved sorting the results. Using a `MutableList` and directly placing items based on their ids will has better time complexity `O(n)` than using `sortedBy` which has `O(nlogn)`.
Implementation can be found [here](https://github.com/software-mansion/starknet-jvm/blob/ce6d87a1366371591ec02809287645833ed0e491/lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt#L104)
 Wdyt?

franciszekjob:
I've addresses all requested changes.

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpBatchRequest.kt
```diff
@@ -0,0 +1,65 @@
...
+    constructor(
+        url: String,
+        jsonRpcRequests: List<JsonRpcRequest>,
+        responseDeserializers: List<KSerializer<out T>>,
+        deserializationJson: Json,
+        service: HttpService,
+        //  The `containsDifferentRequestTypes` parameter is added to overcome a type erasure problem in JVM
+        //  avoiding a clash between similar constructor signatures with generic types.
+        //  Its value doesn't affect the logic.
+        containsDifferentRequestTypes: Boolean,
```
THenry14, line 45: @DelevoXDG please have a say on this. 
It's a bummer we cannot make do without this, but if this is the only way to have single class with multiple cnstructors for batches I guess it's fine. Maybe we could assign a default value to it?

lib/starknet-jvm.md
```diff
@@ -496,6 +499,83 @@ fun main(args: Array<String>) {
 }
 ```
 
+## Making multiple request calls of different types
```
THenry14, line 502:
```suggestion
## Making multiple calls of different types in one request
```
?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
```diff
@@ -61,49 +45,63 @@ internal fun <T> buildJsonHttpDeserializer(
...
-            orderedResults[id] = extractResult(jsonRpcResponse, response.body, jsonElement.toString())
+        if (jsonRpcResponse.result == null) {
+            throw RequestFailedException(message = "Response did not contain a result", payload = response.body)
         }
+        jsonRpcResponse.result
+    }
+}
 
-        orderedResults.filterNotNull()
+internal fun <T> processJsonRpcResponses(
```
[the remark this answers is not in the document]
franciszekjob: Does `getOrderedResults`  seems more descriptive?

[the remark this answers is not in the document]
THenry14: maybe something like `getOrderedRpcResponses`?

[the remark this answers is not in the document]
franciszekjob: `getOrderedRpcResponses` indicates that the output will be a collection of `JsonRpcResponses`, don't you think so?  This function returns instances of `Result`, so imho `getOrderedRpcResponses` may be also missleading. Wdyt?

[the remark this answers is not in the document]
THenry14: Hmm, right. so maybe `getOrderedRpcResults`?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/BuildJsonHttpDeserializer.kt
franciszekjob: Update: resolving this as the implementation ha been approved in upstream branch.

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpBatchRequest.kt
DelevoXDG: > It's a bummer we cannot make do without this

I don't think this should matter anyway, as we normally should only create create `HttpBatchRequest` using `JsonProvider` methods. We might as well make these constructors internal.

> Maybe we could assign a default value to it?

I don't think we can do that, this will lead to the same clash as before. Personally, I would rather use named factory methods instead of one or both of the constructors, similarly to how we deal with e.g. `Felt.fromSigned`. Imo this should be cleaner than using a flag can accept all kinds of values.

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpBatchRequest.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.swmansion.starknet.provider.rpc.JsonRpcRequest
+import com.swmansion.starknet.provider.rpc.buildJsonHttpBatchDeserializer
+import com.swmansion.starknet.provider.rpc.buildJsonHttpBatchDeserializerOfDifferentTypes
+import com.swmansion.starknet.service.http.HttpService
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+import java.util.concurrent.CompletableFuture
+
+class HttpBatchRequest<T> private constructor(
```
DelevoXDG, line 13: Why not 
```suggestion
class HttpBatchRequest<HttpBatchRequestType> private constructor(
```
or 
```suggestion
class HttpBatchRequest<T: HttpBatchRequestType> private constructor(
```

This applies not here specifically, but to every other places where `<T>` is still used. 

I think we might want to use some of that additional type safety that comes with introduction of `HttpBatchRequestType`, WDYT?

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpRequest.kt
```diff
@@ -20,7 +21,7 @@ class HttpRequest<T> private constructor(
     private val service: HttpService,
 ) : Request<T> {
 
-    internal constructor(
+    constructor(
```
DelevoXDG, line 24: Why make it public?

lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetString.kt
```diff
@@ -0,0 +1,8 @@
+package com.swmansion.starknet.data.types
+
+import com.swmansion.starknet.data.serializers.StarknetStringSerializer
+import kotlinx.serialization.Serializable
+
+@JvmInline
+@Serializable(with = StarknetStringSerializer::class)
+value class StarknetString(val value: String) : HttpBatchRequestType
```
DelevoXDG, line 8: This is fine I suppose, although someone might think `StarknetString` implies something from Cairo i.e. `ByteArray`. 

WDYT about `StringResponse`, `IntResponse`? 

I don't have a strong say on this, just a thought.

lib/src/main/kotlin/com/swmansion/starknet/data/types/HttpBatchRequestType.kt
```diff
@@ -0,0 +1,6 @@
+package com.swmansion.starknet.data.types
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+sealed interface HttpBatchRequestType
```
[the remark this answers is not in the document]
franciszekjob: > Why is this called HttpBatchRequestType?

`HttpBatchRequestType` is the common return type of `batchRequestsOfDifferentTypes`. It's name can be improved, do you have any suggestions?

> From what I understand, the same interface is used for singular requests as well?

No, in case of singular requests and batch requests of the same type we want to the types to be fully inferred. Therefore it's not used there.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -1230,25 +1230,49 @@ class ProviderTest {
...
+        val transaction = response[0].getOrThrow() as Transaction
+        val blockNumber = (response[1].getOrThrow() as StarknetInt).value
+        val txStatus = response[2].getOrThrow() as GetTransactionStatusResponse
+
+        assertEquals(transaction.hash, invokeTransactionHash)
+
+        assertNotEquals(0, blockNumber)
+
+        assertEquals(TransactionStatus.ACCEPTED_ON_L2, txStatus.finalityStatus)
+        assertNotNull(txStatus)
```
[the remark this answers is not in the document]
franciszekjob: I think `assertNotNull(txStatus)` should be completely removed, because of mentioned by you assertion below.

lib/src/test/kotlin/starknet/provider/response/JsonRpcResponseTest.kt
```diff
@@ -195,42 +193,91 @@ class JsonRpcResponseTest {
...
-        assertEquals(response[0].executionStatus, TransactionExecutionStatus.SUCCEEDED)
+        val txStatusResponse1 = response[0]
+        val txStatusResponse2 = response[1]
+
+        txStatusResponse1.getOrElse {
+            it as RpcRequestFailedException
+            assertEquals(-32603, it.code)
+            assertEquals("Internal error", it.message)
+            assertEquals("Invalid message selector", it.data)
+        }
```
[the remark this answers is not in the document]
franciszekjob: It seems to me that the logic here is valid. We have an error in mocked responses and in following lines we're handling an exception. It is thrown when accessing the result and to handle it we're using `getOrElse`.

What exactly seems unclear to you?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -73,7 +56,44 @@ class JsonRpcProvider(
...
     }
 
-    fun <T> batchRequests(requests: List<HttpRequest<T>>): BatchHttpRequest<T> {
+    /** Batch multiple HTTP requests (of different result type) together into a single HTTP request
+     *
+     * @param requests list of HTTP requests to be batched together
+     * @require list of requests
+     * @return batch HTTP request
+     */
+    fun batchRequestsOfDifferentTypes(requests: List<HttpRequest<out HttpBatchRequestType>>): HttpBatchRequest<HttpBatchRequestType> {
```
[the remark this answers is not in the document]
franciszekjob: I think it's better 👍

lib/src/test/kotlin/network/account/AccountTest.kt
```diff
@@ -99,7 +99,7 @@ class AccountTest {
             listOf(call),
             skipValidate = false,
         )
-        val estimateFeeResponse = estimateFeeRequest.send().first().overallFee
+        val estimateFeeResponse = estimateFeeRequest.send().values.first().overallFee
```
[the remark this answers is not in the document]
franciszekjob: Afaik there isn't a way to directly get value class's internal value without accessing it.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -73,7 +56,44 @@ class JsonRpcProvider(
         return HttpRequest(url, jsonRpcRequest, responseSerializer, deserializationJson, httpService)
     }
 
-    fun <T> batchRequests(requests: List<HttpRequest<T>>): BatchHttpRequest<T> {
+    /** Batch multiple HTTP requests (of different result type) together into a single HTTP request
+     *
+     * @param requests list of HTTP requests to be batched together
+     * @require list of requests
```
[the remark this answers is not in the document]
franciszekjob: Removed, it was unnecessary.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -73,7 +56,44 @@ class JsonRpcProvider(
         return HttpRequest(url, jsonRpcRequest, responseSerializer, deserializationJson, httpService)
     }
 
-    fun <T> batchRequests(requests: List<HttpRequest<T>>): BatchHttpRequest<T> {
+    /** Batch multiple HTTP requests (of different result type) together into a single HTTP request
```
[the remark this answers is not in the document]
franciszekjob: Sure, is `Batch multiple HTTP requests with JSON-RPC calls (of different result type) together into a single HTTP request`  enough?

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpBatchRequest.kt
franciszekjob: Current implementation keeps type safety for `batchRequests` and allows to use `batchRequestsAny` with all types which implement `HttpBatchRequestType`.  Why should we change it for some of the suggested solutions?

> I think we might want to use some of that additional type safety

Do you have something specific on your mind? As mentioned before, currently all types which implement `HttpBatchRequestType` can ba used in `batchRequestsAny`.

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpRequest.kt
franciszekjob: No need, I changed constructor in both `HttpRequest` and `HttpBatchRequest` to internal.

lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetString.kt
franciszekjob: Together with @THenry14 we had a [discussion](https://github.com/software-mansion/starknet-jvm/pull/461#discussion_r1592379466) about it 😅

Imho, I think that `StarknetString` is a bit far from being confused with `ByteArray`. 

`StringResponse` seems a bit better, but don't you think it might seem to general?

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpBatchRequest.kt
franciszekjob: Great idea, I added `fromRequests` and `fromRequestsAny` to `HttpBatchRequests`. This way those 2 constructors have been removed and the only one is now private (same as in `HttpRequest`).

lib/src/main/kotlin/com/swmansion/starknet/service/http/requests/HttpBatchRequest.kt
franciszekjob: I applied second suggestion in all occurences 👍 

> I think we might want to use some of that additional type safety

Do you have something specific on your mind? Currently `batchRequests` allows to send requests of only one type, `batchRequestsAny` allows to use all types which implement `HttpBatchRequestType`

lib/src/main/kotlin/com/swmansion/starknet/data/types/HttpBatchRequestType.kt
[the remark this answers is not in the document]
DelevoXDG: Why `HttpRequestType` though? Afaik all the classes that implement this interface are **Responses**, not **Requests** 😄

I would go for something like:
1. `Response`
2. `StarknetResponse`
3. `ResponseBase`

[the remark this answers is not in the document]
franciszekjob: Reasonable, let's choose `StarknetResponse` among all proposals - it's more appropriate 😄

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -73,7 +56,13 @@ class JsonRpcProvider(
         return HttpRequest(url, jsonRpcRequest, responseSerializer, deserializationJson, httpService)
     }
 
-    fun <T> batchRequests(requests: List<HttpRequest<T>>): BatchHttpRequest<T> {
+    /** Batch multiple HTTP requests with any JSON-RPC calls together into a single HTTP request
+     *
+     * @param requests list of HTTP requests to be batched together
+     *
+     * @return batch HTTP request
```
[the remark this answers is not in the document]
franciszekjob: Then maybe something like:
"Group multiple HTTP requests with any JSON-RPC calls together into a HTTP request with batched JSON-RPC calls"

wdyt?

lib/src/main/kotlin/com/swmansion/starknet/data/types/HttpBatchRequestType.kt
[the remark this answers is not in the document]
THenry14: Let's use StarknetResponse, as Response is a part of stdlib afaik

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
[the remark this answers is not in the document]
THenry14: let's not mention http at all, rather `Batch multiple calls into a single RPC request` ?

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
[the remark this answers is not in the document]
THenry14: https://github.com/software-mansion/starknet-jvm/pull/461#discussion_r1611206184

THenry14 (approved):

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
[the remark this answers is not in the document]
franciszekjob: Ok, just to emphasize that these can be any requests - is `Batch multiple (any) calls into a single RPC request` ok?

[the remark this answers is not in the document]
THenry14: maybe `Batch multiple various calls into a single RPC request` ?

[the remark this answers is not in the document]
franciszekjob: Seems good

PR #468, opened 2024-05-20
Install tools with asdf action

### LINKED ISSUE #365: Use asdf action to install scarb
### Feature Request

Can simplify `compileContracts.sh` script by installing scarb plugin on CI instead of using asdf action.

franciszekjob:
## Describe your changes
- Add `scarb` installation to CI
   - add `scarb` plugin with `asdf`
   - install versions: `0.4.0`, `0.7.0`, `2.6.0`

## Linked issues

Closes #365 

## Breaking changes


franciszekjob:
Moving the [discussion from #465 ](https://github.com/software-mansion/starknet-jvm/pull/465) here as the change does is related with both `starknet-foundry` and `scarb`.

lib/src/test/resources/compileContracts.sh
```diff
@@ -8,16 +8,9 @@ readonly V1_SCARB_VERSION="0.4.0"
 readonly V2_SCARB_VERSION="0.7.0"
 readonly V2_6_SCARB_VERSION="2.6.0"
 
-# TODO: (#365) Use asdf action to install scarb instead
-echo "Installing scarb..."
-asdf plugin add scarb || true
-echo "Done!"
```
THenry14, line 14: This script is also used to compile contracts locally, on the machine, right? So I suppose we should have a check here for all the versions, to make sure they are indeed installed

lib/src/test/resources/compileContracts.sh
```diff
@@ -8,16 +8,9 @@ readonly V1_SCARB_VERSION="0.4.0"
...
-
-
 echo "Compiling v1 contracts.."
 pushd "$(dirname "$0")" || exit 1
 pushd "$V1_CONTRACT_PATH" || exit 1
-asdf install scarb $V1_SCARB_VERSION || true
 asdf local scarb $V1_SCARB_VERSION || exit 1
```
THenry14, line 14: I wonder if we could use `asdf shell` instead? that way we would not create the .tool-versions in subdirectories, and still be able to use multiple versions across this script

lib/src/test/resources/compileContracts.sh
franciszekjob: Correct, so if we want to leave compile scripts as it is i.e. checking and installing scarb versions, then issue's description isn't applicable:

> Can simplify compileContracts.sh script by installing scarb plugin on CI instead of using asdf action.

THenry14: we do not need to do that, just a check for valid versions should suffice. if those versions are not present we can just error out with an appropriate message.

lib/src/test/resources/compileContracts.sh
franciszekjob: In order to use `asdf shell` in script we would have to source rc file in it.

THenry14: why?

franciszekjob: Because using `asdf shell` in a script without sourcing results in an error saying `Shell integration is not enabled. Please ensure you source asdf in your shell setup.` So we need to either set/source it.

THenry14: what is the effort to source this file, is it under the same path for both linux/macos and shells, or is the path different? Would it make sense to use paths to versions directly?

franciszekjob:
Discussed offline
1. We keep part of `compileContracts.sh` responsible for installing `scarb` and its versions, as it is handy for developers while testing SDK.
2. CI `scarb` installation is performed via `asdf` commands, mostly because we don't want to overcomplicate. 
Unfortunately, we cannot utilize [setup-scarb action](https://github.com/marketplace/actions/setup-scarb) as it doesn't allow to install multiple versions of scarb. asdf action is also not utilized because

lib/src/test/resources/compileContracts.sh
franciszekjob: No need to add extra plugin/version checks here as `asdf` perform checks on its own.

THenry14 (approved):

PR #469, opened 2024-05-22
Move `scarb` installation to CI

### LINKED ISSUE #365: Use asdf action to install scarb

franciszekjob:
## Describe your changes
- Add `scarb` installation to CI
   - add `scarb` plugin with `asdf`
   - install versions: `0.4.0`, `0.7.0`, `2.6.0`

## Linked issues

Closes #365 

## Breaking changes


franciszekjob:
Discussed offline
1. We keep part of `compileContracts.sh` responsible for installing `scarb` and its versions, as it is handy for developers while testing SDK. Also, there's no need to add extra checks whether version/plugin is already installed, as `asdf` performs these checks on its own.
2. CI `scarb` installation is performed via `asdf` commands, mostly because we don't want to overcomplicate versions management. This could occur while using asdf-action and having main `.tool-versions` and multiple `.tool-version` files spread across the project - in the top directory and in subdirectories with contracts.
Unfortunately, we cannot utilize [setup-scarb action](https://github.com/marketplace/actions/setup-scarb) as it doesn't allow to install multiple versions of scarb.

THenry14 (approved):

PR #478, opened 2024-05-27
Add `RequestResult` type

### LINKED ISSUE #477: Change `Result` type in batching requests methods
### Feature Request

Currently `JsonRpcProvider` methods for batching requests (`batchRequests` and `batchRequestsAny`) return values wrapped in kotlin's `Result`. It works in Kotlin but there is no support for it in `Java`, therefore we cannot access result values. `Result` needs to be changed to a custom type.

franciszekjob:
## Describe your changes

- Add custom type `RequestResult` which is a wrapper for kotlin's `Result`

## Linked issues

Closes #477 

## Breaking changes


lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetResult.kt
```diff
@@ -0,0 +1,28 @@
+package com.swmansion.starknet.data.types
+
+sealed class StarknetResult<out T> {
+
+    data class Success<T>(val value: T) : StarknetResult<T>()
+
+    data class Failure<T>(val exception: Exception) : StarknetResult<T>()
+
+    val isSuccessful: Boolean get() = this is Success<T>
```
DelevoXDG, line 9: Let's keep method naming the same as in [Kotlin Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/), otherwise it seems a bit confusing to me.

lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetResult.kt
```diff
@@ -0,0 +1,28 @@
+package com.swmansion.starknet.data.types
+
+sealed class StarknetResult<out T> {
+
+    data class Success<T>(val value: T) : StarknetResult<T>()
+
+    data class Failure<T>(val exception: Exception) : StarknetResult<T>()
```
DelevoXDG, line 7: Why not use Kotlin `Result` under the hood, and avoid implementing any logic ourselves?

lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetResult.kt
```diff
@@ -0,0 +1,28 @@
+package com.swmansion.starknet.data.types
+
+sealed class StarknetResult<out T> {
```
DelevoXDG, line 3: Nit:|

Not sure about this, but maybe `RequestResult`?

ddoktorski (approved):
LGTM, but let's wait for final approval from @DelevoXDG

lib/src/main/kotlin/com/swmansion/starknet/data/types/RequestResult.kt
```diff
@@ -0,0 +1,22 @@
+package com.swmansion.starknet.data.types
+
+data class RequestResult<out T>(val result: Result<T>) {
+    companion object {
+        fun <T> success(value: T): RequestResult<T> = RequestResult(Result.success(value))
+
+        fun <T> failure(throwable: Throwable): RequestResult<T> =
+            RequestResult(Result.failure(throwable))
```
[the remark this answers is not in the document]
franciszekjob: Actually in `failure` we have to keep `RequestResult<T>`, as there is not enough information to infer the type.

PR #484, opened 2024-06-03
Accept key as string in `Provider.getStorageAt()`

### LINKED ISSUE #333: `getStorageAt`: Allow passing `key` as String
Similarly to `Call`, where we can pass `entrypoint` as String, for consistency we should probably allow that for `getStorageAt`'s `key` and, possibly, for some other cases.

franciszekjob:
## Describe your changes

- Add `Provider.getStorageAt()` accepting `key` param as `String`
- Add `JsonRpcProvider.getStorageAt()` implementations
- Add `get storage at with key as string` test in `ProviderTest`
## Linked issues

Closes #333 

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/provider/Provider.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/provider/Provider.kt
@@ -1,5 +1,6 @@
 package com.swmansion.starknet.provider
 
+import com.swmansion.starknet.data.selectorFromName
 import com.swmansion.starknet.data.types.*
 import com.swmansion.starknet.provider.exceptions.RequestFailedException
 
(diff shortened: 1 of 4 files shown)


lib/src/main/kotlin/com/swmansion/starknet/provider/Provider.kt
```diff
@@ -247,6 +248,65 @@ interface Provider {
...
+     * Get a value of a storage variable of contract at the provided address.
+     *
+     * @param contractAddress an address of the contract
+     * @param key an address of the storage variable inside contract
+     * @param blockTag The tag of the requested block.
+     *
+     * @throws RequestFailedException
+     */
+    fun getStorageAt(contractAddress: Felt, key: String, blockTag: BlockTag): Request<Felt> {
+        return getStorageAt(contractAddress, selectorFromName(key), blockTag)
```
ddoktorski, line 263: Why do all the newly added methods contain implementations here? Considering this is an interface, shouldn't they be included in `JsonRpcProvided` instead?

franciszekjob: Initially I thought it could be implemented same as `signV3` in `Account` interface - that's the motivation why I chose to write it this way.
https://github.com/software-mansion/starknet-jvm/blob/c9df29a4e0c420f55211189f741a8c850b63c2cf/lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt#L116-L118
However, let's keep the actual implementation in the class 👍

ddoktorski (approved):

PR #485, opened 2024-06-03
Bump `sncast` version to `0.24.0`

### LINKED ISSUE #479: Bump `sncast` version
### Feature Request

Bump `sncast` version

franciszekjob:
## Describe your changes

- Bump `sncast` version to `0.24.0`
- Add `--hex-format` flag in `DevnetClient.runSnCast()`
- Use the previous approach to create and deploy test accounts

## Linked issues

Closes #479 

## Breaking changes


lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -36,6 +36,7 @@ class StandardAccountTest {
             accountDirectory = Paths.get("src/test/resources/accounts/standard_account_test"),
             contractsDirectory = Paths.get("src/test/resources/contracts"),
         )
+        private val accountName = "standard_account_test"
```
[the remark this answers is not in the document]
ddoktorski: I would prefer not to pass `accountName` as an argument to the `DevnetClient` constructor. Regarding `accountDirectory` and `contractsDirectory`, if they always have the same values, perhaps we could save them as constants and remove them from the parameters list?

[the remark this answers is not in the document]
franciszekjob: Sure 👍

lib/src/test/kotlin/starknet/utils/data/SnCastTypes.kt
```diff
@@ -57,7 +57,7 @@ data class AccountCreateSnCastResponse(
     val accountAddress: Felt,
 
     @JsonNames("max_fee")
-    val maxFee: Felt,
+    val maxFee: Long,
```
franciszekjob, line 60: `max_fee` is now returned as number, not as hex string.

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -390,14 +384,12 @@ class DevnetClient(
...
-        var result = String(process.inputStream.readAllBytes())
-
-        // TODO: remove this - pending sncast update
-        // As of sncast 0.6.0, "account create" outputs non-json data in the beggining of the response
-        val index = result.indexOf('{')
-        // Remove all characters before the first `{`
-        result = if (index >= 0) result.substring(index) else throw IllegalArgumentException("Invalid response JSON")
-
+        // As of sncast 0.24.0, declare command returns three response objects
+        // First two have 'status' field - 'compiling' and 'finished' respectively, and they don't have 'command' field
```
ddoktorski, line 388: FYI `compiling` and `finished` are actually a part of the `scarb build` output

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -28,6 +28,8 @@ import java.nio.file.Path
 import java.nio.file.Paths
 import kotlin.io.path.readText
 
+private const val ACCOUNT_NAME = "standard_account_test"
```
ddoktorski, line 31: Is there any reason we can't use the default account name `__default__`? If I understand correctly, we could remove `ACCOUNT_NAME` and its usages from all the files, and things should work as expected with the default account name.

lib/src/test/kotlin/starknet/utils/data/SnCastTypes.kt
ddoktorski: I think we can still represent it as hex string if we pass the `--hex-format` flag. However, this probably isn't necessary.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
franciszekjob: We can, I wanted to be consistent with previous approach, which assumed passing account names instead of using the default name.
e.g. of previous implementation in `ProviderTest`
https://github.com/software-mansion/starknet-jvm/blob/0ec0897cf9f85ae430d2d56f44c8d8b6dc7be18f/lib/src/test/kotlin/starknet/provider/ProviderTest.kt#L48

ddoktorski: It seems to me, based on the example you provided, that this deployment is used only to obtain the transaction hash of the `DeployAccount` transaction, which is later used in some tests. The execution of transactions (for example by `InvokeContract`) is still performed using the default account. We can confirm with @DelevoXDG to see if there were any other reasons for this approach. However, I believe we should do everything with the default account to simplify the code, as I don't see significant value in introducing a different account name for every individual test file.

PR #487, opened 2024-06-05
Add padded hex string

### LINKED ISSUE #486: Add method for creating padded `Felt`
### Feature Request

Add a method which returns `Felt` in padded format, e.g.
```kotlin
val felt = Felt.fromHex("0x01")
val feltString = felt.hexStringPadded();
// feltString = 0x0000000000000000000000000000000000000000000000000000000000000001
```

franciszekjob:
## Describe your changes

- Adjust `parseHex()` to accept padded hex string
- Add `Felt.hexStringPadded()`,
- Add appropriate tests to `NumAsHexBaseTests` and `FeltTests`

## Linked issues

Closes #486 

## Breaking changes


lib/src/main/kotlin/com/swmansion/starknet/data/ParseHex.kt
franciszekjob, line 1: In case of a padded hex string that represents `0`, all zeros will be trimmed, that's why there's an extra check for this edge case here.

lib/src/main/kotlin/com/swmansion/starknet/data/ParseHex.kt
```diff
@@ -9,5 +9,10 @@ internal fun parseHex(value: String): BigInteger {
     if (!value.startsWith("0x")) {
         throw IllegalArgumentException("Hex must start with 0x")
     }
-    return BigInteger(value.removePrefix("0x"), 16)
+
+    val trimmedValue = value.removePrefix("0x").trimStart('0')
+    if (trimmedValue.isEmpty()) {
+        return BigInteger.ZERO
+    }
+    return BigInteger(trimmedValue, 16)
```
franciszekjob, line 17: In case of a padded hex string that represents `0`, all zeros will be trimmed. After trimming, the value will be empty - that's why there's an extra check for this edge case.

franciszekjob: Correct, in fact we can pass the number with padding zeros to `BigInteger` and there's no need to remove them 👍

lib/src/main/kotlin/com/swmansion/starknet/extensions/ToHex.kt
```diff
@@ -4,3 +4,10 @@ import java.math.BigInteger
 
 @JvmSynthetic
 internal fun BigInteger.toHex(): String = "0x" + this.toString(16)
+
+@JvmSynthetic
+internal fun BigInteger.toHexPadded(): String {
+    val hexString = this.toString(16)
+    val zeros = "0".repeat(64 - hexString.length)
```
ddoktorski, line 11: What happens if `hexString.lenght` is greater than 64?

PR #489, opened 2024-06-07
Add `CairoVersion` enum

### LINKED ISSUE #447: Use enum for Account Cairo version
### Feature Request

Introduce cairo version enum that would be used in Account constructor instead of raw Felt

franciszekjob:
## Describe your changes

- Add `CairoVersion` enum with versions `0` and `1`

## Linked issues

Closes #447 

## Breaking changes

- Change `StandardAccount` constructor `cairoVersion` param from `Felt` to `CairoVersion`


lib/src/main/kotlin/com/swmansion/starknet/account/StandardAccount.kt
```diff
@@ -37,7 +37,7 @@ class StandardAccount @JvmOverloads constructor(
      * @param cairoVersion the version of Cairo language in which account contract is written
      */
     @JvmOverloads
-    constructor(address: Felt, privateKey: Felt, provider: Provider, chainId: StarknetChainId, cairoVersion: Felt = Felt.ZERO) : this(
+    constructor(address: Felt, privateKey: Felt, provider: Provider, chainId: StarknetChainId, cairoVersion: CairoVersion = CairoVersion.ZERO) : this(
```
ddoktorski, line 40: Not exactly related to this PR, but do you know why the default value is 0?

lib/src/main/kotlin/com/swmansion/starknet/data/types/CairoVersion.kt
```diff
@@ -0,0 +1,19 @@
+package com.swmansion.starknet.data.types
+
+import com.swmansion.starknet.extensions.toFelt
+
+enum class CairoVersion(val version: Felt) {
+    ZERO(Felt.ZERO), ONE(Felt.ONE);
```
ddoktorski, line 6: I am never sure if using only option 0 or 1 for the Cairo version is not confusing for users. Could we add a documentation stating that 0 corresponds to the legacy version and 1 corresponds to any contract compiled with Cairo >=1?

franciszekjob: Sure, sounds good to me.

ddoktorski (approved):

PR #490, opened 2024-06-10
Automatically determine `Cairo` version

### LINKED ISSUE #383: Automatically determine cairoVersion in StandardAccount
### Feature Request

We should consider automatically determining `cairoVersion` in `StandardAccount`, similarly to how it's done in starknet.py:
https://github.com/software-mansion/starknet.py/blob/1859b5d2d2be9963c2fa178567cc2ba82a282823/starknet_py/net/account/account.py

Currently, Cairo 0 is assumed as account contract version by default, and if someone doesn't know about this, the error reason `input too long for arguments`, sometimes encoded as a Felt, is not helpful at all.

franciszekjob:
## Describe your changes

- Add `ConstructorAbiEntry`
- Change `accountContractClassHash` in `DevnetClient` to Cairo 1

## Linked issues

Closes #383 

## Breaking changes

- Remove `cairoVersion` param from `StandardAccount` constructor


lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -795,16 +793,20 @@ class StandardAccountTest {
...
+                Felt(3),
                 Felt(0),
                 Felt(123),
                 selectorFromName("another_method"),
+                Felt(3),
                 Felt(2),
+                Felt(5),
+                Felt(10),
+                Felt(20),
+                Felt(30),
```
franciszekjob, line 809: `expectedCalldata` was changed here, because `cairoVersion` which was passed before (Cairo 1) is different than the one which is now automatically determined (Cairo 0). 
`expectedCalldata` scheme before was based on https://github.com/software-mansion/starknet-jvm/blob/f43b8b59674cb1a8347753413443a1836b9deea5/lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt#L98-L112
now it's based on 
https://github.com/software-mansion/starknet-jvm/blob/f43b8b59674cb1a8347753413443a1836b9deea5/lib/src/main/kotlin/com/swmansion/starknet/data/types/Execution.kt#L114-L133

ddoktorski: Would it be possible to use Cairo 1 account in this test instead?

lib/src/main/kotlin/com/swmansion/starknet/data/types/Contracts.kt
```diff
@@ -48,9 +48,20 @@ data class FunctionAbiEntry(
...
-    val type: AbiEntryType,
     val stateMutability: StateMutabilityType? = null,
-) : AbiElement()
+) : AbiElement() {
+    val type: AbiEntryType = AbiEntryType.FUNCTION
+}
+
+@SerialName("constructor")
+@Serializable
+data class ConstructorAbiEntry(
```
ddoktorski, line 58: Could you explain why this is included in this PR?

franciszekjob: In function `StandardAccount.ensureCairoVersion()` which determines Cairo version, there is a line 
```kotlin
val contract = provider.getClassAt(address).send()
``` 
It gives us a contract class definition, including its abi. Running some tests results in getting an abi containing entries with serial name `constructor`. It looks like such case didn't occur in any of our tests before, so it wasn't covered and adding `ConstructorAbiEntry` was omitted. Here is [CI with failure of mentioned tests](https://github.com/software-mansion/starknet-jvm/actions/runs/10005252217/job/27655577538).

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
franciszekjob: Yup, we can. In fact I've updated `accountContractClassHash` in `DevnetClient`, and this way we have a Cairo 1 account.

ddoktorski (approved):

PR #501, opened 2024-07-24
Add `StandardAccount.create()`

franciszekjob:
## Describe your changes
- Restore `StandardAccount` constructor with `cairoVersion` param
- Change default `cairoVersion` to `CairoVersion.ONE`
- Introduce static method `StandardAccount.create()`, which allows to create a `StandardAccount` instance with automatic determination of Cairo version

## Linked issues

Closes

## Breaking changes

### DIFF
--- a/lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
+++ b/lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
@@ -39,14 +39,16 @@ class StandardAccountTest {
         private val rpcUrl = devnetClient.rpcUrl
         private val provider = JsonRpcProvider(rpcUrl)
 
-        private val accountContractClassHash = DevnetClient.accountContractClassHashCairo1
+        private val accountContractClassHash = DevnetClient.accountContractClassHash
         private lateinit var accountAddress: Felt
+        private lateinit var legacyAccountAddress: Felt
         private lateinit var balanceContractAddress: Felt
 
         private lateinit var signer: Signer
 
         private lateinit var chainId: StarknetChainId
         private lateinit var account: Account
+        private lateinit var legacyAccount: Account
 
         @JvmStatic
         @BeforeAll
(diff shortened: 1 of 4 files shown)


lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -85,6 +86,16 @@ class StandardAccountTest {
...
+    @Test
+    fun `creating account with automatic Cairo version determination`() {
+        StandardAccount.create(
+            address = accountAddress,
+            privateKey = Felt(1234),
+            provider = provider,
+            chainId = chainId,
+        )
+    }
+
```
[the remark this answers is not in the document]
franciszekjob: Yes, I've added test both for Cairo 0 and 1.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -85,6 +85,76 @@ class StandardAccountTest {
         StandardAccount(Felt.ZERO, privateKey, provider, chainId)
     }
 
+    @Test
+    fun `creating cairo 0 account with automatic version determination`() {
```
ddoktorski, line 89: nit:
```suggestion
    fun `cairo 0 account with automatic version detection`() {
```
The same suggestion below

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -85,6 +85,76 @@ class StandardAccountTest {
...
+        val call = Call(
+            contractAddress = balanceContractAddress,
+            entrypoint = "increase_balance",
+            calldata = listOf(Felt(10), Felt(20), Felt(30)),
+        )
+        val accountDetailsCairo0 = devnetClient.createDeployAccount(classHash = DevnetClient.accountContractClassHashCairo0, accountName = "cairo0_account").details
```
ddoktorski, line 95: Let's move it to the `before` function and save the address to a field `legacyAccountAddress`.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -85,6 +85,76 @@ class StandardAccountTest {
...
+        val accountAddressCairo0 = accountDetailsCairo0.address
+        val signerCairo0 = StarkCurveSigner(accountDetailsCairo0.privateKey)
```
ddoktorski, line 97: It doesn't matter for this test what signer will be used. We can simply reuse the default `signer` or pass any value as a private key.

franciszekjob: Good remark 👍

ddoktorski (approved):

PR #509, opened 2024-08-12
Merge transaction and transaction payload classes

### LINKED ISSUE #394: Consider merging Transaction and TransactionPayload classes
### Feature Request

Could use a different serializer for sending Transaction classes as a payload. That would simplify updating transaction dataclasses in the future.

franciszekjob:
## Describe your changes

- Remove classes representing transaction payload to get rid of redundancy between them and  the classes explicitly representing transactions
- Add `TransactionSerializer`
- Add custom serializer for declare v2/v3 transactions
- Add `ExecutableTransaction` interface and `ExecutableTransactionSerializer`

## Linked issues

Closes #394 

## Breaking changes

- Use transaction classes instead of payloads
- `Provider`, `JsonRpcProvider`:
   - `getEstimateFee` now accepts `payload` as list of `ExecutableTransaction` instead of `TransactionPayload`
   - `simulateTransactions` now accepts `transactions` as list of `ExecutableTransaction` instead of `TransactionPayload`


lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionSerializer.kt
```diff
@@ -0,0 +1,87 @@
...
+        val typeElement = jsonElement.getOrElse("type") { throw SerializationException("Input element does not contain mandatory field 'type'") }
+
+        val type = decoder.json.decodeFromJsonElement(TransactionType.serializer(), typeElement)
+
+        return when (type) {
+            TransactionType.INVOKE -> deserializeInvoke(decoder, element)
+            TransactionType.DECLARE -> deserializeDeclare(decoder, element)
+            TransactionType.DEPLOY_ACCOUNT -> deserializeDeployAccount(decoder, element)
+            else -> throw IllegalArgumentException("Invalid transaction type '${typeElement.jsonPrimitive.content}'")
+        }
```
DelevoXDG, line 28: This seems odd. How should we deserialize other tx types, e.g. `L1_HANDLER`?

If I remember, we didn't use custom deserialization for `Transaction` classes altogether, but if that's something that's necessary, I think it should handle all cases.

(tbh I'm not sure why all tests pass 😅)

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionSerializer.kt
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private fun deserializeDeclare(decoder: JsonDecoder, element: JsonElement): DeclareTransaction {
+        val versionElement = element.jsonObject.getOrElse("version") { throw SerializationException("Input element does not contain mandatory field 'version'") }
+
+        val version = decoder.json.decodeFromJsonElement(Felt.serializer(), versionElement)
+        return when (version) {
+            Felt(3) -> decoder.json.decodeFromJsonElement(DeclareTransactionV3.serializer(), element)
+            Felt(2) -> decoder.json.decodeFromJsonElement(DeclareTransactionV2.serializer(), element)
+            else -> throw IllegalArgumentException("Invalid declare transaction version '${versionElement.jsonPrimitive.content}'")
```
DelevoXDG, line 73: I think we weren't supposed to be able to deserialize declare transaction **payloads** of version `V0` and `V1`. 

But since we're dealing with `Transaction`s here, not `TransactionPayloads`, I think we are supposed to be able to deserialize historic `V1` and `V0` transactions.

DelevoXDG: The same applies to other type-specific deserializers

lib/src/main/kotlin/com/swmansion/starknet/provider/Provider.kt
```diff
@@ -841,7 +841,7 @@ interface Provider {
      * @param simulationFlags set of flags to be used for simulation
      * @return a list of transaction simulations
      */
-    fun simulateTransactions(transactions: List<TransactionPayload>, blockTag: BlockTag, simulationFlags: Set<SimulationFlag>): Request<SimulatedTransactionList>
+    fun simulateTransactions(transactions: List<Transaction>, blockTag: BlockTag, simulationFlags: Set<SimulationFlag>): Request<SimulatedTransactionList>
```
DelevoXDG, line 844: I don't think we're supposed to be able to simulate / estimate fees / etc. for **every** type of transaction. As for example, there are outdated `V0`, `V1` transactions, that we should able to fetch from chain, but not execute. 

I think we should have an interface that determines if transaction is **executable** (can be treated as a payload and sent for execution), similarly to how it's done in swift, so transactions that are not meant to be executed are not passed in these methods.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
```diff
@@ -153,17 +153,17 @@ class JsonRpcProvider(
         return callContract(call, BlockTag.LATEST)
     }
 
-    override fun deployAccount(payload: DeployAccountTransactionV1Payload): HttpRequest<DeployAccountResponse> {
-        val params = jsonWithDefaults.encodeToJsonElement(payload)
+    override fun deployAccount(payload: DeployAccountTransactionV1): HttpRequest<DeployAccountResponse> {
+        val params = jsonWithDefaults.encodeToJsonElement(TransactionSerializer, payload)
```
DelevoXDG, line 157: Nit: 

Please check this, but afaik this shouldn't be necessary if you previously annotated `Transaction` class with `@Serializable(with = ...)`

DelevoXDG (changes requested):

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionSerializer.kt
franciszekjob: Right, I've added `L1_HANDLER ` and `DEPLOY`.

lib/src/main/kotlin/com/swmansion/starknet/provider/rpc/JsonRpcProvider.kt
franciszekjob: I didn't annotate transaction classes with `@Serializable(with = ...)` because it would create a recursive calls (`serialize` in `TransactionSerializer` would use serializer of specific transaction, which would then go back to `TransactionSerializer` and so on).

Also, transaction classes which inherit from `Transaction` aren't automatically serialized with `TransactionSerializer` even though there's such annotation (not sure why tbh):
```kotlin
@Serializable(with = TransactionSerializer::class)
sealed class Transaction : StarknetResponse {
```

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionSerializer.kt
```diff
@@ -0,0 +1,91 @@
...
+            TransactionType.DEPLOY -> Json.decodeFromJsonElement(DeployTransaction.serializer(), element)
+            TransactionType.L1_HANDLER -> Json.decodeFromJsonElement(L1HandlerTransaction.serializer(), element)
+        }
+    }
+
+    override fun serialize(encoder: Encoder, value: Transaction) {
+        require(encoder is JsonEncoder)
+
+        val jsonObject = when (value) {
+            is InvokeTransactionV3 -> Json.encodeToJsonElement(InvokeTransactionV3.serializer(), value).jsonObject
```
DelevoXDG, line 36: I think we should be using `encoder.json.encodeToJsonElement` instead of `Json.encodeToJsonElement`, to use instance provided by the calling `encoder`, so if there are any extra settings defined at the higher level, we don't override those by using `Json` instance with default ones.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/DeclareTransactionV3Serializer.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.*
+
+internal object DeclareTransactionV3Serializer : KSerializer<DeclareTransactionV3> {
+    override val descriptor = PrimitiveSerialDescriptor("DeclareTransactionV3", PrimitiveKind.STRING)
+
+    override fun serialize(encoder: Encoder, value: DeclareTransactionV3) {
```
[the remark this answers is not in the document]
franciszekjob: I left custom serializers only for declare txs, rest was removed.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionSerializer.kt
```diff
@@ -0,0 +1,92 @@
...
+import com.swmansion.starknet.data.types.*
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.*
+
+internal val transactionIgnoredKeys = listOf("transaction_hash", "contract_address")
```
franciszekjob, line 12: These fields are not in RPC spec, so we remove them before while serializing.

lib/src/main/kotlin/com/swmansion/starknet/provider/Provider.kt
franciszekjob: I added `ExecutableTransaction` interface and used in in `simulateTransaction` and `getEstimateFee` in provider.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionSerializer.kt
THenry14: are they returned by nodes? or what is the reason for this disparity?

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionSerializer.kt
```diff
@@ -0,0 +1,90 @@
...
+            "TransactionSerializer can only serialize ExecutableTransaction instances."
+        }
+        val jsonObject = when (value) {
+            is InvokeTransactionV3 -> encoder.json.encodeToJsonElement(ExecutableTransactionSerializer, value).jsonObject
+            is InvokeTransactionV1 -> encoder.json.encodeToJsonElement(ExecutableTransactionSerializer, value).jsonObject
+            is DeclareTransactionV3 -> encoder.json.encodeToJsonElement(ExecutableTransactionSerializer, value).jsonObject
+            is DeclareTransactionV2 -> encoder.json.encodeToJsonElement(ExecutableTransactionSerializer, value).jsonObject
+            is DeclareTransactionV1 -> encoder.json.encodeToJsonElement(ExecutableTransactionSerializer, value).jsonObject
+            is DeployAccountTransactionV3 -> encoder.json.encodeToJsonElement(ExecutableTransactionSerializer, value).jsonObject
+            is DeployAccountTransactionV1 -> encoder.json.encodeToJsonElement(ExecutableTransactionSerializer, value).jsonObject
```
THenry14, line 46: this looks like it could maybe be refactored?

THenry14 (commented):
all in all looks good. a few nitpicks.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionSerializer.kt
franciszekjob: You can see in `Transaction.kt` file that there are comments about these fields (e.g. [here](https://github.com/software-mansion/starknet-jvm/blob/7b712699a626fc51e9ce665dbb539c57d86c798a/lib/src/main/kotlin/com/swmansion/starknet/data/types/Transaction.kt#L694)). I'm not sure about the specific reason for this disparity, but these fields **are clearly not in the spec.**

E.g., here's deploy account v1 as [kotlin dataclass](https://github.com/software-mansion/starknet-jvm/blob/7b712699a626fc51e9ce665dbb539c57d86c798a/lib/src/main/kotlin/com/swmansion/starknet/data/types/Transaction.kt#L757) and as [spec schema](https://github.com/software-mansion/starknet-jvm/blob/7b712699a626fc51e9ce665dbb539c57d86c798a/lib/src/main/kotlin/com/swmansion/starknet/data/types/Transaction.kt#L757), you can see the difference.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionSerializer.kt
THenry14: What is an overhead for this refactor? Have you considered getting rid of when altogether?
I *think* you could just define a sealed class and make it a parent for all those classes. Then you could have just something like:

```
val jsonObject = encoder.json.encodeToJsonElement(ExecutableTransactionSerializer, value).jsonObject
```

where ExecutableTransaction is your parent class.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionSerializer.kt
franciszekjob: > if they are missing from response, do not fail but just set them to something generic like null or something

Sure, I made contract address optional in `DeployAccountTransaction` + its inheriting classes and in `L1HandlerTransaction`. In `InvokeTransactionV0` it should stay according to spec. Transaction hash is already optional.

> let's reach out to relevant teams to ask about this disparity (maybe those fields should be added to the spec? or maybe nodes should remove them?)

Sure, will contact Ariel

> btw, is this for all the nodes? or is this only for either one of pathfinder/juno?

I know this happens on devnet's node and juno's node (the one that we utilize internally).

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/TransactionSerializer.kt
franciszekjob: :eyes: oops, correct - we can simplify it this way.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Transaction.kt
```diff
@@ -739,8 +697,9 @@ sealed class DeployAccountTransaction : Transaction() {
     @SerialName("class_hash")
     abstract val classHash: Felt
 
+    // not in RPC spec
     @SerialName("contract_address")
-    abstract val contractAddress: Felt
+    abstract val contractAddress: Felt?
```
THenry14, line 702:
```suggestion
    abstract val contractAddress: Felt? = Felt.ZERO,
```

I think this should probably have a default value?

franciszekjob: It's an abstract field - it cannot have default value at this level, its default value is provided in deploy account v1/v3.

THenry14: Ahh yes, sorry I am dzban

THenry14 (approved):

PR #517, opened 2024-09-11
Add `StarknetByteArray.toString`

### LINKED ISSUE #515: Add parsing method for `StarknetByteArray`
### Feature Request

Add a method which allows to parse `StarknetByteArray` to string.
Ref: https://docs.starknet.io/architecture-and-concepts/smart-contracts/serialization-of-cairo-types/#serialization_of_byte_arrays

franciszekjob:
## Describe your changes

Add `StarknetByteArray.toString()` method which parses byt array into its input.

## Linked issues

Closes #515 

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetByteArray.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetByteArray.kt
@@ -26,7 +26,8 @@ data class StarknetByteArray(
             "The length of 'pendingWord' must be between 0 and 30. [$pendingWordLen] given."
         }
         // We skip the edge case when pending word is a null character, because its pendingWord.byteLength is 0 and pendingWordLen is 1.
-        if (!(pendingWord == Felt.ZERO && pendingWordLen == 1)) {
+        val isPendingWordNullChar = pendingWord == Felt.ZERO && pendingWordLen == 1
+        if (!isPendingWordNullChar) {
             require(pendingWord.byteLength == pendingWordLen) {
                 "The length of 'pendingWord' must be equal to 'pendingWordLen'. [${pendingWord.hexString()}] of length [${pendingWord.byteLength}] given."
             }
--- a/lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetByteArray.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetByteArray.kt
@@ -28,7 +28,7 @@ data class StarknetByteArray(
         require(pendingWordLen in 0..30) {
             "The length of 'pendingWord' must be between 0 and 30. [$pendingWordLen] given."
         }
-        // We skip the edge case when pending word is a null character, because its pendingWord.byteLength is 0 and pendingWordLen is 1.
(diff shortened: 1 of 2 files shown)


lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetByteArray.kt
```diff
@@ -37,6 +37,14 @@ data class StarknetByteArray(
         return listOf(data.size.toFelt) + data + listOf(pendingWord, pendingWordLen.toFelt)
     }
 
+    /**
+     * Encode as a String
+     */
+    override fun toString(): String {
+        val shortStrings = data.map { it.toShortString() } + pendingWord.toShortString()
+        return shortStrings.joinToString(separator = "").replace("\u0000", "")
```
franciszekjob, line 45: note: In case when `pendingWord` is equal to 0, it gets serialized to [null char](https://en.wikipedia.org/wiki/Null_character). To get rid of this, we simply replace it.

THenry14: Technicallt - is it possible that null char can be an element of an input string? should we then do this?

franciszekjob: Yup, e.g. following code
```kotlin
println(0.toChar() + "otherString")
```
produces such output
```
otherString
```

Without `replace("\u0000", "")`, two of our test cases fail, when `pendingWord = Felt.ZERO`.

THenry14: I was rather wondering what happens if the input string is something like this: 

"This is my string: \u0000" 

or just 

"\z"

?

franciszekjob: Such code
```kotlin
val ba1 = StarknetByteArray.fromString("This is my string: \u0000")
println(ba1.toString())

val ba2 = StarknetByteArray.fromString("\\z")
println(ba2.toString())
```
produces
```
This is my string: 
\z
```

so it works as expected.

Btw, I've found a bug in the existing `StarknetByteArray.fromString()`. When we use:
```kotlin
StarknetByteArray.fromString("This is my string, I like it!: \u0000")
```
we get an error
```
java.lang.IllegalArgumentException: The length of 'pendingWord' must be equal to 'pendingWordLen'. [0x0] of length [0] given.
```

It's because `\u0000` is represented by `0` i.e. it's bit length is also equal to `0`.
Looks like we should handle this case in [this line](https://github.com/software-mansion/starknet-jvm/blob/feat/515-parse-starknet-byte-array/lib/src/main/kotlin/com/swmansion/starknet/data/types/StarknetByteArray.kt#L28-L30).
Wdyt @THenry14?

franciszekjob: I've added null character case handling and updated test cases with.

THenry14 (approved):
lgtm

PR #530, opened 2024-10-22
Feat/519 rpc 0.8.0

franciszekjob:
## Linked issues

Closes

## Breaking changes


lib/src/main/kotlin/com/swmansion/starknet/provider/Provider.kt
```diff
@@ -651,6 +662,19 @@ interface Provider {
...
+     * Get merkle paths in one of the state tries: global state, classes, individual contract.
+     *
+     * @param classHashes list of class hashes for which we want to prove membership
+     * @param contractAddresses list of contract addresses for which we want to prove membership
+     * @param contractsStorageKeys list of contract address and storage keys pairs
+     *
+     * @throws RequestFailedException
+     */
+    fun getStorageProof(classHashes: List<Felt>? = null, contractAddresses: List<Felt>? = null, contractsStorageKeys: List<ContractStorageKey>? = null): Request<StorageProof>
+
```
franciszekjob, line 677: note: It's a bit exceptional situation, due to the fact that creating overloads would cause a conflict, e.g. for
```kotlin
fun getStorageProof(classHashes: List<Felt>, contractsStorageKeys: List<ContractStorageKey>): Request<StorageProof>
```
and
```kotlin
fun getStorageProof(contractAddresses: List<Felt>, contractsStorageKeys: List<ContractStorageKey>): Request<StorageProof>
```
(compiler won't know which method to use, `contractAddresses` and `classHashes` are of the same type). So unfortunately we need to add default values here.

franciszekjob:
note:
1. Tests will be addressed once devnet is not a blocker abymore.
2. Full description about included changes will be added later.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -129,8 +148,101 @@ data class GetTransactionStatusResponse(
...
+)
+
+@Serializable
+data class ChildrenHashes(
+    @SerialName("left")
+    val left: Felt,
+
+    @SerialName("right")
+    val right: Felt,
+)
```
DelevoXDG, line 217: I think we should call it `MerkleNodeChildrenHashes`, or even better make it a nested class within `MerkleNode`.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -129,8 +148,101 @@ data class GetTransactionStatusResponse(
...
+data class GlobalRoots(
+    @SerialName("contracts_tree_root")
+    val contractsTreeRoot: Felt,
+
+    @SerialName("classes_tree_root")
+    val classesTreeRoot: Felt,
+
+    @SerialName("block_hash")
+    val blockHash: Felt,
+)
```
DelevoXDG, line 193: Same as `ChildrenHashes`

lib/src/main/kotlin/com/swmansion/starknet/data/types/SimulatedTransaction.kt
```diff
@@ -65,7 +65,7 @@ data class FunctionInvocation(
     val messages: List<OrderedMessageL2ToL1>,
 
     @SerialName("execution_resources")
-    val computationResources: ComputationResources,
+    val computationResources: ExecutionResources,
```
DelevoXDG, line 68: In the spec this still points to non-existent `COMPUTATION_RESOURCES`:
https://github.com/starkware-libs/starknet-specs/blob/647caa00c0223e1daab1b2f3acc4e613ba2138aa/api/starknet_trace_api_openrpc.json#L448

Probably something we should clarify

lib/src/main/kotlin/com/swmansion/starknet/data/types/TransactionReceipt.kt
```diff
@@ -107,6 +108,9 @@ data class InvokeTransactionReceipt private constructor(
     @SerialName("finality_status")
     override val finalityStatus: TransactionFinalityStatus,
 
+    @SerialName("failure_reason")
+    override val failureReason: String? = null,
+
```
DelevoXDG, line 113: Correct me if I'm wrong, but from what I gather from the spec `failure_reason` should only be available in `starknet_getTransactionStatus` and `starknet_getMessagesStatus` responses, not receipts.

lib/src/main/kotlin/com/swmansion/starknet/provider/Provider.kt
```diff
@@ -321,6 +321,17 @@ interface Provider {
...
+    /**
+     * Get L1 handler transaction data.
+     *
+     * Get L1 handler transaction data for all L1 → L2 messages sent by the given L1 transaction.
+     *
+     * @param l1TransactionHash The hash of the L1 transaction.
+     *
+     * @throws RequestFailedException
+     */
+    fun getMessagesStatus(l1TransactionHash: NumAsHex): Request<GetMessagesStatusResponse>
```
DelevoXDG, line 333: This should return an array:
https://github.com/starkware-libs/starknet-specs/blob/647caa00c0223e1daab1b2f3acc4e613ba2138aa/api/starknet_api_openrpc.json#L269

lib/src/main/kotlin/com/swmansion/starknet/data/types/Payloads.kt
```diff
@@ -83,6 +89,21 @@ internal data class GetNoncePayload(
     override val blockId: BlockId,
 ) : PayloadWithBlockId()
 
+@Serializable
+internal data class GetStorageProofPayload @JvmOverloads constructor(
+    @SerialName("block_id")
+    val blockId: BlockId,
```
DelevoXDG, line 95: I don't think `BlockId` is needed:
https://github.com/starkware-libs/starknet-specs/blob/647caa00c0223e1daab1b2f3acc4e613ba2138aa/api/starknet_api_openrpc.json#L910

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -79,9 +88,11 @@ data class EstimateFeeResponse(
...
      * Convert estimated fee to resource bounds with applied multipliers.
      *
-     * Calculates max amount as maxAmount = [overallFee] / [gasPrice], unless [gasPrice] is 0, then maxAmount is 0.
-     * Calculates max price per unit as maxPricePerUnit = [gasPrice].
-     * Then multiplies maxAmount by round([amountMultiplier] * 100%) and maxPricePerUnit by round([unitPriceMultiplier] * 100%) and performs integer division by 100 on both.
+     * Calculates max amount l1 as maxAmountL1 = [overallFee] / [l1GasPrice], unless [l1GasPrice] is 0, then maxAmountL1 is 0.
+     * Calculates max amount l2 as maxAmountL2 = [overallFee] / [l2GasPrice], unless [l2GasPrice] is 0, then maxAmountL2 is 0.
+     * Calculates max price per unit l1 as maxPricePerUnitL1 = [l1GasPrice].
+     * Calculates max price per unit l2 as maxPricePerUnitL2 = [l2GasPrice].
+     * Then multiplies maxAmountL1/L2 by round([amountMultiplier] * 100%) and maxPricePerUnitL1/L2 by round([unitPriceMultiplier] * 100%) and performs integer division by 100 on both.
```
DelevoXDG, line 95: Nit:
```suggestion
     * Then multiplies maxAmountL1/L2 by round([amountMultiplier] * 100%) and maxPricePerUnitL1/L2 by round([unitPriceMultiplier] * 100%) and performs integer division by 100 on each.
```

lib/src/main/kotlin/com/swmansion/starknet/data/types/Payloads.kt
franciszekjob: It is, some time ago they added a PR updating storage proof (sorry, I didn't link it in the PR description)
https://github.com/starkware-libs/starknet-specs/pull/232

lib/src/main/kotlin/com/swmansion/starknet/data/types/Payloads.kt
```diff
@@ -83,6 +89,21 @@ internal data class GetNoncePayload(
     override val blockId: BlockId,
 ) : PayloadWithBlockId()
 
+@Serializable
+internal data class GetStorageProofPayload @JvmOverloads constructor(
```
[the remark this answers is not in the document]
franciszekjob: Correct 😅

lib/src/main/kotlin/com/swmansion/starknet/data/types/SimulatedTransaction.kt
franciszekjob: There isn't `COMPUTATION_RESOURCES` schema anymore
https://github.com/starkware-libs/starknet-specs/blob/8cf463b79ba1dd876f67c7f637e5ea48beb07b5b/api/starknet_trace_api_openrpc.json#L465

franciszekjob:
Please also check out the [updated spec](https://github.com/starkware-libs/starknet-specs/blob/v0.8.0/api/starknet_api_openrpc.json), while reviewing. There were some changes so the listed PRs may not be fully up to date.

lib/src/main/kotlin/com/swmansion/starknet/data/types/TransactionReceipt.kt
franciszekjob: > Correct me if I'm wrong, but from what I gather from the spec failure_reason should only be available in `starknet_getTransactionStatus`

Yes, you're absolutely correct. It was added to receipt classes by mistake 😅 . 

> and `starknet_getMessagesStatus` responses

I don't see that `failure_reason` should be included in `starknet_getMessagesStatus`? https://github.com/starkware-libs/starknet-specs/blob/v0.8.0/api/starknet_api_openrpc.json#L266-L290

lib/src/main/kotlin/com/swmansion/starknet/account/Account.kt
```diff
@@ -188,18 +188,22 @@ interface Account {
...
+     * @param l1ResourceBounds L1 resource bounds for the transaction
+     * @param l2ResourceBounds L2 resource bounds for the transaction
      * @return signed deploy account payload
      */
     fun signDeployAccountV3(
         classHash: Felt,
         calldata: Calldata,
         salt: Felt,
         l1ResourceBounds: ResourceBounds,
+        l2ResourceBounds: ResourceBounds,
```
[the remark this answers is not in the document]
franciszekjob: Definitely, makes sense.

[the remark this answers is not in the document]
DelevoXDG: > I just don't fully like the fact that now the param of type `ResourceBoundsMapping` is named `resourceBounds` 😆 , but looks like it's because of [this](https://github.com/starkware-libs/starknet-specs/blob/8cf463b79ba1dd876f67c7f637e5ea48beb07b5b/api/starknet_api_openrpc.json#L2114), so we can leave it.

That was the idea, but also the fact that `resourceBoundsMapping` doesn't seem like a good argument name from UX point of view.

I think such naming convention also makes sense in this particular case:
- `ResourceBounds` - `l1ResourceBounds`, `l2ResourceBounds`
- `ResourceBoundsMapping` - `resourceBounds`

lib/src/main/kotlin/com/swmansion/starknet/data/types/SimulatedTransaction.kt
DelevoXDG: The name's also changed to `Execution resources` there. 

I wonder if we should keep try to keep the old name to avoid breaking change, or do it earlier than later, since the spec is expected to introduce some breaking changes anyways?

lib/src/main/kotlin/com/swmansion/starknet/account/StandardAccount.kt
```diff
@@ -541,7 +541,10 @@ class StandardAccount @JvmOverloads constructor(
     private fun buildEstimateFeeV3Payload(calls: List<Call>, nonce: Felt): List<ExecutableTransaction> {
         val executionParams = InvokeParamsV3(
             nonce = nonce,
-            l1ResourceBounds = ResourceBounds.ZERO,
+            resourceBounds = ResourceBoundsMapping(
+                ResourceBounds.ZERO,
+                ResourceBounds.ZERO,
+            ),
```
DelevoXDG, line 547: Nit:

Should we introduce `ResourceBoundsMapping.ZERO`? 🤔

lib/src/main/kotlin/com/swmansion/starknet/data/types/Payloads.kt
DelevoXDG: Got it, was just checking out the changes from the tag, not branch 😅

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -129,8 +147,101 @@ data class GetTransactionStatusResponse(
...
+@Serializable
+data class NodeHashToNodeMappingItem(
+    @SerialName("node_hash")
+    val nodeHash: Felt,
+
+    @SerialName("node")
+    val node: MerkleNode,
+) {
+    @Serializable
+    sealed interface MerkleNode
```
DelevoXDG, line 221: Are you sure that works?

Seems to me like some custom serializer is needed for `node` 🤔

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -343,21 +365,24 @@ class StandardAccountTest {
...
     }
 
     @Test
     fun estimateMessageFee() {
-        val l1l2ContractCode = Path.of("src/test/resources/contracts_v2/target/release/ContractsV2_l1_l2.sierra.json").readText()
-        val l1l2CasmContractCode = Path.of("src/test/resources/contracts_v2/target/release/ContractsV2_l1_l2.casm.json").readText()
+        val l1l2ContractCode =
+            Path.of("src/test/resources/contracts_v2/target/release/ContractsV2_l1_l2.sierra.json").readText()
+        val l1l2CasmContractCode =
+            Path.of("src/test/resources/contracts_v2/target/release/ContractsV2_l1_l2.casm.json").readText()
```
DelevoXDG, line 385: Just to clarify, are these changes necessary (e.g. dictated by linting)? 👀

DelevoXDG (commented):
I like where this is going 👍🏻 Waiting for the tests

lib/src/main/kotlin/com/swmansion/starknet/account/StandardAccount.kt
franciszekjob: I was also thinking about this, should be handy 👍 .

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
franciszekjob: Oops, not needed. Removed 🫡 .

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
franciszekjob: I added missing `MerkleNodeSerializer` 👍 .

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/MerkleNodeSerializer.kt
```diff
@@ -0,0 +1,42 @@
...
+        return when {
+            binaryNodeKeys.all { jsonElement.containsKey(it) } -> {
+                decoder.json.decodeFromJsonElement(NodeHashToNodeMappingItem.BinaryNode.serializer(), element)
+            }
+            edgeNodeKeys.all { jsonElement.containsKey(it) } -> {
+                decoder.json.decodeFromJsonElement(NodeHashToNodeMappingItem.EdgeNode.serializer(), element)
+            }
+            else -> throw IllegalArgumentException("Invalid MerkleNode JSON object: $jsonElement")
+        }
+    }
```
franciszekjob, line 31: note: We could potentially give an error with more meaningful message (saying which fields are missing and if it's binary/edge node). However, in case of json with keys that belong both to binary and edge node
e.g.
```
{
    "left": ...,
    "path": ...
}
```
we can't explicitly decide which type it is, i.e. we can't give type-specific error. I'm not fully sure if such logic is worth introducing.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/MerkleNodeSerializer.kt
```diff
@@ -0,0 +1,45 @@
...
+        return when (jsonElement.keys) {
+            binaryNodeKeys -> {
+                decoder.json.decodeFromJsonElement(NodeHashToNodeMappingItem.BinaryNode.serializer(), element)
+            }
+            edgeNodeKeys -> {
+                decoder.json.decodeFromJsonElement(NodeHashToNodeMappingItem.EdgeNode.serializer(), element)
+            }
+            else -> throw IllegalArgumentException("Invalid MerkleNode JSON object: $jsonElement")
+        }
+    }
```
franciszekjob, line 34: note: We could potentially give an error with more meaningful message (saying which fields are missing and if it's binary/edge node). However, in case of json with keys that belong both to binary and edge node
e.g.
```
{
    "left": ...,
    "path": ...
}
```
we can't clearly decide which type it is intended to be (binary/edge). I'm not fully sure if such logic is worth introducing. 

A simple improvement would be to display required keys for binary/edge nodes.

lib/src/main/kotlin/com/swmansion/starknet/data/serializers/MerkleNodePolymorphicSerializer.kt
```diff
@@ -0,0 +1,21 @@
...
+        val binaryNodeKeys = NodeHashToNodeMappingItem.BinaryNode.serializer().descriptor.elementNames.toSet()
+        val edgeNodeKeys = NodeHashToNodeMappingItem.EdgeNode.serializer().descriptor.elementNames.toSet()
+
+        return when (jsonElement.keys) {
+            binaryNodeKeys -> NodeHashToNodeMappingItem.BinaryNode.serializer()
+            edgeNodeKeys -> NodeHashToNodeMappingItem.EdgeNode.serializer()
+            else -> throw IllegalArgumentException("Invalid MerkleNode JSON object: $jsonElement")
+        }
+    }
+}
```
franciszekjob, line 21: note: We could potentially give an error with more meaningful message (saying which fields are missing and if it's binary/edge node). However, in case of json with keys that belong both to binary and edge node
e.g.
```
{
    "left": ...,
    "path": ...
}
```
we can't clearly decide which type it is intended to be (binary/edge). I'm not fully sure if such logic is worth introducing. 

A simple improvement would be to display required keys for binary/edge nodes.

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -56,7 +56,7 @@ class DevnetClient(
 
     companion object {
         // Source: https://github.com/0xSpaceShard/starknet-devnet-rs/blob/47ee2a73c227ee356f344ce94e5f61871299be80/crates/starknet-devnet-core/src/constants.rs
-        val accountContractClassHash = Felt.fromHex("0x61dac032f228abef9c6626f995015233097ae253a7f72d68552db02f2971b8f")
+        val accountContractClassHash = Felt.fromHex("0x044cab2e6a3a7bc516425d06d76c6ffd56ae308864dbc66f8e75028e3784aa29")
```
franciszekjob, line 59: note: Class hash of predeployed OZ accounts has changed.

image

franciszekjob:
note: Support for [`starknet_getCompiledCasm`](https://github.com/starkware-libs/starknet-specs/blob/4b351773f92feb7a2b1b9a754ff525ee0ba194bb/api/starknet_executables.json#L11) and websockets has been omitted as there is not need for it ATM.

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -366,13 +410,14 @@ class DevnetClient(
...
             "--accounts-file",
             accountFilePath.absolutePathString(),
-            "--url",
-            rpcUrl,
             "--account",
             accountName,
             command,
             *(args.toTypedArray()),
+            "--url",
+            rpcUrl,
```
DelevoXDG, line 418: Why this change?

lib/kotlin-guide.md
```diff
@@ -209,13 +209,10 @@ val newAccount = StandardAccount(
...
 )
-val l1ResourceBounds = ResourceBounds(
-    maxAmount = Uint64(20000),
-    maxPricePerUnit = Uint128(120000000000),
-)
+
 val params = DeployAccountParamsV3(
     nonce = Felt.ZERO,
-    l1ResourceBounds = l1ResourceBounds,
+    resourceBounds = resourceBounds,
```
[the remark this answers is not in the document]
franciszekjob: In fact, they are defined as a field in test class, but because of this they aren't visible in code snippet (took from test). Fixed.

lib/src/main/kotlin/com/swmansion/starknet/data/types/ExecutionResources.kt
```diff
@@ -0,0 +1,18 @@
+@file:JvmName("Resources")
```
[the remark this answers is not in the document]
franciszekjob: Not needed, accidentally left.

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -246,83 +246,50 @@ class StandardAccountTest {
...
 
             val feeEstimates = request.send()
             feeEstimates.values.forEach {
                 assertNotEquals(Felt.ZERO, it.overallFee)
                 assertEquals(
-                    it.gasPrice.value * it.gasConsumed.value + it.dataGasPrice.value * it.dataGasConsumed.value,
+                    it.l1GasPrice.value * it.l1GasConsumed.value + it.l1DataGasPrice.value * it.l1DataGasConsumed.value + it.l2GasPrice.value * it.l2GasConsumed.value,
                     it.overallFee.value,
                 )
             }
```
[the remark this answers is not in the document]
franciszekjob: > Correct me if I'm wrong, but there's no reason to run assertions forEach estimate fee, since you're now estimating a fee for just one tx 🤔

Absolutely, it could be simplified without the loop.

> Isn't removing v1 and v2 txs an objective of https://github.com/software-mansion/starknet-jvm/issues/532?
> I understand where you're coming from, but because this process seems to be split over multiple PRs, some > 
things might be left without attention.

Right, it should be removed in following PR, restored to make things clear 👍 .

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
franciszekjob: url was a common argument before, now it's required for specific commands, see [foundry's changelog](https://github.com/foundry-rs/starknet-foundry/blob/master/CHANGELOG.md#changed-17) as reference.

lib/src/main/kotlin/com/swmansion/starknet/data/types/Responses.kt
```diff
@@ -79,10 +88,6 @@ data class EstimateFeeResponse(
     /**
      * Convert estimated fee to resource bounds with applied multipliers.
      *
-     * Calculates max amount as maxAmount = [overallFee] / [gasPrice], unless [gasPrice] is 0, then maxAmount is 0.
-     * Calculates max price per unit as maxPricePerUnit = [gasPrice].
-     * Then multiplies maxAmount by round([amountMultiplier] * 100%) and maxPricePerUnit by round([unitPriceMultiplier] * 100%) and performs integer division by 100 on both.
-     *
```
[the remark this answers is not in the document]
franciszekjob: Wdym? We're not performing these calculations anymore.

lib/src/main/kotlin/com/swmansion/starknet/data/types/SimulatedTransaction.kt
```diff
@@ -65,7 +65,10 @@ data class FunctionInvocation(
     val messages: List<OrderedMessageL2ToL1>,
 
     @SerialName("execution_resources")
-    val computationResources: ComputationResources,
+    val executionResources: ExecutionResources,
```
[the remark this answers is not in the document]
franciszekjob: God catch

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -1144,72 +853,7 @@ class StandardAccountTest {
...
-            ).send()
-
-            assertEquals(2, simulationResult2.values.size)
-            assertTrue(simulationResult.values[0].transactionTrace is InvokeTransactionTraceBase)
-            assertTrue(simulationResult.values[0].transactionTrace is InvokeTransactionTrace)
-            assertTrue(simulationResult.values[1].transactionTrace is DeployAccountTransactionTrace)
-        }
-
-        @Test
+        @Disabled("TODO(#538): It fails because tx resources don't cover validation or the minimal transaction fee, even though they are be high enough to validate.")
```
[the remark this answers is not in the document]
franciszekjob: It passes now thanks to fix with inner call execution resources 👍

lib/src/test/kotlin/starknet/account/StandardAccountTest.kt
```diff
@@ -625,29 +528,15 @@ class StandardAccountTest {
...
-            val result = account.executeV1(call, estimateFeeMultiplier = 1.59).send()
+            val result = account.executeV3(call, resourceBounds).send()
 
             val receipt = provider.getTransactionReceipt(result.transactionHash).send()
 
             assertTrue(receipt.isAccepted)
         }
 
         @Test
+        @Disabled("FIXME")
```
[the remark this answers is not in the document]
franciszekjob: Added issue number.

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -107,7 +108,7 @@ class DevnetClient(
             stateArchiveCapacity.value,
         )
         devnetProcess = devnetProcessBuilder.start()
-        devnetProcess.waitFor(3, TimeUnit.SECONDS)
+        devnetProcess.waitFor(10, TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
franciszekjob: In fact we can bring it back to 3s, before it failed because devnet took a bit longer than that to start, now it's ok, may be related to forking configuration that I've used at that time. Restored to original value.

.github/workflows/checks.yml
```diff
@@ -33,39 +36,78 @@ jobs:
...
+      - name: Download starknet-devnet
         run: |
-          git clone https://github.com/0xSpaceShard/starknet-devnet-rs.git starknet-devnet-rs
-          pushd ${{ github.workspace }}/starknet-devnet-rs
-          git checkout ${{ env.DEVNET_SHA }}
-          popd
+          if [[ "$RUNNER_OS" == "Linux" ]]; then
+            ARCHIVE_URL="https://github.com/franciszekjob/starknet-devnet/releases/download/v0.13.4-rc.0/starknet-devnet-x86_64-unknown-linux-gnu.tar.gz"
+          elif [[ "$RUNNER_OS" == "macOS" ]]; then
+            ARCHIVE_URL="https://github.com/franciszekjob/starknet-devnet/releases/download/v0.13.4-rc.0/starknet-devnet-aarch64-apple-darwin.tar.gz"
```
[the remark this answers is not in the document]
franciszekjob: > Also, I've noticed that this is customly built version, from your own fork - can you provide any details on the reason for that change?

 I wanted to use binary installation, but devnet team didn't have a pre/release (supporting gas changes) at that time, so I switched to fork, form which I could build a custom release.

> Correct me if I'm wrong, but your [tag](https://github.com/franciszekjob/starknet-devnet/commits/v0.13.4-rc.0/) basically adds commit and reverts that same commit, so it seems to me you can just use c657668906e0186b8b570520e371ac89f247a40f (or some newer commit) from the shardlabs repo instead.

You're right, we could use commit before these two. Now let's use a newer one (from [starknet-0.13.4-snrs](https://github.com/0xSpaceShard/starknet-devnet/tree/starknet-0.13.4-snrs) branch targets `starknet-0.13.4`).

> Thoughts?

Let's bring it back to use cloned repo and specific commit 👍

lib/kotlin-guide.md
```diff
@@ -273,72 +270,11 @@ val feePayload = provider.getEstimateFee(listOf(payloadForFeeEstimation)).send()
 
 
 ## Deploying account V1
```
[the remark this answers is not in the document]
franciszekjob: It's because the test v1 deployment was deleted, hence, it wasn't generated in guide. Restored 🫡

lib/kotlin-guide.md
```diff
@@ -273,72 +270,11 @@ val feePayload = provider.getEstimateFee(listOf(payloadForFeeEstimation)).send()
...
-val tx = provider.getTransaction(response.transactionHash).send() as DeployAccountTransactionV1
-// Invoke function to make sure the account was deployed properly
-val call = Call(balanceContractAddress, "increase_balance", listOf(Felt(10)))
-val result = account.executeV1(call).send()
-
-val receipt = provider.getTransactionReceipt(result.transactionHash).send()
-```
 
 
 ## Estimating fee for deploy account V1 transaction
```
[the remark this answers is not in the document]
franciszekjob: Same as in https://github.com/software-mansion/starknet-jvm/pull/530/files#r2005453320

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
DelevoXDG: Hmmm, I get where you're coming from; but, since this is no longer universal flag, and we might want to use some other commands in the future, maybe  `--url` should only be added when constructing params for each specific calls

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -202,10 +207,18 @@ class DevnetClient(
...
+            "--l1-gas-price",
+            "10000000000000",
+            "--l2-gas",
+            "1000000000",
+            "--l2-gas-price",
+            "100000000000000000",
+            "--l1-data-gas",
+            "100000",
+            "--l1-data-gas-price",
+            "10000000000000",
```
DelevoXDG, line 221: Not sure hardcoding these is ideal, maybe we should instead either:
- pass them as args to all applicable methods (`invokeContract`, `deployAccount`, etc.) 
- add separate method that constructs fee-related params, and use it in all mentioned methods

WDYT?

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -381,7 +426,7 @@ class DevnetClient(
         val error = String(process.errorStream.readAllBytes())
         requireNoErrors(command, error)
 
-        // As of sncast 0.24.0, declare command returns three response objects
+        // As of sncast 0.40.0, declare command returns three response objects
```
DelevoXDG, line 429: Just to clarify, are you sure this is still not fixed? 🤔

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -815,6 +829,85 @@ class ProviderTest {
...
+            on { send(any()) } doReturn HttpResponse(true, 200, mockedResponse)
+        }
+        val provider = JsonRpcProvider(rpcUrl, httpService)
+
+        val request = provider.getStorageProof(
+            blockId = BlockId.Number(0),
+        )
+        val response = request.send()
+
+        assertNotNull(response)
```
DelevoXDG, line 896: afaik it's not optional
```suggestion

        assertNotNull(response)
```

DelevoXDG (commented):
Some minor things left, seems good otherwise

lib/src/main/kotlin/com/swmansion/starknet/data/types/Resources.kt
```diff
@@ -1,98 +1,33 @@
...
-
-    @SerialName("ecdsa_builtin_applications")
-    override val ecdsaApplications: Int? = null,
-
-    @SerialName("bitwise_builtin_applications")
-    override val bitwiseApplications: Int? = null,
-
-    @SerialName("keccak_builtin_applications")
-    override val keccakApplications: Int? = null,
+open class InnerCallExecutionResources(
```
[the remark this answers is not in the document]
franciszekjob: No need, removed.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
franciszekjob: What's the difference between actual code and suggested change?

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
franciszekjob: Yup, I've checked it, here is example output:

```
[{"status":"compiling","message":"Contracts v0.1.0 (/Users/franciszekjob/Projects/starknet/starknet-jvm/lib/src/test/resources/contracts/Scarb.toml)"},
{"status":"finished","message":"release target(s) in 1 second"},
{"transaction_hash":"0x0033c593134b6054f0822f24996f9770cc900d1c210040397934db9fada4cb3a","command":"declare","class_hash":"0x031de86764e5a6694939a87321dad5769d427790147a4ee96497ba21102c8af9"}]
```

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
franciszekjob: > pass them as args to all applicable methods (invokeContract, deployAccount, etc.)

In fact, what's the benefit of that? I don't see any tbh, we use it only for internal tests setup.

> add separate method that constructs fee-related params, and use it in all mentioned methods

That's fine 👍

franciszekjob: In fact, we can use them combined - refactored.

.github/workflows/checks.yml
```diff
@@ -58,14 +61,32 @@ jobs:
...
+        run: |
+          if [ "${{ runner.os }}" = "Linux" ]; then
+            ASDF_PLATFORM="linux-amd64"
+          else
+            ASDF_PLATFORM="darwin-arm64"
+          fi
+          mkdir -p "$HOME/.local/bin"
+          DOWNLOAD_URL="https://github.com/asdf-vm/asdf/releases/download/v${LATEST}/asdf-v${LATEST}-${ASDF_PLATFORM}.tar.gz"
+          echo "Downloading asdf from $DOWNLOAD_URL"
+          curl -fsSL "https://github.com/asdf-vm/asdf/releases/download/v${LATEST}/asdf-v${LATEST}-${ASDF_PLATFORM}.tar.gz" | tar xzf - -C "$HOME/.local/bin"
```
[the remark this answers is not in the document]
franciszekjob: Fine, will address it in separate PR

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
DelevoXDG: Sorry! fixed

lib/src/test/kotlin/network/account/AccountTest.kt
```diff
@@ -261,13 +265,23 @@ class AccountTest {
...
+                maxPricePerUnit = Uint128(10000000000000000),
+            ),
+            l2Gas = ResourceBounds(
+                maxAmount = Uint64(100000000000000),
+                maxPricePerUnit = Uint128(1000000000000000000),
+            ),
+            l1DataGas = ResourceBounds(
+                maxAmount = Uint64(100000000000),
+                maxPricePerUnit = Uint128(10000000000000000),
+            ),
```
DelevoXDG, line 280: https://github.com/software-mansion/starknet-jvm/pull/533#discussion_r2020505333

Same as here, can we add format numbers with underscores, and replace this with var use / helper method **where possible**?

We can always add resource bounds if we have `docsStart` / `docsEnd` comments, and not add them otherwise.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
```diff
@@ -86,6 +86,44 @@ class ProviderTest {
...
+                    "finality_status": "ACCEPTED_ON_L2"
+                },
+                {
+                    "transaction_hash": "0x123",
+                    "finality_status": "ACCEPTED_ON_L2",
+                    "failure_reason": "Example failure reason"
+                }
+            ]
+        }
+        """.trimIndent()
```
DelevoXDG, line 107: Have you looked into [sending messages](https://0xspaceshard.github.io/starknet-devnet/docs/postman#mock-transactions) in devnet?

Not sure this necessary important to block this particular PR, but overall, if not hard to implement, would make these tests a bit more grounded in reality.

DelevoXDG (commented):
Two minor test-related comments, good job otherwise 👍🏻

lib/src/test/kotlin/network/account/AccountTest.kt
franciszekjob: > Same as here, can we add format numbers with underscores, 

👍 

> replace this with var use / helper method where possible?

This will require us to remember that in tests which will get embedded, we will have to swap this helper method to local (test scope) resource bounds variable.

lib/src/test/kotlin/starknet/provider/ProviderTest.kt
franciszekjob: https://github.com/software-mansion/starknet-jvm/issues/540

DelevoXDG (approved):

PR #533, opened 2025-03-11
Remove write api for old transactions

### LINKED ISSUE #532: Remove write api for v1, v2 transactions
### Feature Request
We should remove write api for old txs:
- Remove docs examples including txs other than v3
- Remove relevant methods/classes
- Remove tests

franciszekjob:
## Describe your changes

Remove write API for old transactions.

## Linked issues

Closes #532 

## Breaking changes

Remove all write methods for v1 and v2 transactions.


lib/src/main/kotlin/com/swmansion/starknet/data/types/Params.kt
DelevoXDG, line 1: This unrelated to this file, but shouldn't we update `Transaction.kt` as well, and remove constructors that are meant for creating tx for execution from `DeclareTransactionV2`, `DeclareTransactionV1`, `DeclareTransactionV0`, related `Cairo1ContractDefinition` (it's not supported anymore, right), and same for other tx types?

We also can surely remove `ConvertingToPayloadFailedException`

lib/guide.md
```diff
@@ -1017,9 +811,19 @@ val privateKey = Felt(0x1)
...
+        maxPricePerUnit = Uint128(10000000000000000),
+    ),
+    l2Gas = ResourceBounds(
+        maxAmount = Uint64(100000000000000),
+        maxPricePerUnit = Uint128(1000000000000000000),
+    ),
+    l1DataGas = ResourceBounds(
+        maxAmount = Uint64(100000000000),
+        maxPricePerUnit = Uint128(10000000000000000),
+    ),
```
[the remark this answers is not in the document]
franciszekjob: > Can we add underscores for numer values i.e. 1_000_...?

👍 

> Maybe we should move these to helper method, or set as variable in test class?

That was my initial idea, but the problem us that these values won't be embedded in code sample, in the docs, i.e. we should keep them directly in the test.

lib/src/test/kotlin/network/account/AccountTest.kt
```diff
@@ -173,48 +108,17 @@ class AccountTest {
         val feeEstimateRequest = provider.getEstimateFee(listOf(declareTransactionPayload), BlockTag.PENDING)
 
         val feeEstimate = feeEstimateRequest.send().values.first()
-        assertNotEquals(Felt(0), feeEstimate.l1GasConsumed)
+        assertTrue(feeEstimate.l1GasConsumed == Felt(0))
```
DelevoXDG, line 111: Nit:
```suggestion
        assertEquals(Felt(0), feeEstimate.l1GasConsumed)
```

lib/src/main/kotlin/com/swmansion/starknet/data/types/Transaction.kt
```diff
@@ -742,38 +656,7 @@ data class DeployAccountTransactionV1 private constructor(
...
-        nonce: Felt = Felt.ZERO,
-    ) : this(
-        classHash = classHash,
-        contractAddress = contractAddress,
-        contractAddressSalt = salt,
-        constructorCalldata = calldata,
-        version = if (forFeeEstimate) TransactionVersion.V1_QUERY else TransactionVersion.V1,
-        nonce = nonce,
-        maxFee = maxFee,
-        hash = TransactionHashCalculator.calculateDeployAccountV1TxHash(
```
DelevoXDG, line 765: This probably means we can remove `calculateDeployAccountV1TxHash` as well

lib/src/main/kotlin/com/swmansion/starknet/data/types/Transaction.kt
```diff
@@ -225,7 +225,7 @@ data class InvokeTransactionV1 private constructor(
     @SerialName("nonce")
     override val nonce: Felt,
 
-) : InvokeTransaction(), DeprecatedTransaction, ExecutableTransaction {
+) : InvokeTransaction(), DeprecatedTransaction {
```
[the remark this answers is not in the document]
DelevoXDG: > maybe even with default values, that allows setting the tx hash, manually.

Yes, but what you linked is the constructor I suggest removing in favor of constructor that accepts plain hash - we're no longer sending this transactions for execution, so I'm not sure there's any value in us calculating their hashes. 

WDYT?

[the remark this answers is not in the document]
franciszekjob: Ok fine, removed.

lib/kotlin-guide.md
```diff
@@ -8,24 +8,21 @@ Although written in Kotlin, Starknet-jvm has been created with compatibility wit
 <!-- TOC -->
 ## Table of contents
 * [Using provider](#using-provider)
-* [Reusing provider](#reusing-providers)
+* [Reusing providers](#reusing-providers)
```
DelevoXDG, line 11: Nit: 

Would actually make more sense to retitle the section to "Reusing provider"

lib/src/main/kotlin/com/swmansion/starknet/data/TransactionHashCalculator.kt
```diff
@@ -67,33 +48,6 @@ object TransactionHashCalculator {
...
-        version: TransactionVersion,
-        maxFee: Felt,
-        nonce: Felt,
-    ): Felt {
-        val contractAddress = ContractAddressCalculator.calculateAddressFromHash(
-            classHash = classHash,
-            calldata = calldata,
-            salt = salt,
-        )
-        return transactionHashCommon(
```
DelevoXDG, line 85: We can delete `transactionHashCommon`

DelevoXDG (approved):
Approving, but please address the two remaining comments

PR #555, opened 2025-04-14
Android improvements

franciszekjob:
## Describe your changes

- include libs for `aarch64` and `x86_64`
- set page size to 16 KB
- print stack trace when native loader fails
- modify lib path resolution logic in `NativeLoader.getLibPath()`

## Linked issues

Closes

## Breaking changes

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -39,15 +39,15 @@ jobs:
           cache: 'gradle'
 
       # See https://github.com/software-mansion/starknet-jvm/issues/171
-      - name: Set up GCC 7
-        run: sudo apt update && sudo apt install -y g++-7
+      - name: Set up GCC 11
+        run: sudo apt update && sudo apt install -y g++-11
         if: matrix.os == 'ubuntu-22.04'
       - name: Force usage of gcc 7
         run: |
           # First delete any version other than 7
           sudo find /usr/lib/gcc/x86_64-linux-gnu/ -maxdepth 1 -mindepth 1 ! -name 7 -exec rm -rf {} +
-          echo "CC=/usr/bin/gcc-7" >> $GITHUB_ENV
-          echo "CXX=/usr/bin/g++-7" >> $GITHUB_ENV
+          echo "CC=/usr/bin/gcc-11" >> $GITHUB_ENV
+          echo "CXX=/usr/bin/g++-11" >> $GITHUB_ENV
         shell: bash
         if: matrix.os == 'ubuntu-22.04'
 
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -151,9 +151,9 @@ jobs:
           cp -r download/lib-macOS*/* lib/build/libs/shared/darwin/
           mkdir -p lib/build/libs/shared/linux/{amd64,arm64,aarch64,x86_64}
           cp -r download/lib-Linux-x86_64/* lib/build/libs/shared/linux/amd64/
+          cp -r download/lib-Linux-x86_64/* lib/build/libs/shared/linux/x86_64/
           cp -r download/lib-Linux-aarch64/* lib/build/libs/shared/linux/arm64/
           cp -r download/lib-Linux-aarch64/* lib/build/libs/shared/linux/aarch64/
-          cp -r download/lib-Linux-x86_64/* lib/build/libs/shared/linux/x86_64/
       - name: Bundle universal JAR
         run: ./gradlew lib:jar
       - name: Rename JAR
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -9,7 +9,7 @@ on: push
 
 jobs:
 #  verify_version:
-#    runs-on: ubuntu-latest
+#    runs-on: ubuntu-22.04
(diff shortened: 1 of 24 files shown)


lib/src/main/kotlin/com/swmansion/starknet/crypto/NativeLoader.kt
```diff
@@ -62,8 +63,28 @@ internal object NativeLoader {
...
+                if (androidAbi != null) {
+                    return "/jni/$androidAbi/$name.so"
+                }
+
+                // Fallback to the default path
+                "/linux/$architecture/$name.so"
+            }
+
+            else ->
+                throw UnsupportedPlatform(operatingSystem.name, architecture)
```
franciszekjob, line 87: note: I've checked returned lib path for device with `arm64-v8a` as supported abi (for temporary purpose method wasn't private, hence I could call it):

image

.github/workflows/build-and-publish.yml
```diff
@@ -149,9 +149,11 @@ jobs:
           # Mac uses a universal binary
           mkdir -p lib/build/libs/shared/darwin
           cp -r download/lib-macOS*/* lib/build/libs/shared/darwin/
-          mkdir -p lib/build/libs/shared/linux/{amd64,arm64}
+          mkdir -p lib/build/libs/shared/linux/{amd64,arm64,aarch64,x86_64}
           cp -r download/lib-Linux-x86_64/* lib/build/libs/shared/linux/amd64/
           cp -r download/lib-Linux-aarch64/* lib/build/libs/shared/linux/arm64/
+          cp -r download/lib-Linux-aarch64/* lib/build/libs/shared/linux/aarch64/
+          cp -r download/lib-Linux-x86_64/* lib/build/libs/shared/linux/x86_64/
```
THenry14, line 156: nit: let's group it, so x86 and arm are next to each other

crypto/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@ include_directories("${CMAKE_SOURCE_DIR}/poseidon/bindings")
 include_directories("${CMAKE_SOURCE_DIR}/pedersen/crypto-cpp/src")
 include_directories("${CMAKE_SOURCE_DIR}/pedersen/bindings")
 
+set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-z,max-page-size=0x4000")
```
THenry14, line 11: what does it do?

lib/src/main/kotlin/com/swmansion/starknet/crypto/NativeLoader.kt
```diff
@@ -62,8 +63,28 @@ internal object NativeLoader {
     private fun getLibPath(system: SystemType, architecture: String, name: String): String {
         return when (system) {
             SystemType.MacOS -> "/darwin/$name.dylib"
-            SystemType.Linux -> "/linux/$architecture/$name.so"
-            else -> throw UnsupportedPlatform(operatingSystem.name, architecture)
+            SystemType.Linux -> {
```
THenry14, line 66: maybe it would be possible to have a separate system type for android?

franciszekjob: Keep in mind that if Android app uses JAR (which is possible) then the path should still include linux. Your suggestion makes sense but it won't work in all possible cases.

crypto/CMakeLists.txt
franciszekjob: It sets page size to 16 KB, [ref](https://developer.android.com/guide/practices/page-sizes).

lib/src/main/kotlin/com/swmansion/starknet/crypto/NativeLoader.kt
THenry14: this will be if android is x86 right? or some other cases are possible as well?
if true, maybe we should also check the architecture for android?

let's think about it and discuss first, if the effort is not worth it it may be good idea to just leave it as is right now

crypto/CMakeLists.txt
THenry14: min sdk is set to 24 which is older than android 15 right? from the ref you've linked you can learn that older androids had only 4 kB, so won't this be a problem?

franciszekjob: As [docs](https://android-developers.googleblog.com/2024/08/adding-16-kb-page-size-to-android.html?utm_source=chatgpt.com) say:

> In Android 15, we’ve refactored Android from the ground up to support running at different page sizes, thus making it page-size agnostic.

So I think we can safely change it.

franciszekjob: Tbh I can't find thr exact information in the docs, but I've tested the lib on emulator with Android 14 (below 15) and argent team member confirmed it.

franciszekjob:
@THenry14 I've moved the changes into separate PRs

lib/src/main/kotlin/com/swmansion/starknet/crypto/NativeLoader.kt
franciszekjob: It turns out this changes is not needed - this logic is fine and we don't have to adjust it.

PR #561, opened 2025-05-08
Adjust `architecture` resolution

franciszekjob:
## Linked issues

Closes

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/crypto/NativeLoader.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/crypto/NativeLoader.kt
@@ -17,7 +17,12 @@ internal object NativeLoader {
     }
 
     private val architecture: String by lazy {
-        System.getProperty("os.arch")
+        val arch = System.getProperty("os.arch")
+        when {
+            arch.contains("aarch64") || arch.contains("arm64") -> "arm64"
+            arch.contains("x86_64") || arch.contains("amd64") -> "x86_64"
+            else -> arch
+        }
     }
 
     fun load(name: String) = load(name, operatingSystem, architecture)


lib/src/main/kotlin/com/swmansion/starknet/crypto/NativeLoader.kt
```diff
@@ -17,7 +17,14 @@ internal object NativeLoader {
     }
 
     private val architecture: String by lazy {
-        System.getProperty("os.arch")
+        // Currently, our generated jar includes 2 subdirectories under linux: amd64 and arm64.
+        // Following adjustment is needed to load the correct library, for missing aarch64 and x86_64 directories.
+        val arch = System.getProperty("os.arch")
+        when {
+            arch.contains("aarch64") || arch.contains("arm64") -> "arm64"
+            arch.contains("x86_64") || arch.contains("amd64") -> "amd64"
```
DelevoXDG, line 25: shouldn't we check if it equals, not contains 🤔?

lib/src/main/kotlin/com/swmansion/starknet/crypto/NativeLoader.kt
```diff
@@ -17,7 +17,14 @@ internal object NativeLoader {
...
     private val architecture: String by lazy {
-        System.getProperty("os.arch")
+        // Currently, our generated jar includes 2 subdirectories under linux: amd64 and arm64.
+        // Following adjustment is needed to load the correct library, for missing aarch64 and x86_64 directories.
+        val arch = System.getProperty("os.arch")
+        when {
+            arch.equals("aarch64") || arch.equals("arm64") -> "arm64"
+            arch.equals("x86_64") || arch.equals("amd64") -> "amd64"
+            else -> arch
+        }
```
DelevoXDG, line 27: Can be simplified to:
```kt
        val arch = System.getProperty("os.arch")
        when (arch) {
            "aarch64", "arm64" -> "arm64"
            "x86_64", "amd64" -> "amd64"
            else -> arch
        }
```
or even:
```kt
        val arch = System.getProperty("os.arch")
        when (arch) {
            "aarch64" -> "arm64"
            "x86_64" -> "amd64"
            else -> arch
        }
```

lib/src/main/kotlin/com/swmansion/starknet/crypto/NativeLoader.kt
DelevoXDG, line 1: Just to clarify, have you tested this locally on some image that has such specific arch? 🤔 

I also wonder if there's any runners that would allow us to run at least basic tests on CI, that only cover the scope of binary dependencies?

Similar to these:
```kt
public class DepsTest {
    @Test
    public void testPedersen() {
        var pedersen = StarknetCurve.pedersen(Felt.ONE, new Felt(2));
        assertNotEquals(Felt.ZERO, pedersen);
    }

    @Test
    public void testPoseidon() {
        var poseidon = Poseidon.poseidonHash(Felt.ONE, new Felt(2));
        assertNotEquals(Felt.ZERO, poseidon);
    }
}
```

Issue #563, opened 2025-05-15
Strange behaviour of pedersen(Iterable)


MEJIOMAH17:
What is the reason why ```pedersen(listOf(a,b)) != pedersen(a,b)```?

PR #564, opened 2025-05-15
Fix pedersen hash

MEJIOMAH17:
## Describe your changes

Fix https://github.com/software-mansion/starknet-jvm/issues/563

## Linked issues

Closes https://github.com/software-mansion/starknet-jvm/issues/563

## Breaking changes

- [ ] pedersen(Iterable) will return different hashes


franciszekjob (commented):
Could you please add a test checking result of calling pedersen both on array (of length 2) and simply 2 elements 🙏 ?

PR #571, opened 2025-07-16
Update pattern when moving *.jar files

franciszekjob:
## Describe your changes

The issue arised because we tried to move files matched by strict pattern (`lib-[0-9].[0-9][0-9].[0-9].jar`, similarly for javadoc and source JARS), which resulted in crash when creating rc version.

## Linked issues

Closes

## Breaking changes

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -155,7 +155,7 @@ jobs:
       - name: Bundle universal JAR
         run: ./gradlew lib:jar
       - name: Rename JAR
-        run: mv lib/build/libs/lib-[0-9].[0-9][0-9].[0-9].jar starknet.jar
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-publish.yml
```diff
@@ -155,7 +155,7 @@ jobs:
       - name: Bundle universal JAR
         run: ./gradlew lib:jar
       - name: Rename JAR
-        run: mv lib/build/libs/lib-[0-9].[0-9][0-9].[0-9].jar starknet.jar
+        run: mv lib/build/libs/lib-*.jar starknet.jar
```
franciszekjob, line 158: note: After running `./gradlew lib:jar` only one file (with lib jar) will be present so it's safe to widen the pattern.

PR #579, opened 2025-07-27
Tip auto-estimation

franciszekjob:
## Linked issues

Closes

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/helpers/Tip.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/helpers/Tip.kt
@@ -1,3 +1,5 @@
+@file:JvmName("Tip")
+
 package com.swmansion.starknet.helpers
 
 import com.swmansion.starknet.data.types.*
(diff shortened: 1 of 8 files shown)


lib/src/test/kotlin/starknet/helpers/TipTest.kt
franciszekjob, line 1: note: It's really convenient to use mocks in such case when we need to manipulate txs contained in the block.

lib/src/main/kotlin/com/swmansion/starknet/helpers/Tip.kt
```diff
@@ -0,0 +1,49 @@
+@file:JvmName("Tip")
```
franciszekjob, line 1: note: It's needed for Java, so the import path is `import static com.swmansion.starknet.helpers.Tip.estimateTip;` instead of `import static com.swmansion.starknet.helpers.TipKt.estimateTip;`

lib/build.gradle.kts
```diff
@@ -255,6 +255,8 @@ dependencies {
     implementation("com.squareup.okhttp3:okhttp:4.12.0")
 
     implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.7.1")
+
+    implementation("org.nield:kotlin-statistics:1.2.1")
```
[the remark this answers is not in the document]
franciszekjob: Yeah, apache commons seems better. Changed.

PR #593, opened 2025-11-12
Add changes from RPC 0.10.0-rc.1

### LINKED ISSUE #591: Support RPC 0.10
### Feature Request

Self-descriptive

franciszekjob:
## Describe your changes

- Add `StorageKey` class
- `EmittedEvent`: add `transactionIndex` and `eventIndex` fields
- Add `MigratedClassItem` class
- `StateDiff`: add `migratedCompiledClasses` field

## Linked issues

Closes #591

## Breaking changes

- `ContractsStorageKeys`: field `storageKeys` is now of type `List<StorageKey>`
- `PreConfirmedStateUpdateResponse`: field `oldRoot` is nullable

### DIFF
--- a/lib/src/test/kotlin/starknet/utils/DevnetClient.kt
+++ b/lib/src/test/kotlin/starknet/utils/DevnetClient.kt
@@ -412,6 +412,7 @@ class DevnetClient(
         val lines = String(process.inputStream.readAllBytes()).trim().split("\n").filter { it != "null" }
         val result = lines.last()
 
+        // TODO(#596)
         // Workaround: "command" field is not present in sncast response anymore, hence
         // we need to put it to JSON synthetically
         val jsonElement = json.parseToJsonElement(result).jsonObject.toMutableMap()


lib/src/main/kotlin/com/swmansion/starknet/data/types/Event.kt
```diff
@@ -63,7 +63,18 @@ data class EmittedEvent @JvmOverloads constructor(
 
     @SerialName("transaction_hash")
     val transactionHash: Felt,
-)
+
+    @SerialName("transaction_index")
+    val transactionIndex: Int,
```
franciszekjob: note: We can't use `UInt here`

lib/src/test/kotlin/starknet/utils/DevnetClient.kt
```diff
@@ -405,9 +409,29 @@ class DevnetClient(
...
+        }
+        jsonElement["command"] = JsonPrimitive(commandStr)
+
+        // Remove fields with null values
+        jsonElement.entries.removeIf { it.value is JsonNull }
+
+        return json.decodeFromString(
+            SnCastResponsePolymorphicSerializer,
+            json.encodeToString(JsonObject(jsonElement)),
+        )
```
franciszekjob: note: Technically we should bring it back in `sncast` because it was a breaking change (which we didn't notice 😅 ).

lib/src/main/kotlin/com/swmansion/starknet/data/types/Event.kt
cptartur: Why is that?

PR #595, opened 2025-11-13
Blake hash support

### LINKED ISSUE #592: Change hashing function from poseidon to blake
Change hashing function from poseidon (currently used) to blake, for CASM hashing.
It's available in `java.security`, e.g.

```kotlin
import java.security.MessageDigest
// ...
MessageDigest digest = MessageDigest.getInstance("BLAKE-256"); // or 512
```

franciszekjob:
## Describe your changes

Use blake hash for CASM class hash calculation. For RPC < 0.10.0, poseidon will still be used.

## Linked issues

Closes #592 

## Breaking changes

### DIFF
--- a/lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
+++ b/lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
@@ -27,12 +27,14 @@ object Blake {
         val unpackedInts = mutableListOf<Int>()
 
         for (value in values) {
-            val valueBytes = value.value.toByteArray().let {
-                if (it.size < 32) {
-                    ByteArray(32 - it.size) + it
-                } else {
-                    it.copyOfRange(it.size - 32, it.size)
-                }
+            val rawBytes = value.value.toByteArray()
+
+            require(rawBytes.size <= 32) {
+                "rawBytes size ${rawBytes.size} exceeds 32 bytes"
+            }
+
+            val valueBytes = rawBytes.let {
+                ByteArray(32 - it.size) + it
             }
 
             if (value.value < smallThreshold) {


lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
```diff
@@ -0,0 +1,149 @@
...
+        val hash = blake2sHash(feltToNative(first), feltToNative(second))
+        return pack256LeToFelt(hash)
+    }
+
+    /**
+     * Converts an iterable of Felts to their native byte array representations and computes their Blake2s hash.
+     *
+     * @param values Iterable of Felt values to convert.
+     * @return Felt representation of the hash.
+     */
```
cptartur: I think docs of both of these should just say it computes the hash of 2 felts. Conversions are an implementation detail

lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /**
+     * Converts a Felt to its native byte array representation (32 bytes, big-endian).
+     *
+     * @param felt Felt value to convert.
+     * @return Byte array representation of the Felt.
+     */
+    @JvmStatic
+    fun blake2sHash(first: ByteArray, second: ByteArray): ByteArray {
```
cptartur: This should be private, don't see a valid usecase for user computing hash from plain bytearrays.

lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
```diff
@@ -0,0 +1,149 @@
...
+                bytes[i] = ((word shr (8 * i)) and 0xFF).toByte()
+            }
+            bytes.asList()
+        }.toByteArray()
+
+        return blake2sToFelt(byteStream)
+    }
+
+    /**
+     * Converts a Felt to its native byte array representation (32 bytes, big-endian).
```
cptartur: There's no conversion going on here

lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
```diff
@@ -0,0 +1,149 @@
...
+import com.swmansion.starknet.extensions.toFelt
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+import java.math.BigInteger
+import java.security.MessageDigest
+import java.security.Security
+
+object Blake {
+
+    private val smallThreshold: BigInteger = BigInteger.ONE.shiftLeft(63)
+    private val bigMarker = 1 shl 31
```
cptartur: This could be const val

lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
```diff
@@ -0,0 +1,149 @@
...
+     *
+     * @param felts List of Felt values to hash.
+     * @return Felt representation of the hash.
+     */
+    @JvmStatic
+    fun blake2sHash(felts: List<Felt>): Felt {
+        val u32Words = encodeFeltsToU32s(felts)
+
+        // Serialize as little-endian bytes
+        val byteStream = u32Words.flatMap { word ->
```
cptartur: Nit, rename this because Streams are an existing concept in java.

lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Computes the Blake2s hash of a list of Felts.
+     *
+     * @param felts List of Felt values to hash.
+     * @return Felt representation of the hash.
+     */
+    @JvmStatic
+    fun blake2sHash(felts: List<Felt>): Felt {
+        val u32Words = encodeFeltsToU32s(felts)
```
cptartur: What naming scheme do we want to follow there, in kotlin/java Int is always u32, so it's bit weird that we copied the naming scheme from rust impl directly.

It makes it easier to compare but doesn't make much sense in Java world.

lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
```diff
@@ -0,0 +1,149 @@
...
+
+        // Serialize as little-endian bytes
+        val byteStream = u32Words.flatMap { word ->
+            val bytes = ByteArray(4)
+            for (i in 0 until 4) {
+                bytes[i] = ((word shr (8 * i)) and 0xFF).toByte()
+            }
+            bytes.asList()
```
cptartur: Not 100% sure if that works correctly but maybe ByteBuffer can be used here (and in other places maybe?) so we don't have to do binary operations ourselves


```suggestion             ByteBuffer.allocate(4).order(LITTLE_ENDIAN).putInt(word).array().asList()
```

lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * Computes the Blake2s hash of the given data and converts it to a Felt.
+     *
+     * @param data Byte array to hash.
+     * @return Felt representation of the hash.
+     */
+    private fun blake2sToFelt(data: ByteArray): Felt {
+        val digest = MessageDigest.getInstance("BLAKE2S-256", "BC")
+        val hash = digest.digest(data)
+        return pack256LeToFelt(hash)
```
cptartur: We could do something along the lines of this (please verify) and then you likely won't need to setup any java security as you do above.


```suggestion
        val digest = Blake2sDigest(256)
        digest.update(first, 0, first.size)
        digest.update(second, 0, second.size)
        val hash = ByteArray(32)
        digest.doFinal(hash, 0)
        return hash
```

cptartur: Are we sure that the digest returned by this is little endian?

lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /**
+     * Computes the Blake2s hash of a list of Felts.
+     *
+     * @param felts List of Felt values to hash.
+     * @return Felt representation of the hash.
+     */
+    @JvmStatic
+    fun blake2sHash(felts: List<Felt>): Felt {
```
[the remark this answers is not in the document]
franciszekjob: > How's that different signature wise from the method we have below with signature like `values: Iterable<Felt>`?

See [Posiedon implementation](https://github.com/software-mansion/starknet-jvm/blob/main/lib/src/main/kotlin/com/swmansion/starknet/crypto/Poseidon.kt) - I wanted these signatures to be the same.

lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
cptartur: It shouldn't, if you replace all usages of `MessageDigest` with `Blake2sDigest` then it runs for me without any setup of the security provider.

franciszekjob: Yes, it looks so - https://github.com/bcgit/bc-java/blob/65c3dc40c22c5dce2dc9c05de4da29705ea869b5/core/src/main/java/org/bouncycastle/crypto/digests/Blake2sDigest.java#L502

lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
```diff
@@ -0,0 +1,149 @@
...
+     * Converts a Felt to its native byte array representation (32 bytes, big-endian).
+     *
+     * @param felt Felt value to convert.
+     * @return Byte array representation of the Felt.
+     */
+    @JvmStatic
+    fun blake2sHash(first: ByteArray, second: ByteArray): ByteArray {
+        val digest = MessageDigest.getInstance("BLAKE2S-256", "BC")
+        digest.update(first)
+        return digest.digest(second)
```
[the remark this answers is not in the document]
franciszekjob: Added common `blake2sHashBytes`.
`blake2sHash(first: ByteArray, second: ByteArray)`  and `blake2sToFelt(data: ByteArray)` use it

lib/src/main/kotlin/com/swmansion/starknet/crypto/Blake.kt
```diff
@@ -0,0 +1,107 @@
...
+     *
+     * @param hashBytes Byte array containing the hash.
+     * @return Felt representation of the packed value.
+     */
+    private fun pack256LeToFelt(hashBytes: ByteArray): Felt {
+        require(hashBytes.size >= 32) { "need at least 32 bytes to pack" }
+
+        val reversed = hashBytes.copyOfRange(0, 32).reversedArray()
+        val value = BigInteger(1, reversed)
+        return value.mod(Felt.PRIME).toFelt
```
cptartur: I'm not 100% sure how the reference implementation behaves in case the value is too big to felt within felt (and if that's possible). It looks like it might just fail for them in case the hash cannot fit in felt, but it'd require better verification.

PR #607, opened 2025-12-08
Allow using devnet from path

ksew1:
commit-id:6a561d0b

---

**Stack**:
- #609
- #608
- #607 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*


cptartur (approved):
See my comments, but otherwise looks good

